PR #5, opened 2021-07-07
Retain initial font size while zooming and display zoom level

ButterscotchVanilla:
- Retains the initial font size by extending the font class to store the initial size values
 - Updated the GUI class to utilize the new `ScalableFont` class for font scaling
 - Displays the current zoom level on the zoom button

### DIFF
--- a/src/main/java/io/eiren/gui/ScalableFont.java
+++ b/src/main/java/io/eiren/gui/ScalableFont.java
@@ -28,7 +28,7 @@ public class ScalableFont extends Font {
 			this.initPointSize = sourceFont.getInitSize2D();
 			
 			this.size = this.initSize;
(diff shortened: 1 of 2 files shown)


Eirenliel:
> Retains the initial font size by extending the font class to store the initial size values
> Updated the GUI class to utilize the new ScalableFont class for font scaling

But why is it needed?

> Displays the current zoom level on the zoom button

Converting float to text directly is really bad move (even though on numbers here it won't be something like 1.3000002), especially just showing it without context. Would be better to do something like `"GUI Zoom (x" + StringUtils.prettyNumber(zoom, 2) + ")"`. But also don't see a real reason for this, since GUI will be replaced completely.

PR #16, opened 2021-07-21
Update build.gradle

adigyran:
It is not working properly currently, i propose to remove it by now until further testing


adigyran:
This breaks autobuild :(

PR #32, opened 2021-08-21
Fix driver parsing bug

kitlith:
Even though this may be replaced/removed soon (if my project goes well) this is still a bug that other people might run into in the meantime, and I want to verify something against this version of input.

build.gradle changes are unrelated tbh, i just wanted a working run configuration and ended up massaging it into a structure i'm slightly more familiar with. more details in the description for that specific commit.

there are two TODOs that I added that can be fixed before merging, I just wanted to fix the main bug, and maybe we don't care about those specific details.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,8 +7,8 @@
  */
 
 plugins {
-    // Apply the java-library plugin to add support for Java Library
-    id 'java-library'
+	id 'application'
+	id "com.github.johnrengelman.shadow" version "6.1.0"
 }
 
 sourceCompatibility = 1.8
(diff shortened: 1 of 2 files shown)


Eirenliel:
Don't think it's a good idea to replace one implementation with third one, I made another one last time here: https://github.com/SlimeVR/SlimeVR-Server/blob/main/src/main/java/io/eiren/vr/bridge/SteamVRPipeInputBridge.java#L70

kitlith:
yeah can
 probably replace it with something based on that, will work on it tomorrow.

abb128:
From what I understand, the new updateHMD only reads a single entry and finishes? Wouldn't this pose a problem if the server is running slower than the driver is sending new data?

kitlith:
@abb128 well, not really? that depends on whether you think data should be straight up ignored or not.

https://github.com/SlimeVR/SlimeVR-Server/blob/83eaf62bad7f9d4006a939c42a6e6716e6276e2d/src/main/java/io/eiren/vr/bridge/NamedPipeVRBridge.java#L69-L75

Key point here is that it will only wait and assume that there are no more entries if an entry was not parsed.

frankly, though, as Eirenliel mentioned, if we're going to fix this we should probably fix it based on the new implementation, or maybe even de-duplicate it a bit.

i know i said i'd get to that today, but that didn't happen and i have no idea how i'll feel in the next few days.

Eirenliel:
Not relevant anymore: https://github.com/SlimeVR/SlimeVR-Server/commit/7c8a3941474c5bf3cdff57c65e961469e2daf107

PR #43, opened 2021-08-31
Window and Taskbar logo + Assets Downloader

Blixtdraken:
- The application window have the SlimeVR logo
- Logo downloads on start of program if not already exist
- Makes an assets directory on launch if not already exist

Logo download source: [https://www.dropbox.com/s/tidq5gdng1egoew/logo.png?dl=1](https://www.dropbox.com/s/tidq5gdng1egoew/logo.png?dl=1)  
-(I will not ever remove it from my Dropbox unless you change the source)


Eirenliel:
This is not a good solution. It should be included in the repository, not downloaded. I'll make proper implementation when I'm back home.

Blixtdraken:
> This is not a good solution. It should be included in the repository, not downloaded. I'll make proper implementation when I'm back home.

Ok then

Issue #22, opened 2021-08-12
A "Skeleton Offset"-like feature to help VRChat's avatar bind-in weirdness.


Torikari:
SEE UPDATE 1, THIS SHOULDN'T BE AT THE TOP PRIORITY AS IT'S ACTUALLY PRETTY MUCH VRCHAT'S FAULT.

While the default Head Offset value of 10 is great at increasing skeleton stability, it seems to have an effect to VRChat avatar bind-in because how it shifts the whole skeleton back, most noticeably causing some avatar's leg to not completely bend or bend too much when raising over hip level, this can be fixed by simply decreasing Head Offset to 0, this also offers more consistency when binding into different avatars, but at the cause of skeleton instability.

An additional Skeleton Offset or similar feature would be neat to balance out the skeleton offset caused by Head Offset, helping bind-in consistency in VRChat while maintaining skeleton stability.

Update 1: Problem described above happens in avatar with view position a bit on the back side, same avatar with view position that is more forward will work fine, so this doesn't have much to do with binding consistency, but more about to cope with VRChat avatar's different view position, and to get the best body motion without editing the avatar, or having to move view position to potentially undesirable location.

Eirenliel:
Implemented in #102

PR #77, opened 2021-11-06
Skeleton offset

Louka3000:
skeleton offset by offset the chest and neck horizontally to help with vrchat being weird.

https://github.com/SlimeVR/SlimeVR-Server/issues/22


Eirenliel:
This is not how #22 should be fixed. You are moving the whole skeleton forward, but you should be moving each tracker separately, without it affecting the whole skeleton hierarchy (this is also why leg offset is wrong). It sort of works, but it's not ideal. Skeleton needs refactoring to make this simpler...

PR #81, opened 2021-11-11
Keybinding: If JIntellitype fails to load, still allow server to run

carl-anders:
I thought that simply using the function "isJIntellitypeSupported" would be enough, but a user error in discord shows that JIntellitype is not to be trusted. Wrapping the init code in a try catch should resolve any issues tho. Hopefully.


carl-anders:
The new changes have been tested by the user and the server does not crash anymore.

PR #98, opened 2022-01-01
Ankles and knees can default to feet trackers

Louka3000:
If no ankles or knee trackers are found, default to feet trackers.
This is for when people mistake feet and ankle position in SlimeVR server.


Eirenliel:
What I don't like about it, is that if people actually attach trackers to feet, but miss ankles or knees, they will have a bad time...

Louka3000:
Hmm it doesn’t really matter too much anyways given that new GUI should make the tracker role assignment much easier to understand…

Eirenliel:
👍

PR #99, opened 2022-01-02
Add RSSI to trackers

deiteris:
Adds RSSI reported by trackers. Shows `N/A` if tracker is not reporting this data, and shows signal strength in percentage and dBm. The range is: -40 dBm = 100%, -95 dBm = 0%.
image

Related issue: https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/44


src/main/java/dev/slimevr/vr/trackers/TrackersUDPServer.java
```diff
@@ -373,6 +373,19 @@ public void run() {
 						socket.send(new DatagramPacket(rcvBuffer, bb.position(), connection.address));
 						System.out.println("[TrackerServer] Sensor info for " + connection.sensors.get(0).getName() + "/" + sensorId + ": " + sensorStatus);
 						break;
+					case 19:
+						if(connection == null)
+							break;
+						if(connection.isOwoTrack)
```
Eirenliel, line 379: Don't need this check, it's only for conflicting packet IDs

Eirenliel (changes requested):

PR #104, opened 2022-01-08
Ignore everything in .idea

ColdIce1605:
Idea creates lots of project files, this makes it so contributors don't have to deal with making sure they don't commit a idea related file to the Repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ build
 
 /bin/
 
+# User-specific stuff
+.idea/*
+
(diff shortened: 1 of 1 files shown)


adigyran:
already added in new gui branch bc was created in Idea, main branch mostly created using eclipse, you can get from new gui branch

ColdIce1605:
Ah, :eyes:

Eirenliel:
Well that's settled then, I guess.

PR #105, opened 2022-01-11
Show battery level reported by tracker

deiteris:
Also keeps backward compatibility with owotrack
image

### DIFF
--- a/src/main/java/dev/slimevr/vr/trackers/TrackersUDPServer.java
+++ b/src/main/java/dev/slimevr/vr/trackers/TrackersUDPServer.java
@@ -33,6 +33,12 @@ import io.eiren.util.collections.FastList;
  */
 public class TrackersUDPServer extends Thread {
 
+	enum ProtocolCompat {
+		OWOTRACK_LEGACY,
+		OWOTRACK8,
+		SLIMEVR
+	}
+
 	/**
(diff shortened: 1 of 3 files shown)


src/main/java/dev/slimevr/vr/trackers/TrackersUDPServer.java
```diff
@@ -234,7 +231,7 @@ public void run() {
 					case 17: // PACKET_ROTATION_DATA
 						if(connection == null)
 							break;
-						if(connection.isOwoTrack)
+						if(connection.firmware.equals("owoTrack"))
```
Eirenliel, line 234: Please revert it to boolean check, this boolean should be set on tracker connect, and not string check at every packet.

Eirenliel (changes requested):

src/main/java/dev/slimevr/vr/trackers/TrackersUDPServer.java
deiteris: I did this because this check doesn't apply to owoTrack for Android as it reports `owoTrack8` in firmware (haven't checked for iOS), and compatibility with both versions cannot be expressed with a single boolean check. If you're concerned by string comparison performance, then see updated version with enum.

PR #107, opened 2022-01-15
Updated default body proportions (again

Louka3000:
- Shorter torso length.
- Should be ideal for people around 5"10'?


Louka3000:
People tend to have a much longer torso and much shorter legs than they should in the body proportions ;)

PR #118, opened 2022-02-05
Switch to latest gradle and update dependencies

deiteris:
Successfully builds and works (tested most of the functionality, so nothing seems to be borken, but feel free to re-check). Few questions/notes though:
1. ~~I removed `com.illposed.osc:javaosc-core` because it's not used anywhere + introduces log4j vulnerabilities (see https://mvnrepository.com/artifact/com.illposed.osc/javaosc-core/0.8). Do we need it at all?~~
2. ~~Do we need `com.google.guava:guava` anywhere? Removing it from this project and `slime-java-commons` would reduce the binary size by 4-5mb and doesn't seem to affect anything.~~
3. ~~`org.slf4j:slf4j-simple` was added since I removed  `com.illposed.osc:javaosc-core` that was exposing logging via `slf4j`.~~
4. We still target java 8, while java 11 is our minimum. Should we change `targetCompatibility` and `sourceCompatibility`?
5. `org.yaml:snakeyaml` was removed since it is already provided by `slime-java-commons`.

Related PR: https://github.com/Eirenliel/slime-java-commons/pull/3


Eirenliel:
`com.illposed.osc` was planned to use for future OSC support. I don't think we are vulnerable to l4j, there aren't really any third-party data being processed?

deiteris:
Okay, then I'll add it back. We might want to keep an eye on this issue anyway.
https://github.com/hoijui/JavaOSC/issues/62

hoijui:
> Okay, then I'll add it back. We might want to keep an eye on this issue anyway. [hoijui/JavaOSC#62](https://github.com/hoijui/JavaOSC/issues/62)

It's fixed now (there is no more logging implementation in `com.illposed.osc:javaosc-core` anymore, as it should always have been, sorry :/ )

Eirenliel:
nice @hoijui 😄 we should probably bump our version to the latest..

hoijui:
ohh.. it is not released yet. would you need that?
I just released `0.9-SNAPSHOT`, for testing, if that works for you.

PR #121, opened 2022-02-09
Fix spelling mistakes

ButterscotchV:


src/main/java/dev/slimevr/bridge/WebSocketVRBridge.java
```diff
@@ -36,7 +36,7 @@
 	private final List<? extends ShareableTracker> shareTrackers;
 	private final List<ComputedTracker> internalTrackers;
 	
-	private final HMDTracker internalHMDTracker = new HMDTracker("itnernal://HMD");
+	private final HMDTracker internalHMDTracker = new HMDTracker("internal://HMD");
```
ButterscotchV, line 39: I believe this may be a breaking change, would need to double check this

Eirenliel: It's not.

PR #122, opened 2022-02-11
Change the minimum compatible Java version

Kamilake:
SlimeVR does not appear to be working normally on Java 8.
Inside the source code, there is a if state that limits the Java version to 11 or higher, resulting in the following error message: "SlimeVR start-up error! A minimum of Java 11 is required."

I think it is necessary to change the minimum Java version even in a build environment, and in a multi Java environment, the JDK to be executed at a low value is determined, so the server cannot be built in a low version. Therefore, it has been modified to 11 or higher. Will it be okay?


Eirenliel:
Should be fine, let's try :D

PR #126, opened 2022-02-16
Fix possible nullpo with ping packets

deiteris:
When there are lots of sensors on single connection, it may fail with NullPointerException. See https://discord.com/channels/817184208525983775/823930029070876736/943564649444561006


Eirenliel:
This should be handled by an iterator, it's just code that wasn't updated

deiteris:
Yeah, iterator would work better for this. If you already have a code, I'll just close this now.

PR #130, opened 2022-03-08
Switch feeder app from old protocol to new protocol.

kitlith:
I should have done this awhile ago. >_>

I'm not quite sure what's going on, but using this protocol, the role IDs don't seem to be used. I am sending the IDs used in TrackerRole.

### DIFF
--- a/src/main/java/dev/slimevr/VRServer.java
+++ b/src/main/java/dev/slimevr/VRServer.java
@@ -71,9 +71,10 @@ public class VRServer extends Thread {
 			bridges.add(driverBridge);
 			//*/
 			// Create named pipe bridge for SteamVR input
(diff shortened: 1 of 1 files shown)


kitlith:
I figured out what was going on, v0.2.1 of the feeder sends LEFT_CONTROLLER and RIGHT_CONTROLLER instead of LEFT_HAND and RIGHT_HAND to fix them not showing up.

Currently, I still send the hmd as HEAD instead of HMD since i'm not actually passing the hmdTracker to this bridge. I'm not sure how we want to handle the HMD with the feeder involved.

Eirenliel:
We need to test if feeder and driver give the same data, but probably driver would be more precise, because we need uncalibrated space? idk, needs more testing, but they shouldn't both be used at the same time either way.

PR #131, opened 2022-03-10
Change FORWARD to FRONT for tracker mounting

MightyGood:
Front fits better with the other options, and should cause less confusion about what the drop-down items mean.


Eirenliel:
yes

PR #134, opened 2022-03-13
Change default proportions and allow half-decimal editing

Louka3000:
- Better initial and default body proportions for people that don't want to calibrate
- Now changes bone length to half decimal (0.5). Rounds up to the nearest 0.5.


Louka3000:
It could be useful for head offset and neck length, considering those are small values that affect the whole skeleton’s sliding.
I could make a button to toggle 0.5 adjustments. A “Precision adjust” toggle…

PR #135, opened 2022-03-14
Prettier changing proportions

Louka3000:
- Removes the .0 if the value is whole (only see decimal if is .5)
- Add precision adjust button unselected by default to toggle between 1 and 0.5 adjustment.


Louka3000:
This is what it looks like owo https://cdn.discordapp.com/attachments/948579740225261598/953051764045787136/unknown.png

PR #137, opened 2022-03-23
Basic Interpolation and Extrapolation

Louka3000:
- Slerps quaternions
- Interpolation introduces delay but can smooth out movement
- Extrapolation may make movement more snappy and have less delay.
- Configurable settings.


src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
```diff
@@ -116,7 +154,12 @@ public boolean getPosition(Vector3f store) {
 	
 	@Override
 	public boolean getRotation(Quaternion store) {
-		store.set(rotQuaternion);
+		if(movementFilterTickCount > 0 && movementFilterAmount != 1 && previousRots.getFirst() != null){
+			store.set(new Quaternion().slerp(rotQuaternion, previousRots.getFirst(), movementFilterAmount));
```
ButterscotchV, line 158: Is there any way to reduce instantiating objects? Considering the tick count should be a fixed number, maybe this could be an array of quaternions that are looped over instead?

src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
```diff
@@ -101,6 +133,12 @@ public void tick() {
 				calculateLiveMagnetometerCorrection();
 			}
 		}
+		if(movementFilterTickCount != 0){
+			previousRots.addLast(new Quaternion(rotQuaternion));
```
ButterscotchV, line 137: Considering the tick count should be a fixed number, maybe this could be an array of quaternions that are looped over instead?

Louka3000: Like a circle buffer?

src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
Louka3000: Yeah true imma do that

src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
Louka3000: Also changed to Quaternion.clone here instead of creating a new Quaternion

ButterscotchV: Yeah, something like that, since it's happening hundreds of times per second, I think it'd be smart to try to keep allocation down a bit

src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
ButterscotchV: New change looks good 👍

src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
Louka3000: What do you think about that? http://www.museful.net/2012/software-development/circulararraylist-for-java

Louka3000: Alright. I'm using a circle buffer now. Performance should be better. _This is also what my dad said would be the most optimized for this kind of usage but I was too lazy to implement it._

Issue #61, opened 2021-10-03
Work without waist and chest trackers


Eirenliel:
Right now it works if you don't have waist or chest trackers, but it doesn't rotate at all. Yaw of waist and chest should match yaw of the headset if chest and waist trackers are missing probably. This is bad either way, but it will let test some stuff more easily without the need for full set of slimevr.

Legs with missing trackers should also match their yaw to the previous tracker in the chain.

Louka3000:
solved in https://github.com/SlimeVR/SlimeVR-Server/pull/138

PR #138, opened 2022-03-23
Align spine yaw with HMD when no spine trackers are present

Louka3000:
https://github.com/SlimeVR/SlimeVR-Server/issues/61


src/main/java/dev/slimevr/vr/processor/skeleton/SimpleSkeleton.java
```diff
@@ -407,17 +411,29 @@ protected void updateLocalTransforms() {
...
+			}
+			if(hipTracker.getRotation(rotBuf1)) {
+				waistNode.localTransform.setRotation(rotBuf1);
+				hipNode.localTransform.setRotation(rotBuf1);
+				trackerWaistNode.localTransform.setRotation(rotBuf1);
+			}
 		}
-		if(waistTracker.getRotation(rotBuf1)) {
+		else if(hmdTracker != null){ // If no spine tracker, allign spine yaw with HMD
+			rotBuf1 = new Quaternion().fromAngles(0, FastMath.atan2(rotBuf1.getY(), rotBuf1.getW())*2f, 0);
```
Eirenliel, line 430: This is completely wrong way to get a yaw...

Eirenliel (changes requested):

src/main/java/dev/slimevr/vr/processor/skeleton/SimpleSkeleton.java
Louka3000: Why? atan2(rotBuf1.getY(), rotBuf1.getW()) works...
How would you do it?

Eirenliel: There is a method `Quaternion.toAngles(float[3])` that sets provided array to [pitch, yaw, roll]. `atan2(rotBuf1.getY(), rotBuf1.getW())` doesn't cover all cases. Also, you're recreating rotBuf1, instead of just doing `rotBuf1.fromAngels(...)`.

Louka3000: Would you prefer `rotBuf1 = rotBuf1.fromAngles(0, rotBuf1.toAngles(null)[1], 0);` ?
I can do that but it just seems more expensive to compute each tick...

Louka3000: > There is a method `Quaternion.toAngles(float[3])` that sets provided array to [pitch, yaw, roll]. `atan2(rotBuf1.getY(), rotBuf1.getW())` doesn't cover all cases. Also, you're recreating rotBuf1, instead of just doing `rotBuf1.fromAngels(...)`.

I wrote the above message before seeing this btw. I'll do that 1 sec

Eirenliel: I would prefer `rotBuf1.fromAngles(0, rotBuf1.toAngles(anglesBuf)[1], 0);` and declare `angelsBuf` as a private static final to re-use. It's more expensive to calculate all angles, but is more correct. Even more correct would be to add a getYaw() method to Quaternion: https://github.com/Eirenliel/slime-java-commons/commit/94acd0dfaee0dca5ad51811c8f3bcf8facc964f2 :P

Louka3000: Alright, used GetYaw(). Should be good now (tested in SteamVR too)

On another note, should I open another PR in the fixYaw method to use GetYaw now that we have that?

It does `		targetTrackerRotation.toAngles(angles);
		targetTrackerRotation.fromAngles(0, angles[1], 0);`

Eirenliel: You can, but the calibration should use another method via vectors anyway (this too, but the math framework haven't been figured out yet)

PR #146, opened 2022-04-03
Fix ambiguous usage of "Legs"

Louka3000:
- Leg in tracker roles is now knee
- Legs in SteamVR trackers to enable is now Feet


Louka3000:
For tracker roles, this goes “knees, ankles, feet”. 

Could also do “Upper_leg, lower_leg, feet”. 

Can keep it like it currently is in this PR and handle what the user sees in the new GUI too.

Loler920a:
Reason: To prevent confusion when mounting physical trackers

PR #147, opened 2022-04-03
Fix "could not parse packet" warning for PingPongPackets

TheDevMinerTV:
Not sure why it was made this way, but whatever LOL

Not judging the code style but that's not how it works.


Eirenliel:
Because it used to be a list. "Not judging" but definetely judging...

PR #149, opened 2022-04-05
Hand tracking support + elbows without controllers

Louka3000:
This extends the HMD skeleton to support arm tracking. We do end up with kinda duplicate bones tho because of elbow tracking skeletons (from the controllers)

Basically: The HMD skeleton handles arm tracking and the Controller skeletons handle elbow tracking (even if the user is using arm tracking). ~~If the skeleton doesn't find a controller, it'll use the SlimeVR hand from the HMD skeleton.~~ It uses the elbow from the shoulders to hand chain.
This makes it so:

- If user has controllers: elbows go from controllers with tracker on lower arm and possibility for upper arm tracker.

- If user has no controller (or no feeder app) and enables Elbow SteamVR trackers: elbows go from chest/shoulders, using an upper arm tracker.

- If user has no controllers and enables Hand SteamVR trackers: SteamVR trackers on the users hand appear using upper arm, forearm and hand trackers. Use chest tracker as well for more precision!

This also adds a lot of body proportions. Very fun.

Main use of this is for mocap and vtubing, **not** VR.


Louka3000:
Video on hand tracking in action: https://cdn.discordapp.com/attachments/833895618450948106/960707193001349200/2022-04-04_20-54-18.mp4

Eirenliel:
[Hand tracking support](https://trello.com/c/Y4KpidUG/42-hand-tracking-support)

Eirenliel:
If the elbow tracker uses hand tracker to build its own chain, it's bad. Elbow tracker should use either Conroller -> Elbow chain or if controller doesn't exist, Shoulder -> Elbow.

Louka3000:
sure, so just change make the computed tracker use either one depending on if there’s a controller or not right :3

Eirenliel:
yes, @Louka3000!

Louka3000:
> yes, @Louka3000! that would make elbows more precise when using hand tracking, at least.

I don't think it actually makes it more precise because of the way I was doing it (all the offsets were the same back and forth basically), but it did make my code a whole lot better lol.
Should be good now :p

PR #151, opened 2022-04-05
Neck tracking support owo

Louka3000:
This is not a joke.
As I suspected, adding a neck tracker does help with reducing sliding. I tested it in VRChat myself.
Think of it like adding a fourth spine tracker.
Anyway, this can be useful to mocap suits and furries, pls merge ty owo.
Actual use case: https://discord.com/channels/817184208525983775/930205281609187418/960904944490266715

*Yes, the precision increase is* **very** *small and most people wouldn't be able to notice it, but existing mocap suits with trackers going on the back of the neck would benefit from it.*


Eirenliel:
[Neck tracking support](https://trello.com/c/3S3r1ZOK/41-neck-tracking-support)

PR #152, opened 2022-04-05
Add full reset command to WebSocket bridge

mgschwan:
Players should be able to do a full reset from within the headset without going through the GUI first.
This is required for the headless standalone server on the Quest.

There is no delay before initiating the reset like in the main GUI because this is the responsibility of the app calling for the reset


TheDevMinerTV:
I think you should rather help in the protocol, since that will probably be used in the WebSocketVRBridge as well (we have a thread for this in the `development` channel called `new-network-protocol`)

Eirenliel:
> I think you should rather help in the protocol, since that will probably be used in the WebSocketVRBridge as well (we have a thread for this in the `development` channel called `new-network-protocol`)

There is enough place for both :P

PR #154, opened 2022-04-11
Make WebSocketVRBridge compatible with Android

mgschwan:
Change JSONObject.optFloat to JSONObject.optDouble because optFloat does not exist anymore on recent Android versions

The timeout should also be removed. At least in my tests it did not work reliably and dropped the connection, but to remain compatible with the desktop version it's left in with a comment until it's determined safe to remove or the Android issues are resolved through other means.


Eirenliel:
it used to use a fork of JSON library where i added optFloat myself

PR #156, opened 2022-04-12
Added overlay_protocol

TheButlah:


TheButlah:
@Eirenliel before I continue, just wanted you to give this a quick glance over. I decided that it might be a bad idea to block the overlay on the advanced union support for flatc-rust, so I introduced a temporary protocol in the meantime. The flatbuffer schema is finished I think, but the `OverlayBridge` is still WIP.

Eirenliel:
> all checks have failed

idk i'm not to fond of the idea of pusing something temporarry in. there is nothing more permanent than temporarry. overlay is not the blocking feature for reviews, imo.

Eirenliel:
Okay then I don't mind.

loucass003:
Is this still needed ? Shoud we close it. @TheButlah ?

TheButlah:
No longer needed now that the main protocol is compatible with rust

PR #159, opened 2022-04-24
Headless Server / SolarXR Protocol implementation

loucass003:


.gitmodules
```diff
@@ -1,3 +1,9 @@
 [submodule "slime-java-commons"]
 	path = slime-java-commons
 	url = https://github.com/Eirenliel/slime-java-commons.git
+[submodule "slimevr_protocol"]
```
Eirenliel: Shouldn't you remove this after including solarxr?

.settings/org.eclipse.buildship.core.prefs
```diff
@@ -1,2 +1,13 @@
+arguments=
+auto.sync=false
+build.scans.enabled=false
+connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)
 connection.project.dir=
 eclipse.preferences.version=1
+gradle.user.home=
+java.home=C\:/Program Files/Java/jdk-17.0.1
```
Eirenliel, line 8: That shouldn't be here. Hava home path can't be project-specific settings, they can be different for each user.

log_last.log.1
```diff
@@ -0,0 +1,2 @@
+2022-03-24 21:58:25 [SEVERE] SlimeVR start-up error! Required ports are busy. Make sure there is no other instance of SlimeVR Server running.
```
Eirenliel: Logs in git, please add it to .gitignore. But also just don't run from the project folder to avoid the project pollution 😛...

log_main.log.1
```diff
@@ -0,0 +1,2 @@
+2022-03-24 21:58:25 [SEVERE] SlimeVR start-up error! Required ports are busy. Make sure there is no other instance of SlimeVR Server running.
```
Eirenliel: Ignore!

Eirenliel (changes requested):
This is only the start of the review

PR #163, opened 2022-04-30
Fix iteration on the sensors, i used the key of the map instead of the index of the array

loucass003:


Eirenliel:
why not
```Java
                DeviceData.startTrackersVector(fbb, device.sensors.size());
		device.sensors.forEach((key, value) -> {
			DeviceData.addTrackers(DataFeedBuilder.createTrackerData(fbb, mask.getTrackerData(), value));
		});
		return fbb.endVector();
```

loucass003:
> why not
> 
> ```java
>                 DeviceData.startTrackersVector(fbb, device.sensors.size());
> 		device.sensors.forEach((key, value) -> {
> 			DeviceData.addTrackers(DataFeedBuilder.createTrackerData(fbb, mask.getTrackerData(), value));
> 		});
> 		return fbb.endVector();
> ```

it wont work because of how flatbuffers work, you cannot do nested creation of table.

in that specific case you would get the following error: ``FlatBuffers: object serialization must not be nested.``

PR #169, opened 2022-05-02
.gitignore eclipse stuff

TheButlah:
The eclipse-specific project settings conflict with the Java VSCode extension, causing the files to get changed. To me this is the fault of vscode, but it would be nice if we didn't commit the eclipse stuff in.

Does removing it cause any problems for eclipse users?


Eirenliel:
Do not ignore `.settings` files, they're required for formatting and checks. We discussed it before, only `.settings` and `.classpath` need to be ignored.

TheButlah:
OK, I took a look at the `.settings` files and determined that only one of them is related to formatting. I've gone ahead and .gitignored the `.classpath` and `.project` files, as well as all `.settings` that are unrelated to formatting

PR #172, opened 2022-05-03
Added eclipse autoformatter and CONTRIBUTING.md

TheButlah:
I've been attempting to find an autoformatter that hooks into gradle for us to use. Unfortunately the most highly praised one, [google-java-format](https://github.com/google/google-java-format) offers no ability to use tabs instead of spaces, and I know you said that was important to you.

As an alternative, I found that [spotless](https://github.com/diffplug/spotless) can hook into eclipse's autoformatter and format the files that way. Spotless also provides an IntelliJ and VSCode integration. So I'm pretty sure we can support all 3 java editors out of the box (and any others can just use gradle from the CLI directly).

This requries a `spotless.eclipseformat.xml` file for us to use as the source of truth for the autoformatting. @Eirenliel could you please follow the instructions [here](https://github.com/diffplug/spotless/blob/main/ECLIPSE_SCREENSHOTS.md) to generate a file according to your preference? I can then update this PR with that file.


Eirenliel:
Why would I generate it according to *my* preference, when y'all bullied me into adopting your .editorconfig? It's not the source of truth, not my formatting.

TheButlah:
@Eirenliel I have updated the PR with the latest stuff from your branch.

I deleted the eclipse editor prefs, because I would like the single source of truth to be `spotless.xml`, and you can easily import this into eclipse. Also because the VSCode java plugin messes up if it sees a `.settings` folder with eclipse stuff.

Louka3000:
This works just fine, but VSCode warns me that "There is no valid JAVA_HOME setting to launch Gradle Language Server."

https://cdn.discordapp.com/attachments/818062281636053012/976697150052368424/unknown.png

PR #173, opened 2022-05-03
Computed trackers use TrackerRole to determine TrackerPosition, if null

TheButlah:
This will be important in the overlay. I think we need a larger, separate effort to shift to using `TrackerPosition` instead of `TrackerRole` in most of the codebase and treat `TrackerRole` as a steamvr-specific thing, but we can do that later


Eirenliel:
I had a mapping from tracker positions to tracker roles... Idea is that we support way more positions than SteamVR for future-proofing.

PR #162, opened 2022-04-27
Format all files based on .editorconfig

loucass003:

PR #176, opened 2022-05-08
Eclipse formatting

Eirenliel:
Make eclipse formatter better follow our formatting conventions (#162). Format everything using it, mostly enforcing new lines.


src/main/java/dev/slimevr/NetworkProtocol.java
```diff
@@ -1,8 +1,5 @@
 package dev.slimevr;
 
 public enum NetworkProtocol {
-	OWO_LEGACY,
-	SLIMEVR_RAW,
-	SLIMEVR_FLATBUFFER,
-	SLIMEVR_WEBSOCKET
+	OWO_LEGACY, SLIMEVR_RAW, SLIMEVR_FLATBUFFER, SLIMEVR_WEBSOCKET
```
TheDevMinerTV, line 4: pretty sure that this isn't intended

src/main/java/dev/slimevr/bridge/PipeState.java
```diff
@@ -1,7 +1,5 @@
 package dev.slimevr.bridge;
 
 public enum PipeState {
-	CREATED,
-	OPEN,
-	ERROR
+	CREATED, OPEN, ERROR
```
TheDevMinerTV, line 4: same here

src/main/java/dev/slimevr/gui/SkeletonConfigGUI.java
```diff
@@ -114,13 +66,13 @@ public void mouseClicked(MouseEvent e) {
...
 				// Only use a timer on configs that need time to get into position for
 				switch (config) {
-					case TORSO:
-					case LEGS_LENGTH:
-						add(new TimedResetButton("Reset", config), c(4, row, 2));
-						break;
-					default:
-						add(new ResetButton("Reset", config), c(4, row, 2));
-						break;
+				case TORSO:
```
TheDevMinerTV, line 69: pretty sure that this should be indentet one more to the right

TheDevMinerTV (changes requested):
only looked over it quickly, but I found these things

src/main/java/dev/slimevr/bridge/PipeState.java
Eirenliel: Eclipse always done it, if there are no arguments it joins the lines in enums

TheButlah:
checking on this now, will post my thoughts

TheButlah:
PR doesn't appear to compile rn, were some non-formatting changes introduced on accident?

TheButlah (commented):
This PR doesn't include the changes from #172, so I cannot really test if the format matches what is described in the eclipse settings. So for now, I'll just assume that what I see is what the format actually describes.

Another thing to mention: We should just pick an existing eclipse configuration and just slightly tweak it, rather than come up with our own entirely custom from scratch configuration. Otherwise its going to be a ton of bike-shedding while we decide on a format

Still going through and checking the format, have several critiques but I'm worried about getting into bike shedding territory and don't want this PR to devolve into that

Eirenliel:
> PR doesn't appear to compile rn, were some non-formatting changes introduced on accident? Looks like the LogManager stuff?

Don't forget to update submodules. All checks have passed, so it's fine.

Eirenliel:
> Another thing to mention: We should just pick an existing eclipse configuration and just slightly tweak it, rather than come up with our own entirely custom from scratch configuration. Otherwise its going to be a ton of bike-shedding while we decide on a format

That's the built-in Java configuration except it's tabs only instead of mixed and spaces removed in array declarations. So it's pretty much java standard from eclipse's point of view

TheButlah:
> So it's pretty much java standard from eclipse's point of view
Ok, good to know! Thanks for telling me :)

> Don't forget to update submodules
🤦

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -131,7 +132,8 @@ protected void positionReceived(Position positionMessage) {
 		if (tracker != null) {
 			if (positionMessage.hasX())
 				tracker.position.set(positionMessage.getX(), positionMessage.getY(), positionMessage.getZ());
-			tracker.rotation.set(positionMessage.getQx(), positionMessage.getQy(), positionMessage.getQz(), positionMessage.getQw());
+			tracker.rotation.set(positionMessage.getQx(), positionMessage.getQy(), positionMessage.getQz(),
```
TheButlah, line 135: We should decide whether we want the behavior to be "fill to max col limit then wrap to newline", or "if there is a need to wrap, break every argument into its own newline".

I prefer the latter, as it is what `black` and `cargo fmt` use and I find it looks elegant, and reduces diff noise. The current formatting uses the former. What are y'alls thoughts?

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -188,7 +190,9 @@ protected T getInternalRemoteTrackerById(int trackerId) {
 	@VRServerThread
 	protected void reconnected() {
 		for (ShareableTracker tracker : sharedTrackers) {
```
TheButlah, line 192: Looks like the autoformatter is not forcing tab based indentation, but rather taking whatever indentation style is used in the current file. In this case, this file looks like its using 2 spaces instead of tabs.

IMO the autoformatter should force whatever indentation style that is used to be consistent across all files, not just within a file

src/main/java/dev/slimevr/gui/AutoBoneWindow.java
```diff
@@ -79,12 +80,14 @@ public void run() {
...
 											if (poseRecorder.isReadyToRecord()) {
 												setText("Recording...");
-												// 1000 samples at 20 ms per sample is 20 seconds
+												// 1000 samples at 20 ms per sample is
+												// 20 seconds
 												int sampleCount = server.config.getInt("autobone.sampleCount", 1000);
 												long sampleRate = server.config.getLong("autobone.sampleRateMs", 20L);
-												Future<PoseFrames> framesFuture = poseRecorder.startFrameRecording(sampleCount, sampleRate);
+												Future<PoseFrames> framesFuture = poseRecorder
+														.startFrameRecording(sampleCount, sampleRate);
```
TheButlah, line 88: Really not a fan of the double-indent when lines are broken up for wrapping. Can we change it to single indent? Unless I'm missing something I don't see the reason for it. Its a lot more clear in this tab-indented file why it is an extravagant use of whitespace (the files prior to this are two-space indented)

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -212,14 +216,17 @@ public void addSharedTracker(ShareableTracker tracker) {
 		if (sharedTrackers.contains(tracker))
 			return;
 		sharedTrackers.add(tracker);
-		TrackerAdded.Builder builder = TrackerAdded.newBuilder().setTrackerId(tracker.getTrackerId()).setTrackerName(tracker.getDescriptiveName()).setTrackerSerial(tracker.getName()).setTrackerRole(tracker.getTrackerRole().id);
+		TrackerAdded.Builder builder = TrackerAdded.newBuilder().setTrackerId(tracker.getTrackerId())
+				.setTrackerName(tracker.getDescriptiveName()).setTrackerSerial(tracker.getName())
```
TheButlah, line 220: See prior comment about newline wrapping behavior, this is another example but this time with chained function calls instead of function arguments. I would prefer that we do a newline per chained call when needing to wrap lines,  but I'd like to hear y'alls opinion.

src/main/java/dev/slimevr/protocol/DataFeedBuilder.java
```diff
@@ -176,8 +163,10 @@ public static int createDeviceData(FlatBufferBuilder fbb, int id, DeviceDataMask
 		return DeviceData.endDeviceData(fbb);
 	}
 
-	public static int createSyntheticTrackersData(FlatBufferBuilder fbb, TrackerDataMaskT trackerDataMaskT, List<Tracker> trackers) {
-		if (trackerDataMaskT == null) return 0;
+	public static int createSyntheticTrackersData(FlatBufferBuilder fbb, TrackerDataMaskT trackerDataMaskT,
```
TheButlah, line 166: Same thing as my prior comment about newline behavior on function chaining and arguments. In this case, I would prefer that when a wrap is necessary, each parameter gets its own newline.

.settings/org.eclipse.jdt.core.prefs
```diff
@@ -387,14 +454,37 @@ org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_method_decla
...
+org.eclipse.jdt.core.formatter.keep_simple_do_while_body_on_same_line=false
+org.eclipse.jdt.core.formatter.keep_simple_for_body_on_same_line=false
+org.eclipse.jdt.core.formatter.keep_simple_getter_setter_on_one_line=false
+org.eclipse.jdt.core.formatter.keep_simple_while_body_on_same_line=false
+org.eclipse.jdt.core.formatter.keep_switch_body_block_on_one_line=one_line_never
+org.eclipse.jdt.core.formatter.keep_switch_case_with_arrow_on_one_line=one_line_never
 org.eclipse.jdt.core.formatter.keep_then_statement_on_same_line=false
-org.eclipse.jdt.core.formatter.lineSplit=200
+org.eclipse.jdt.core.formatter.keep_type_declaration_on_one_line=one_line_never
+org.eclipse.jdt.core.formatter.lineSplit=120
```
TheButlah, line 479: I would strongly prefer a column limit of 100 or less, for these reasons:

* Code reads better when it is constrained to a fairly small column limit as your eyes can simply scan down the page.
* 100 is already a generous limit, 120 starts to defeat the point of a column limit to begin with. Python's `black` only permits 88, 80 is the standard in the linux kernel, `cargo fmt` uses 100 by default, and google's java code style uses 100. To my knowledge, *no one* uses 120, at that point they just don't enforce a limit.
* Lets all code fit on the screen without the need to scroll right
* Important when editing from command line, either in `git diff` or in `vim`
* Important for those of us with only 1 monitor (like me 😢 )
* Important for viewing two files side by side, or having web browser + ide side by side
* Important for those that use vertical monitors (very common)

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
TheButlah: Wait what? Checking this in the comment view shows it as a tab, but checking it in the "changes" view on github shows it as two spaces. Whats going on?

TheButlah:
Overall, format looks decent, but I favor the newline-per-overflowed-thing rather than "make max use of line until col limit"

Also, the column limit comment I left is very important to me 😅 

I also have some concerns about the heavily indented nature of some parts of the codebase (i.e. the JavaFX code block stuff). That might just be a problem with how that part of the code is written though, and not an autoformatter problem. Can probably be done slightly differently to behave more nicely with the autoformatter. Will need to experiment

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
TheButlah: OK, checking out the branch makes it clear that its actually tabs, so I guess that resolves this comment. Super weird it didn't display that way on github.

Feel free to mark as resolved @Eirenliel

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
Eirenliel: That's not possible in Eclipse formatter, I discussed it in discord. It can be formatted by hand like this or a post-processor can do this.

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
Eirenliel: Same answer as the previous comment.

src/main/java/dev/slimevr/bridge/ProtobufMessages.java
```diff
@@ -4,55 +4,32 @@
 package dev.slimevr.bridge;
 
 public final class ProtobufMessages {
```
Eirenliel, line 6: That shouldn't be formatted actually, it's auto-generated...

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
TheButlah: Its actually possible, it depends on the wrapping mode. You can see an example here, in this case one of the new() calls is not wrapped because it doesn't hit the 100 character limit, and the other is wrapped, with each argument on a new line:

image

You can control this behavior in the GUI by clicking on this:
image

src/main/java/dev/slimevr/gui/AutoBoneWindow.java
TheButlah: Update: Yes its totally possible to control the indent level, I confirmed this in the eclipse settings:
image

TheButlah:
@Eirenliel please try this formatter configuration:
[spotless.zip](https://github.com/SlimeVR/SlimeVR-Server/files/8649393/spotless.zip)
 
To test it out, import it like so:
image

Key changes:
* I made closing parens go to a newline if they wrap. This prevents ambiguity in wrapped lines and their following lines like the following scenario:
image
image
In the absence of a wrap, curly braces and parens are still all in one line.

* I made binary operators wrap:
image
image

* Column limit set to 100
* Indentation on wrap is set to 1 indent instead of 2
* Insert final newline in each file (helps git diffs not mess up)

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
Eirenliel: That's what I did and it wrapped EVERYTHING
img

src/main/java/dev/slimevr/gui/AutoBoneWindow.java
Eirenliel: I know you can, idk if it looks better though with single indent.

.settings/org.eclipse.jdt.core.prefs
Eirenliel: > Important for those of us with only 1 monitor (like me 😢 )

Does your monitor only fit 80 charaters?

> Code reads better when it is constrained to a fairly small column limit as your eyes can simply scan down the page.

IMO, code reads better when you don't have 10 lines that are effectively one line of code. I agreed to line limit at all because you literally forcing me, but I disagree with this reasoning. In 95% of code reading you don't care for arguments of a function call, only the call itself. That's at least how I read code.

Eirenliel: I put 120 because .editroconfig has 120, and all point of this is to match it. You should've said this all in #162...

TheButlah: I did say this in discord

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
TheButlah: Please check my configuration, it should handle this better

.settings/org.eclipse.jdt.core.prefs
TheButlah: My monitor looks like this generally. I have marked out the 80, 100, and 20 character column limits on the vertical rulers so you can see them clearly. I have a fairly widescreen monitor mind you, and its 32 inches so its quite large. I believe this to be a best case scenario of someone with one monitor. Note that IDE annotations make this *even longer* than 120.

No other windows or code, everything fits on the screen (but don't disregard my statements about readability of vertical oriented code)
image
image

From here on out, things don't fit at 120 characters. For example, side-by-side editing of code (extremely common, I reference flatbuffer schema while working on java).
image
image

Github reviews/documentation reference/whatever else you use a web browser for.
image
image

Maybe its my fault for not resizing my windows 🤔 Lets see how small I need to make my browser to get it to fit
image
image

TheButlah: I hope this demonstrated that the concerns I'm raising are not strange edge cases that no one runs into. I use these development workflows routinely (I never have my IDE in fullscreen mode, because I'm always looking at documentation, or on discord, or looking at github at the same time)

PR #186, opened 2022-05-29
Adding Serial RTS / DTR control elements

unlogisch04:
Adding control elements for RTS / DTR line.
Allow a bigger range of D1 mini dev boards to be used with the WiFi window.
See:
https://discord.com/channels/817184208525983775/878727840118505533/977740333955821599


Eirenliel:
Please add some comments to requests. Including why the user would need it, a use-case for users, otherwise it's extra buttons ppl will get confused by, imo.

unlogisch04:
waiting for https://github.com/SlimeVR/SolarXR-Protocol/pull/21

PR #189, opened 2022-06-01
Skeleton data

Eirenliel:
Many changes to skeleton from different contributors.

* Force our naming conventons for bones
  * Sync with protocol repo
* Rename some classes to be more understandable
* Remove some unused stuff
* Added extended spine model by @Louka3000 
  * Sync with commons repo
* Added method to export current bone structure from skeleton for use in overlay


TheButlah:
Will review tomorrow after work, most likely

src/main/java/dev/slimevr/vr/trackers/TrackerPosition.java
```diff
@@ -103,12 +102,26 @@ public static Optional<TrackerPosition> getByDesignation(String designation) {
...
+			designation = "body:right_upper_leg";
+		} else if (designation.equalsIgnoreCase("body:left_ankle")) {
+			designation = "body:left_lower_leg";
+		} else if (designation.equalsIgnoreCase("body:right_ankle")) {
+			designation = "body:right_lower_leg";
+		} else if (designation.equalsIgnoreCase("body:left_forearm")) {
+			designation = "body:left_lower_arm";
+		} else if (designation.equalsIgnoreCase("body:right_forearm")) {
+			designation = "body:right_lower_arm";
+		}
```
Louka3000, line 124: Just realized I forgot a check here.
Need to check for "body:left_upperarm" and replace it by "body:left_upper:arm" (same with right!!!)

Louka3000 (changes requested):

src/main/java/dev/slimevr/vr/trackers/TrackerPosition.java
Louka3000: Not many people used the upperarm, since it's only in early, beta, builds.
What would be even better would be handling when `designation` doesn't correspond to any TrackerPosition. Currently it just throws an exception and kinda breaks IIRC.

Louka3000: Forgot we can push everywhere except main so I fixed it :3 
Altho handling when `designation` doesn't correspond to anything would be a good idea... I just checked it and it throws an exception and prevents any more trackers from appearing in GUI.

src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
```diff
@@ -164,10 +170,20 @@ public class HumanSkeleton extends Skeleton implements SkeletonConfigCallback {
 	protected ComputedHumanPoseTracker computedRightElbowTracker;
 	protected ComputedHumanPoseTracker computedLeftHandTracker;
 	protected ComputedHumanPoseTracker computedRightHandTracker;
+	// #endregion
+
+	// #region FK Settings
```
TheButlah, line 175: I would prefer if settings like this are extracted out into a separate config class instead of being fields inline with the class they configure, because it helps reduce the large number of fields in this class. But thats out of scope of this PR.

Feel free to mark as resolved when you've read @Eirenliel

src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
```diff
@@ -202,8 +218,8 @@ protected HumanSkeleton(List<? extends ComputedHumanPoseTracker> computedTracker
 		// #endregion
 
 		// #region Assemble skeleton arms from chest
```
TheButlah, line 220:
```suggestion
		// #region Assemble skeleton arms from neck
```

src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
```diff
@@ -202,8 +218,8 @@ protected HumanSkeleton(List<? extends ComputedHumanPoseTracker> computedTracker
...
-		chestNode.attachChild(leftShoulderNodeHmd);
-		chestNode.attachChild(rightShoulderNodeHmd);
+		neckNode.attachChild(leftShoulderNodeHmd);
```
TheButlah, line 221: Just confirming that this makes sense @Louka3000?

src/main/java/dev/slimevr/vr/trackers/TrackerRole.java
```diff
@@ -1,13 +1,11 @@
...
 /**
- * The SteamVR tracker role.
- *
  * The tracker role classifies the position and the role of a tracker on user's
- * body or playspace (like CAMERA or BEACON), using SteamVR naming. Tracker
- * roles are hints for interacting programs what the tracker means, and they do
- * not correspond to body's bones on purpose. Example: virtual vive trackers for
- * SteamVR vs actual SlimeVR trackers.
+ * body or playspace (like CAMERA or BEACON). Tracker roles are hints for
+ * interacting programs what the tracker means, and they do not correspond to
```
TheButlah, line 6: I think we are accidentally overwriting the updated comment here, should be the one on the left

TheButlah (changes requested):
I could not review the actual skeleton logic btw, but everything else I reviewed

src/main/java/dev/slimevr/poserecorder/PoseFrames.java
```diff
@@ -138,7 +138,8 @@ public float getMaxHmdHeight() {
 	public float getMaxHeight(TrackerPosition trackerPosition) {
 		float maxHeight = 0f;
 
-		PoseFrameTracker hmd = TrackerUtils.findTrackerForBodyPosition(trackers, trackerPosition);
+		PoseFrameTracker hmd = TrackerUtils
+			.findNonComputedHumanPoseTrackerForBodyPosition(trackers, trackerPosition);
```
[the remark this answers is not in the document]
Eirenliel: It's not, and I warned you that there shouldn't be a distinction before...

src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
```diff
@@ -121,32 +121,38 @@ public class HumanSkeleton extends Skeleton implements SkeletonConfigCallback {
 		"Right-Hand-Tracker-Hmd",
 		false
 	);
-	protected final Vector3f hipVector = new Vector3f();
-	protected final Vector3f ankleVector = new Vector3f();
 	// #endregion
 	protected final Quaternion kneeRotation = new Quaternion();
 	// #region Buffers
 	private final Vector3f posBuf = new Vector3f();
+	private final Quaternion rotBuf1 = new Quaternion();
```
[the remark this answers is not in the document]
Eirenliel: They're buffers, correct, and that's what we use everywhere, our naming conventions. Some comments would be fine, I guess.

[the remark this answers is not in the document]
Eirenliel: It won't be more efficient, no.

src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
Louka3000: Yes, `neckNode` is controlled by the physical chest tracker whereas `chestNode` is controlled by the waist tracker (or whichever one replaces it).
I made the adjustments in the bone length config (whatever it’s called) as well)

src/main/java/dev/slimevr/vr/trackers/TrackerRole.java
Eirenliel: That seem to be some github shenanigans showing diff, it's as it should be in the branch right now.

src/main/java/dev/slimevr/vr/trackers/TrackerPosition.java
Eirenliel: If you fixed it, then approve the merge :nya_a:

Louka3000 (approved):

build.gradle
```diff
@@ -100,7 +100,6 @@ spotless {
 	// 	indentWithSpaces(2)  // YAML cannot contain tabs: https://yaml.org/faq.html
 	// }
 	java {
-		removeUnusedImports()
```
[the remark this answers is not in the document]
TheButlah: can you post a screenshot? I've never seen that. I did test this and confirm that unused imports were no longer being removed, once we remove the line in question.

TheButlah (approved):
LGTM, after it gets rebased. I had some unresolved questions about the computed/noncomputed stuff, which I've posted in server channel. Also a review workflow question which I've posted in project management channel. Neither of these prohibit merge tho.

TheButlah (changes requested):
Sorry, forgot that the removeUnusedImport is still an issue

build.gradle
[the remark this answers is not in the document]
Eirenliel: I returned it, it works fine after initialization, but i had a problem when initializing. If i encouter it again, i'll open a separate issue. It might be the problem with commons actually...

TheButlah (approved):

PR #191, opened 2022-06-06
Fix Tracker role optional value

loucass003:


Eirenliel (approved):

TheButlah:
Not able to review since I'm so unfamiliar with the code, deferring to @Eirenliel

Eirenliel:
You can merge when you rebase.

Louka3000:
There's no more option for NONE with all these changes smh https://cdn.discordapp.com/attachments/695087258029522944/983517213887721492/unknown.png

Is that something the new GUI will handle?

loucass003:
Interesting, didnt saw that. For the new gui it fine because i treat the value 0 as none and offset everyting by one. But for the old gui this is wierd indeed. None was not in the enum anyway it was an option aded in the gui. I guess it got removed when refactoring which is not good.

Eirenliel:
That's why I said we need NONE option -__-

loucass003:
> That's why I said we need NONE option -__-

No we dont need the NONE option. Either we use null or optional but we dont need the none option in the enum
I have fixed this in a upcoming PR.
I am gonna close this pr as i have a new one comming soon that address both optionals + the missing none option

Eirenliel:
Yes.

PR #194, opened 2022-06-10
Fix optional in the feeder and bring back none option in the gui

loucass003:


Louka3000 (approved):
This indeed fixes NONE in the old GUI and doesn't break when using the feeder app.

Eirenliel (approved):

PR #206, opened 2022-06-29
Fixing skeleton stuff

Louka3000:
- Added `forceArmsFromHMD `to `SkeletonConfigToggle`.
- Changed `SkeletonConfigToggle`'s config prefix from "body.extendedModel." to "skeleton.toggles", as I want to separate FK settings from body proportions. *(yes this breaks backwards compatibility, but the only way to change the values is manually in vrconfig.yml and I doubt anyone would :p)*
- Make shoulders independent from the upper and lower arms in `UpdateLocalTransforms`, and fix them not moving properly when the user has no arm tracker. This is for easier debugging.
- Bring back the bones @TheButlah deleted, but don't send them if `sendAllBones == false` (which it is hardcoded to be currently)
- Properly don't send certain bones when only using a subset of trackers. I tested in butlah's overlay and it works! This allows, for example, to only see the spine overlays when only using spine trackers and no legs.

### DIFF
--- a/src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
+++ b/src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
@@ -131,6 +131,10 @@ public class HumanSkeleton extends Skeleton implements SkeletonConfigCallback {
 	private final Quaternion rotBuf4 = new Quaternion();
 	protected boolean hasSpineTracker;
 	protected boolean hasKneeTrackers;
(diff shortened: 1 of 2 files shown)


TheButlah:
Did you manage to test this with the overlay yet?

Louka3000:
> Did you manage to test this with the overlay yet?

yes, 
> I tested in butlah's overlay and it works!

Eirenliel (approved):

PR #214, opened 2022-07-13
Skeleton arms rework and cleaning up

Louka3000:
- Removed duplicated arms. Now we have a single pair of arms and detach them using `detachCompletely()` and reasign them to a different parent when we want to switch between controller arm-tracking and HMD arm-tracking.
This should make developing with arms much easier, as we only have 1 pair of arms to use, and can call `public boolean isTracking(Left/Right)ArmFromController()` to know how to make it behave. Probably should add those 2 booleans to the protocol to be read eventually. This is needed for stuff like the SlimeVR overlay to display the arms properly without having to add duplicate BoneTypes.
- Moved fancy yaw/roll rotation stuff to methods.
- Removed `updatePoseAffectedByConfig`, `setHmd` and `onlyFillWaistAndFeet`, as they didn't seem to have any real use.


src/main/java/dev/slimevr/vr/processor/TransformNode.java
```diff
@@ -78,4 +78,15 @@ public void combineWithParentGlobalRotation(Transform parent) {
 			.multLocal(worldTransform.getTranslation())
 			.addLocal(parent.getTranslation());
 	}
+
+	public void detachCompletely() {
```
Eirenliel, line 82: Please rename to something like `detachWithChildren()`

Eirenliel (approved):

Eirenliel:
Awesome refactoring, needs minor change.

src/main/java/dev/slimevr/vr/processor/TransformNode.java
Louka3000: how about `detachParentAndChildren()`?

Eirenliel: It doesn't detaches parent, it detaches from parent though

Eirenliel (approved):

PR #216, opened 2022-07-13
Add open BVH folder button

Kamilake:
I made a open BVH folder button because I often wander around looking for BVH directory. Will it be okay?


Eirenliel:
Sorry, this is an edit for old GUI and the button name is not clear what it does. BVH will have its own place in new GUI, and this should be added there.

PR #227, opened 2022-08-05
Config refactor round 1

loucass003:
This is the first step to a better config code. What i did here is mostly move all the config code into its own files so we can iterate/improve on that. 

My endgoal would be to move to someting like snakeyaml (similar to Gson) that allow class based config.


src/main/java/dev/slimevr/config/ConfigManager.java
```diff
@@ -0,0 +1,91 @@
...
+		this.trackersConfig.loadConfig();
+		this.autoboneConfig.loadConfig();
+		this.filteringConfig.loadConfig();
+		this.windowConfig.loadConfig();
+		this.bridgeConfig.loadConfig();
+	}
+
+	@ThreadSafe
+	public synchronized void saveConfig() {
+		this.trackersConfig.saveConfig();
```
ButterscotchV, line 55: Why is this saving the tracker config? If there's a good reason then maybe it can be added as a comment, then this can be marked as resolved

src/main/java/dev/slimevr/config/FilteringConfig.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.HashMap;
+
+
+public class FilteringConfig {
+
+	private final ConfigManager configManager;
+
+	public static String CONFIG_ROOT = "filters";
+	public static float DEFAULT_INTENSITY = 0.3f;
+	public static int DEFAULT_TICK = 1;
```
ButterscotchV, line 18: Do we need these? Are these useful to have anymore now that the default can just be on the field?

src/main/java/dev/slimevr/config/FilteringConfig.java
```diff
@@ -0,0 +1,86 @@
...
+			.getInt("tickCount", DEFAULT_TICK);
+	}
+
+	public void updateTrackersFilters(TrackerFilters filter, float amount, int ticks) {
+		setType(filter.name());
+		setAmount(amount);
+		setTicks(ticks);
+
+		IMUTracker imu;
+		for (Tracker t : Main.vrServer.getAllTrackers()) {
```
ButterscotchV, line 51: This is some weird class dependency, unsure if this can be resolved, but it's not the best IMO

src/main/java/dev/slimevr/config/TrackersConfig.java
```diff
@@ -0,0 +1,85 @@
...
+					if (tc.trackerName.equals(c.get("name"))) {
+						cfg = c;
+						break;
+					}
+				}
+				if (cfg == null) {
+					cfg = new HashMap<>();
+					trackersConfig.add(cfg);
+				}
+				tc.saveConfig(new YamlNode(cfg));
```
ButterscotchV, line 58: This seems to be a chunk of code from somewhere, but the variable names could be a little more descriptive

src/main/java/dev/slimevr/vr/processor/skeleton/SkeletonConfigToggle.java
```diff
@@ -0,0 +1,38 @@
+package dev.slimevr.vr.processor.skeleton;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+public enum SkeletonConfigToggle {
```
ButterscotchV, line 7: This enum is duplicated, please look at `SkeletonConfigToggles.java`
https://github.com/SlimeVR/SlimeVR-Server/blob/main/src/main/java/dev/slimevr/vr/processor/skeleton/SkeletonConfigToggles.java

ButterscotchV (changes requested):
Please look into the problem with `SkeletonConfigToggles.java` being duplicated as `SkeletonConfigToggle.java` with this PR

src/main/java/dev/slimevr/config/ConfigManager.java
loucass003: https://github.com/SlimeVR/SlimeVR-Server/pull/227/files#diff-26e9670613ebf70e6487d7401ad4cf394b1a64fe102f0f0520f59ce8084b5a29L209-L232

This is what the previous saveConfig function used to do. So i did the same to keep the same execution

src/main/java/dev/slimevr/config/FilteringConfig.java
loucass003: No, y ou still need a default value because the filtering ils always reading a value

src/main/java/dev/slimevr/config/FilteringConfig.java
loucass003: Wdym by wierd class dependency

src/main/java/dev/slimevr/config/TrackersConfig.java
loucass003: This is was in the VRServer.class inside the saveConfig function. I moved it as is

src/main/java/dev/slimevr/config/FilteringConfig.java
ButterscotchV: Depending on a static insurance of the VR server when everything else takes in an instance of it, but I guess it's fine

ButterscotchV (approved):
Looks good, I won't hold it back over some old code looking a little weird

loucass003:
Already obselete Sorry @ButterscotchV <3

PR #228, opened 2022-08-06
Make elbows track from HMD by default

Loler920a:
Make elbows track from HMD by default
It appears alot of users are put off elbow trackers due to their first initial impressions and seem to be unaware of the alternate solution which would grant them more stability. 

This is due to how inside out tracking works, and how most of our users would likely be quest 2 users.

From Controllers:
+ (Pro) Can do shoulders without having additional trackers on shoulders
+ (Pro) Don't need chest. Nice for vive tracker users
+ (Note) Seems to benefit base station tracking more
- (Con) Using inside out is quite bad, and is exaggerated more when hand movement is involved 
- (Con) Controller offset may differ from person to person as they may hold the controller differently to others

(I would consider this for base station users and an "advanced" option)

From HMD
+ (Pro)  More stable
- (Con/Note) No shoulders (Unless u shoulder mount IMUs)
+ (Pro) Controller hand movement does not influence the elbows or shoulders, no elbow and shoulder rocking just because I decide to rotate my hand/wrist
+ (Pro) Much more of a plug-and-play experience (user-friendly)


Louka3000:
You can actually have shoulders now without controllers, but it's a pain to mount!

Also, since you've used arm tracking a lot, do you think the default values (shoulder distance, shoulders width, etc...) are good? We can change those too while we're at it.

Loler920a:
Yup, Yup, was just editing the comment

Louka3000:
Also, for the record, toggling between the two is already supported in config and protocol. Just need the GUI I guess.

Louka3000 (approved):

Loler920a:
Oh jeez, I can't even remember the default values. I'll test that out
Also IMO, shoulder distance and width is still confusing for me to visualise in my head

Loler920a:
Yep hence the little note at the end for the first part
"(I would consider this for base station users and an "advanced" option)" +_+ even I'm not bothered to try mess with that with my current setup

Eirenliel:
So how many trackers do you need for this to work though, and where they should be?

Louka3000:
Two.
One on each upper arm.

Eirenliel (approved):

PR #229, opened 2022-08-07
Config refactor round 2

loucass003:
This is a complete redo of the config system.

It uses the Jackson library to have the configuration based on class definitions. (similar to Gson).
Jackson default language is Json but it has Yaml Parsers so i use it that way to keep the yml format.


Louka3000 (commented):
I’m smol brain and don’t have my PC rn so my reviewing is quite limited, but seems good looking at the files

src/main/java/dev/slimevr/vr/trackers/TrackerFiltering.java
```diff
@@ -1,12 +0,0 @@
-package dev.slimevr.vr.trackers;
-
-public class TrackerFiltering {
-	// TODO move initializing logic from IMUTracker to here so we don't have to
-	// load settings once per tracker. We can just cache them in this class.
-	// Also will make adding filtering to other types of trackers than
-	// IMUTracker much easier if we need to. -Erimel
```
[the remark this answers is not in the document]
loucass003: @ButterscotchV asked me to remove it

src/main/java/dev/slimevr/config/serializers/BooleanMapDeserializer.java
```diff
@@ -0,0 +1,4 @@
+package dev.slimevr.config.serializers;
+
+public class BooleanMapDeserializer extends MapDeserializer<Boolean> {
```
TheButlah, line 3: Add class-level comment on this and the similar classes that says that it is here just to parameterize the MapDeserializer in the `@JsonDeserialize` annotation

TheButlah (commented):
Very pog, simplifies a lot

src/main/java/dev/slimevr/config/VRConfig.java
```diff
@@ -0,0 +1,94 @@
...
+	public TrackerConfig getTracker(Tracker tracker) {
+		TrackerConfig config = trackers.get(tracker.getName());
+		if (config == null) {
+			config = new TrackerConfig(tracker);
+			trackers.put(tracker.getName(), config);
+		}
+		return config;
+	}
+
+	public void loadTrackerConfig(Tracker tracker) {
```
TheButlah, line 75: "save" and "load" implies file IO to me, but these don't actually do IO - they are just updating the in-memory values. I know its named this way because thats what the functions are called in the `Tracker` interface, but consider renaming.

```suggestion
	public void readFromTracker(Tracker tracker) {
```

src/main/java/dev/slimevr/config/VRConfig.java
```diff
@@ -0,0 +1,94 @@
...
+		TrackerConfig config = getTracker(tracker);
+		tracker.loadConfig(config);
+	}
+
+	public void saveTrackerConfig(Tracker tracker) {
```
TheButlah, line 80:
```suggestion
	public void writeToTracker(Tracker tracker) {
```

TheButlah (changes requested):

src/main/java/dev/slimevr/vr/trackers/TrackerFiltering.java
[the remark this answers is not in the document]
ButterscotchV: Not sure I did? I was just wondering why the consts were there, it was strange having default consts in one config class (`FilteringConfig.java`) and the default value in the fields in another (`AutoboneConfig.java`)

src/main/java/dev/slimevr/config/AutoboneConfig.java
```diff
@@ -0,0 +1,28 @@
+package dev.slimevr.config;
+
+public class AutoboneConfig {
```
ButterscotchV, line 3: Minor gripe, the `B` not being capitalized, this change will require renaming the file too but GitHub doesn't let you do that -w-
```suggestion
public class AutoBoneConfig {
```

ButterscotchV (commented):
Waiting to review again once the changes from other reviews are resolved, otherwise it looks good

src/main/java/dev/slimevr/vr/trackers/TrackerFiltering.java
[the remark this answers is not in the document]
Louka3000: either way, the content of that file before the PR was just temporarily.
Futura can bring the file back with the TODO or I’ll just add it back when I do my filtering PR anyway, doesn’t matter.

ButterscotchV (approved):

Louka3000 (approved):

Eirenliel (changes requested):
Good job. Needs fix for VRWorkout and can be merged.

Eirenliel (approved):

Eirenliel (approved):

PR #232, opened 2022-08-12
Add "fast reset" user action to protobuf bridge.

kitlith:
This gives the feeder app the ability to send the Fast Reset action to the server.

### DIFF
--- a/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
+++ b/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
@@ -177,6 +177,9 @@ public abstract class ProtobufBridge<T extends VRTracker> implements Bridge {
 				// TODO : Check pose field
 				Main.vrServer.resetTrackers();
 				break;
+			case "fast reset":
+				Main.vrServer.resetTrackersYaw();
+				break;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -177,6 +177,9 @@ protected void userActionReceived(UserAction userAction) {
 				// TODO : Check pose field
 				Main.vrServer.resetTrackers();
 				break;
+			case "fast reset":
```
Eirenliel, line 180: The naming doesn't make sense. You call reset "calibrate", and fast reset - "fast reset". Please update naming to match, and I also don't think that using space is a good idea in the name, maybe use underscore `_`?

Eirenliel (changes requested):

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
kitlith: Understandable. I didn't change calibrate to reset for backwards compatibility reasons. Perhaps I can make reset trigger for both "calibrate" and "reset", and mark accepting "calibrate" as deprecated?

I'll change "fast reset" to "fast_reset". That said, whitespace *should* be fine here.

kitlith:
I mucked up the formatting, apparently. I'll fix that later, or someone else can.

src/main/java/dev/slimevr/bridge/ProtobufBridge.java
Eirenliel: I didn't mean it will cause problems, just don't think spaces are good in such places conceptually. Also yes, adding both as an option with one deprecated is fine!

Eirenliel (approved):

PR #236, opened 2022-08-24
Remove Slime Java Commons submodule,

Eirenliel:
Move all used files to the project directly


TheButlah (commented):
Where does org/json and java/com/jme3 come from? Are these modifications of any pre-existing java libs?

If so, it makes sense to fork the original lib, apply our changes, and depend on them via gradle instead of including them directly. Otherwise this is a ton of new code to maintain and it become harder over time because our changes from the original lib are not properly tracked.

Eirenliel:
org/json i believe is json.org lib, but i don't remember why it was included. it doesn't require any maintenance.

com/jme3 is from jMonkeyEngine and it's a fork that does not need to be kept up with the original, it diverged like 6 years ago. also original lib is megabytes of code...

Eirenliel:
besides, this is not new code. it was all in Slime Java Commons, @TheButlah. It's actually less code now...

Louka3000 (changes requested):
https://github.com/SlimeVR/SlimeVR-Server/pull/238

loucass003 (approved):

PR #238, opened 2022-08-30
fix comments and remove deprecated method

Louka3000:
The Quaternion slerp method was throwing a (harmless) error in the terminal. It's completely unused and deprecated so I removed it.
Fixed comments too that showed in red in IDEA.

note: this is for https://github.com/SlimeVR/SlimeVR-Server/tree/remove-slime-commons , NOT main.


Eirenliel:
This is definetely a wrong exception, JME is an engine and not a part of java. So the exception should be different, maybe IllegalArgumentException or smth

Louka3000:
> This is definetely a wrong exception, JME is an engine and not a part of java. So the exception should be different, maybe IllegalArgumentException or smth

Right... some of them throw it in the code, I'm just blind :p Fixed it anyway!

TheDevMinerTV (approved):

PR #244, opened 2022-09-21
Refactor code to Java 17

### LINKED ISSUE #243: Change to Java 17
I was thinking of making a PR to add support for Linux, but I need to use Unix sockets. For this I would need a library with JNI or use the classes added in Java 16.

To-do list (should be done in separate PRs)
- [ ] Upgrade Java Version
- [ ] Changes switches to pattern matching
- [ ] Fix closures not using function references and other related stuff
- [ ] Implement Java std lib functions where it's possible
- [ ] Simplify code branching to a result with boolean operations

ImUrX:
fixes #243


build.gradle
```diff
@@ -12,8 +12,8 @@ plugins {
 	id "com.diffplug.spotless" version "6.5.1"
 }
 
-sourceCompatibility = 1.11
-targetCompatibility = 1.11
+sourceCompatibility = 1.17
```
deiteris, line 15: I'd suggest using the `JavaVersion.VERSION_17` constant for that.
```suggestion
sourceCompatibility = JavaVersion.VERSION_17
```

build.gradle
```diff
@@ -12,8 +12,8 @@ plugins {
...
+targetCompatibility = 1.17
```
deiteris, line 16: Same for this one
```suggestion
targetCompatibility = JavaVersion.VERSION_17
```

build.gradle
```diff
@@ -55,16 +55,16 @@ dependencies {
 	implementation 'com.github.jonpeterson:jackson-module-model-versioning:1.2.2'
 	implementation 'org.apache.commons:commons-math3:3.6.1'
 	implementation 'org.apache.commons:commons-lang3:3.12.0'
-	implementation 'net.java.dev.jna:jna:5.10.0'
-	implementation 'net.java.dev.jna:jna-platform:5.10.0'
+	implementation 'net.java.dev.jna:jna:5.12.1'
```
deiteris, line 58: If we want to keep the deps up to date, maybe it would be better if we use are version range prefix? Targeting latest minor version should be fine like that `5.+`. Same for everything else.

ImUrX:
There might be some code that can use the new stuff, should I check on that and include it on this PR or do y'all prefer to do it on a separate PR?

src/main/java/com/jme3/math/FastMath.java
```diff
@@ -695,13 +695,7 @@ public static float tan(float fValue) {
...
 	 */
 	public static int sign(int iValue) {
-		if (iValue > 0) {
-			return 1;
-		}
-		if (iValue < 0) {
-			return -1;
-		}
-		return 0;
+		return Integer.compare(iValue, 0);
```
TheButlah, line 698: There is no guarantee that this function returns `-1`, `0`, and `1`. Implementations of the compare merely require values less than 0, equal to 0, or greater than 0.

From the docs:
> Returns:
>     the value 0 if x == y; a value less than 0 if x < y; and a value greater than 0 if x > y

We should therefore keep the original implementation here.

ImUrX: I changed it to Integer.signum() then

Eirenliel:
@ImUrX please don't change what's not broken... i understand new syntax might be apealing to you, but 1) it doesn't warrant changing half of the code for this, 2) i don't think the new switches are more readable.

ImUrX:
@Eirenliel I don't think most of the changes are about the switches tho, but I do prefer pattern matching instead of the switch-break style. I think the ones that benefit the most are the small ones that get to be one-liners.
The other part is that switches are not normal. they dont use braces. Pattern matching does and I think it's good for that to happen.

Add reviews on the ones you don't want to use pattern-matching on and we can discuss it on a per-case basis.

ImUrX:
btw, rereading your comment. Most things I did are equivalent but faster or more readable or both, I didn't make new methods but just used the ones that Java provides, it's not risky to use them and it's easily checkable if I did because they are already so well-documented

src/main/java/com/jme3/math/FastMath.java
deiteris: Why do you think there's no guarantee? From OpenJDK implementation, it returns all these possible options and does the same comparison when compared to zero.
https://github.com/openjdk/jdk/blob/f8878cb0cc436993ef1222bc13b00b923d91aad1/src/java.base/share/classes/java/lang/Integer.java#L1486

TheButlah:
@Eirenliel BTW, we have had bugs in the past due to missing `break` statements in switch cases that got through code review. I think that if Java provides a new way of doing it, its not a bad idea to adopt it. Just my 2 cents. And so far going through the code, I haven't found any java-provided functions that could be less optimized than the way we were doing it already. In fact its likely to be faster

ImUrX:
[I did ask](https://github.com/SlimeVR/SlimeVR-Server/pull/244#issuecomment-1254268436) but didnt get a response, so thought it wasn't important if I did so :c
I think I may be able to do some git magic and PR only the version change tho if that's actually needed

TheButlah (changes requested):
I would refrain from modifying the com/jme3 files because those originate from the jmonkey engine and its best not to modify them, because that code is well tested and not written by us.

ImUrX:
as a side comment I would make a suggestion to change to an actual library instead of including jmonkey's codebase. Libs like ``ojAlgo`` or ``UJMP`` or even Apache's Common Math. This will bring a performance improvement for sure and make the codebase a lot cleaner and make things simpler license-wise

ImUrX:
Reading #236 seems there is a little history about why this is all in here lol

Eirenliel:
> @Eirenliel BTW, we have had bugs in the past due to missing `break` statements in switch cases that got through code review. I think that if Java provides a new way of doing it, its not a bad idea to adopt it. Just my 2 cents. And so far going through the code, I haven't found any java-provided functions that could be less optimized than the way we were doing it already. In fact its likely to be faster
> 
> That being said, @ImUrX its generally a good idea to ask about things like this first, and make smaller, incremental PRs because it makes it easier for reviewers to go "hey I like this change lets accept all of this type of refactor" or "no I don't like this particular type of refactor"

okay, then you can proceed i guess

TheButlah:
Honestly there are so many files changed that reviewing this will literally take 2 hours...
Screen Shot 2022-09-23 at 10 49 28 PM

ImUrX:
I had IDEA detect them and I reviewed everything by hand, I think I only automated deleting unused imports

TheButlah:
> I had IDEA detect them

Excellent, then I won't feel too bad if I ask you to throw away that work and break it into separate PRs. Could you:
* Make this PR just for upgrading the java version
* Make a separate PR *just* for changing the switch statements to the new way of doing it
* Make a separate PR for the closure-related changes
* Make a separate PR for the use of java std lib functions instead of our manual way of doing them
* Make a separate PR for the "return result of bool expression instead of branching to do that" changes

Basically, my thought is that by putting all of one type of refactoring in its own PR it will make it easier to review because:
1. Everything in the PR is the same type of change, so it will be immediately clear if something slipped past that was not supposed to be there
2. The number of LOC changed is much smaller in each PR, so we can do one at a time without taking 3 hours

Hopefully you will understand why we are asking this hassle of you, I apologize if it is a bit annoying, but its just too hard to review this PR in its current state.

ImUrX:
i understand but IDEA didnt do most of the work, i wasted like 2 hours doing this 😭

src/main/java/dev/slimevr/protocol/DataFeedBuilder.java
```diff
@@ -192,7 +190,7 @@ public static int createDeviceData(
 		if (!mask.getDeviceData())
 			return 0;
 
-		if (device.getTrackers().size() <= 0)
+		if (device.getTrackers().size() == 0)
```
ButterscotchV, line 193: Why? I don't see any reason to do `==`, `<=` seems fine to me, doesn't have any performace difference, and provides safety with implementations or extremely weird bugs

src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
```diff
@@ -1684,34 +1635,24 @@ public void resetSkeletonConfig(SkeletonConfigOffsets config) {
 				{
 					skeletonConfig.setOffset(SkeletonConfigOffsets.TORSO, null);
 				}
-				break;
-			case CHEST: // Chest is 57% of the upper body by default (shoulders
+			}
+			case CHEST -> // Chest is 57% of the upper body by default
+							// (shoulders
 				// to chest)
```
ButterscotchV, line 1641: This comment is split weirdly, just some visual annoyance
```suggestion
			case CHEST -> // Chest is 57% of the upper body by default
							// (shoulders to chest)
```

src/main/java/dev/slimevr/websocketapi/WebSocketVRBridge.java
```diff
@@ -93,17 +93,20 @@ public void onMessage(WebSocket conn, String message) {
...
+					}
+					case "action" -> {
 						parseAction(json, conn);
 						return;
-					case "config": // TODO Ignore it for now, it should only
-									// register HMD in our test case with id
+					}
+					case "config" -> { // TODO Ignore it for now, it should only
+						// register HMD in our test case with id
 						// 0
```
ButterscotchV, line 106: Weirdly split comment
```suggestion
					case "config" -> { // TODO Ignore it for now, it should only
						// register HMD in our test case with id 0
```

ButterscotchV (changes requested):
- In `Main.java`, the minimum Java version will need to be bumped from `JavaVersion.JAVA_11`
- I don't know if I agree with all the removal of the access modifiers, but it's fine other than just my opinion
- The org.json stuff looks fine, but it's hard to say without unit tests, we really need to look into using Jackson like we already have as described in #246

TheButlah:
I would like to second the opinion of @ButterscotchV , if changing the access modifiers makes it *more* visible, that should be avoided

ImUrX:
In which cases is it okay to remove visibility, like public interfaces have public methods by default they wouldn't be protected, enums constructors can only be private constructors, etc.

src/main/java/dev/slimevr/bridge/ProtobufMessages.java
```diff
@@ -24,67 +24,82 @@ public final class ProtobufMessages {
```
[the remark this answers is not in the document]
ImUrX: I had the message hidden by IDEA lol

src/main/java/dev/slimevr/websocketapi/WebSocketVRBridge.java
ImUrX: this one is weird, [look at the original](https://github.com/SlimeVR/SlimeVR-Server/blob/main/src/main/java/dev/slimevr/websocketapi/WebSocketVRBridge.java#L102)

ButterscotchV (approved):
Good to me, what I meant with the access modifiers is they're all the same, but I just personally like to explicitly define it, this is perfectly fine though and it's just a visual difference

ImUrX:
@Eirenliel what's left for this to merge?

Eirenliel:
I'm doing a review, but still, the installer needs to be updated to Java 17 before this can be released.

ImUrX:
I can do a PR for it if it's needed

Eirenliel:
whoever does it, it's needed yes

ImUrX:
I meant that if you want me to make the PR. Sure, I will do it lol

TheButlah:
Blocked by https://github.com/SlimeVR/SlimeVR-Installer/pull/28

ButterscotchV (approved):
Changes since last review are tiiiny, it still looks good to me

Eirenliel (approved):

loucass003 (approved):

loucass003:
We are all good i think @Eirenliel

Eirenliel (approved):

PR #267, opened 2022-11-12
Add Gui to the server

loucass003:
Combine both server and gui in the same repo.

I wanted to make gradle launch / compile both apps .... but could not find how to start two tasks in parallell
so i have not done it. If someone knows how it can be added later.

### DIFF
--- a/resources/ThirdPartyNotices.txt
+++ b/resources/ThirdPartyNotices.txt
@@ -1,0 +1,1446 @@
+NOTICES
+
+Listed below are the licenses of the projects the SlimeVR Server has code from.
+
+
+---------------------------------------------------------------
+Apache Commons CLI
+Copyright 2002-2022 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (https://www.apache.org/).
+---------------------------------------------------------------
+
+---------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Gui/.github/workflows/build-gui.yml
```diff
@@ -0,0 +1,40 @@
+name: Build GUI
```
TheDevMinerTV, line 1: shouldn't be here

Gui/.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
TheDevMinerTV, line 1: shouldn't be here

Gui/.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
```
TheDevMinerTV, line 1: won't get recognized by VSCode here

TheDevMinerTV (changes requested):

Eirenliel:
so how will the history look like :nya_a:

also please use lower case letters for folder names...

loucass003:
> so how will the history look like :nya_a:

History is gone

PR #271, opened 2022-11-14
Create a mapping for remote sensor id to internal.

carl-anders:
This allows UDP setup packets to be sent out of order.
Sensor ID's are now decoupled from the internal id representation in slimevr.

The reason for adding this is that on my wifi udp packets often seem to be rearranged. I have to restart the slimevr server one or two times for it to recognize all my trackers with extensions. So if sensor 1 is sent first then it will be assigned to sensor 0 because of the way the code is currently written. But with a mapping layer it will remember the real position in the trackers list.

This will also allow programs such as slimevr wrangler to send an unique id with every joycon, making it so users don't have to constantly reassign their tracker positions.

### DIFF
--- a/src/main/java/dev/slimevr/vr/trackers/udp/UDPDevice.java
+++ b/src/main/java/dev/slimevr/vr/trackers/udp/UDPDevice.java
@@ -8,6 +8,7 @@ import io.eiren.util.collections.FastList;
 
 import java.net.InetAddress;
 import java.net.SocketAddress;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 public class UDPDevice extends Device {
@@ -27,6 +28,7 @@ public class UDPDevice extends Device {
 	public int firmwareBuild = 0;
 	public boolean timedOut = false;
 	private final FastList<Tracker> trackers = new FastList<>();
+	private final ConcurrentHashMap<Integer, Integer> remote_id_map = new ConcurrentHashMap<>();
 
 	public UDPDevice(SocketAddress address, InetAddress ipAddress) {
 		this.address = address;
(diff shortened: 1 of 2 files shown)


src/main/java/dev/slimevr/vr/trackers/udp/UDPDevice.java
```diff
@@ -27,6 +28,7 @@ public class UDPDevice extends Device {
 	public int firmwareBuild = 0;
 	public boolean timedOut = false;
 	private final FastList<Tracker> trackers = new FastList<>();
+	private final ConcurrentHashMap<Integer, Integer> remote_id_map = new ConcurrentHashMap<>();
```
Eirenliel, line 31: please change it to `HashMap<Integer, Tracker> trackerByRemoteId = new HashMap<>();` there is no need to use additional re-mapping of IDs you can fetch trackers directly, and there is no need to use ConcurrentHashMap as all UDP processing is done in one thread.

Eirenliel (changes requested):

src/main/java/dev/slimevr/vr/trackers/udp/UDPDevice.java
carl-anders: > there is no need to use additional re-mapping of IDs you can fetch trackers directly

What do you mean with this? Tracker sensor ID's must be created in order, starting with 0, going 1, 2, 3 etc.
The row above you see is a FastList, which is not able to have arbitrary ID's. This is a problem when UDP packets are sent in the wrong order. My trackers sometimes send a set up packet for sensor 1, then sensor 0. This means that the sensor 1 info is set to the sensor 0 info, and the sensor 0 info is ignored, because it "already exists".

Eirenliel: You don't need to map remote sensor id to local sensor id, you can map remote sensor id directly to the sensor object. Also the name of the variable is wrong, java uses camelCase.

carl-anders: Other code needs the trackers object to be a list that is contiguous. A bigger rewrite of the TrackersUDPServer.java file might be nice too, but right now this is the simplest way to add a mapping layer.

Eirenliel: You can literally have both, you know.

Eirenliel: I understand the problem, I just think that it's not a good solution to it. It's the one with the least code change, yes, but it's not the best way. Either you need to duplicate the list and add a map nearby, or replace the list with the map and change the code that uses the list, imo.

PR #274, opened 2022-11-18
OSC Support

Louka3000:
Support for sending tracker data to applications (currently VRChat) via OSC.

### DIFF
--- a/src/main/java/dev/slimevr/osc/VRCOSCHandler.java
+++ b/src/main/java/dev/slimevr/osc/VRCOSCHandler.java
@@ -7,11 +7,9 @@ import com.illposed.osc.transport.OSCPortOut;
 import com.jme3.math.FastMath;
 import com.jme3.math.Quaternion;
 import com.jme3.math.Vector3f;
-import dev.slimevr.bridge.PipeState;
 import dev.slimevr.config.OSCConfig;
 import dev.slimevr.platform.windows.WindowsNamedPipeBridge;
 import dev.slimevr.vr.processor.HumanPoseProcessor;
-import dev.slimevr.vr.processor.skeleton.Skeleton;
 import dev.slimevr.vr.trackers.*;
 import io.eiren.util.collections.FastList;
 import io.eiren.util.logging.LogManager;
@@ -131,23 +129,18 @@ public class VRCOSCHandler {
 
 	void handleReceivedMessage(OSCMessageEvent event) {
 		timeAtLastOSCMessageReceived = System.currentTimeMillis();
-		if (
-			steamvrBridge.getPipeState() != null
-				&& steamvrBridge.getPipeState() != PipeState.OPEN
-		) {
+		if (steamvrBridge != null && steamvrBridge.isConnected()) {
 			if (hmd.getStatus() != TrackerStatus.OK) {
 				hmd.setStatus(TrackerStatus.OK);
 			}
 			hmd.position
 				.set(
-					new Vector3f(
-						0f,
-						(float) event
-							.getMessage()
-							.getArguments()
-							.get(0) * humanPoseProcessor.getUserHeightFromConfig(),
-						0f
-					)
+					0f,
+					(float) event
+						.getMessage()
+						.getArguments()
+						.get(0) * humanPoseProcessor.getUserHeightFromConfig(),
+					0f
 				);
 			hmd.rotation.set(Quaternion.IDENTITY);
 			hmd.dataTick();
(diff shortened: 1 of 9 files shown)


src/main/java/dev/slimevr/osc/VRCOSCHandler.java
```diff
@@ -0,0 +1,252 @@
...
+			if (hmd.getStatus() != TrackerStatus.OK) {
+				hmd.setStatus(TrackerStatus.OK);
+			}
+			hmd.position
+				.set(
+					0f,
+					(float) event
+						.getMessage()
+						.getArguments()
+						.get(0) * humanPoseProcessor.getUserHeightFromConfig(),
```
Louka3000, line 142: might want a more optimized way to do this but I’m too tired to figure it out

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
```diff
@@ -0,0 +1,252 @@
...
+
+		// Stops listening and closes OSC port
+		if (oscReceiver != null && oscReceiver.isListening()) {
+			oscReceiver.stopListening();
+		}
+		if (oscSender != null && oscSender.isConnected()) {
+			try {
+				oscSender.close();
+			} catch (IOException e) {
+				LogManager.warning("[VRCOSCHandler] Error closing the OSC sender: " + e);
```
ButterscotchV, line 77: This seems like a pretty bad error, maybe log it as severe?
```suggestion
				LogManager.severe("[VRCOSCHandler] Error closing the OSC sender: " + e);
```

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
```diff
@@ -0,0 +1,252 @@
...
+			// Instantiates the OSC receiver
+			try {
+				int port = config.getPortIn();
+				oscReceiver = new OSCPortIn(
+					port
+				);
+				LogManager.info("[VRCOSCHandler] Listening to port " + port);
+			} catch (IOException e) {
+				LogManager
+					.warning("[VRCOSCHandler] Error listening to the port " + config.getPortIn());
```
ButterscotchV, line 91: This should *definitely* be severe
```suggestion
					.severe("[VRCOSCHandler] Error listening to the port " + config.getPortIn());
```

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
```diff
@@ -0,0 +1,252 @@
...
+				LogManager
+					.info(
+						"[VRCOSCHandler] Sending to port "
+							+ port
+							+ " at address "
+							+ address.toString()
+					);
+			} catch (IOException e) {
+				LogManager
+					.warning(
```
ButterscotchV, line 120: This should be severe
```suggestion
					.severe(
```

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
ButterscotchV: this looks really weird, what on earth is going on here 🤔

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
```diff
@@ -0,0 +1,252 @@
...
+					&& hmd.getStatus() == TrackerStatus.OK
+			) {
+				hmd.setStatus(TrackerStatus.DISCONNECTED);
+			}
+		}
+
+		// Send OSC data
+		if (oscSender != null && oscSender.isConnected()) {
+			for (int i = 0; i < shareableTrackers.size(); i++) {
+				if (trackersEnabled[i]) {
```
ButterscotchV, line 169: This line is scary, but looking into the code for how `shareableTrackers` is acquired, there's not much that can be done about it... A check for the bounds of `trackersEnabled` is possible but really not necessary, I *really* don't like how `HumanPoseProcessor` does this

ButterscotchV (changes requested):

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
Louka3000: Getting VRC's Upright value (HMD'y position scaled to 0-1) and scaling it back to 0-userheight with proportions.

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
ButterscotchV: I'm gonna resolve this due to its irrelevance to this PR, hopefully it can just be tackled in some refactor later down the line

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
ButterscotchV: Maybe SkeletonConfig could cache this... Otherwise I don't think there's a much better solution with the current setup of things

ButterscotchV (approved):

src/main/java/dev/slimevr/osc/VRCOSCHandler.java
Louka3000: caching it in SkeletonConfig and calling it from HumanPoseProcessor

Eirenliel (approved):

PR #290, opened 2022-11-23
Feature/gui osc implem

loucass003:


Eirenliel (approved):

ButterscotchV (approved):
I won't be picky, so I'll approve this, we need to reconsider how we handle setting config stuff and especially getting default config values in the future

Louka3000 (approved):

PR #292, opened 2022-11-24
update java 17 download link in run.bat

Louka3000:
Related: do we want to keep the run.bat or should we remove it?


ButterscotchV (approved):
I think we should just remove it, this change is fine though

PR #295, opened 2022-11-24
Fix appDir() causing GUI config to not be loaded

ButterscotchV:
Gives output of `invalid value: 21, expected one of: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20` at startup, the result is already unused, so this just removes it to resolve the issue.

### DIFF
--- a/gui/src/hooks/config.ts
+++ b/gui/src/hooks/config.ts
@@ -4,7 +4,6 @@ import {
   writeFile,
   createDir,
 } from '@tauri-apps/api/fs';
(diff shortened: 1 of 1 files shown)


TheDevMinerTV (approved):
LGTM

loucass003:
Bruh moment, tori must have broken it because it always has been there ...

loucass003 (approved):

PR #306, opened 2022-11-25
Clarify Licensing

TheButlah:
Now that there are more than just MIT license in this repo, I wanted to make some clarifications.

One additional thing other than just clarifying is that this makes all contributions *from here on out* dually licensed. We canot dual license the server just yet because the server predated this requirement, but having the explicit statement that new contributions will be dually licensed makes it easier to be safe and not get a situation where someone thought their code was only MIT but its actually dual.

It also covers our asses if we want to dual license the server in the future, there are really only a few of us that have made nontrivial changes that would need to agree - any new contributors would have already agreed to dual license.


TheButlah:
Let me know what your thoughts are :)

Eirenliel (approved):

Louka3000:
From a website stating the differences between MIT and Apache:
> Another key difference is that the Apache license requires developers to disclose any major changes they make to the original source code. The modified source code does not need to be revealed, but a notice of the modification is required. However, any unmodified code must retain the Apache license.

With SlimeVR now being dual-licensed, does that mean people using our code must disclose big code changes?

Eirenliel:
> With SlimeVR now being dual-licensed, does that mean people using our code must disclose big code changes?

They can use it under either of the licenses, so they can simply use MIT license. They don't have to follow both, they have to contribute under both.

Louka3000:
Alright I see, thanks for clarifying

PR #310, opened 2022-11-26
Make a dialog appear when rust app panics

ImUrX:
This implements a cross platform crate that shows the actual panic. Pretty helpful when not launching from terminal.
Also implements ~~random~~helpful window titles when one appears.
crash window

### DIFF
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -1,15 +1,28 @@
 #![cfg_attr(
-	all(not(debug_assertions), target_os = "windows"),
+	all(not(debug_assertions), windows),
 	windows_subsystem = "windows"
 )]
 
 use clap::Parser;
 use clap_verbosity_flag::{InfoLevel, Verbosity};
+use native_dialog::MessageDialog;
+use native_dialog::MessageType;
+use rand::seq::SliceRandom;
+use rand::thread_rng;
 use std::env;
+use std::panic;
 use std::path::PathBuf;
 use tauri::api::clap;
 use tauri::api::process::Command;
 use tauri::Manager;
+
+static POSSIBLE_TITLES: &[&str] = &[
+	"Panicking situation",
+	"looking for spatula",
+	"never gonna give you up",
+	"never gonna let you down",
+	"uwu sowwy",
+];
 
 #[derive(Parser)]
 #[clap(version, about)]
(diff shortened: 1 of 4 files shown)


gui/src-tauri/src/main.rs
```diff
@@ -1,16 +1,29 @@
 #![cfg_attr(
-	all(not(debug_assertions), target_os = "windows"),
+	all(not(debug_assertions), windows),
 	windows_subsystem = "windows"
 )]
 
 use clap::Parser;
 use clap_verbosity_flag::{InfoLevel, Verbosity};
+use native_dialog::MessageDialog;
+use native_dialog::MessageType;
```
TheButlah, line 9:
```suggestion
use native_dialog::{MessageDialog, MessageType};
```

gui/src-tauri/src/main.rs
```diff
@@ -1,16 +1,29 @@
...
+use rand::seq::SliceRandom;
+use rand::thread_rng;
 use std::env;
+use std::panic;
 use std::path::PathBuf;
 use tauri::api::clap;
 use tauri::api::process::Command;
 use tauri::Manager;
 
+static POSSIBLE_TITLES: &[&str] = &[
```
TheButlah, line 19:
```suggestion
const POSSIBLE_TITLES: &[&str] = &[
```

TheButlah (changes requested):
.

gui/src-tauri/src/main.rs
ImUrX: const for arrays isnt a smart idea, if we ever use that variable more than one time it will be making 2 arrays.
More info in here: https://doc.rust-lang.org/std/keyword.const.html#compile-time-constants

TheButlah: TIL

TheButlah (approved):

loucass003:
Can we add some more dialogs ?

we need one when java is not detected.
Also is there no way to detect webview directly and have a dedicated message for it ?

loucass003 (changes requested):
.

TheButlah:
Just make new PRs for those, let him merge this

Eirenliel (approved):

ImUrX:
> Can we add some more dialogs ?
> 
> we need one when java is not detected (make sure it does not trigger in dev mode tho rofl ) Also is there no way to detect webview directly and have a dedicated message for it ?

The Java one is possible and shouldn't be that hard.
I tried making a webview one but it won't even be able to launch because it fails to link with the unexisting webkitgtk shared object. I wouldn't expect people to install it outside using an AppImage or from their package manager, so this shouldn't matter. And the error the linker returns says it didnt find webkit2gtk-4, so I think that's good enough. (only on console tho :/)
The Webview error on Windows is weird, look I got an example here
image
It's a ``CreateWebview(WebView2Error(WindowsError()))``, when there is already a ``WebviewNotFound`` error. We can detect that and suggest installing WebView2 from Microsoft, don't know if it's the best idea tho.
Here you can see the errors btw https://docs.rs/tauri/latest/tauri/enum.Error.html

PR #316, opened 2022-11-28
Only write changed toggles and values to config

Louka3000:
Only writing changed toggles and values allows us to change the defaults and have the change reflect on the users that didn't change the default values.
Still writing the default skeleton offsets (bone lengths) to config, since people liking the default values and not wanting to change to a new default is much more likely.

I also fixed skating correction and floor clip being switched around in RPCSettingsBuilder and removed some unused code in SkeletonConfig.
Also make reset all only reset offsets
Also add autobone performance improvements by not doing stuff we already did (by Butterscotch)
Also make reset all use HMD to reset all (it didn't before because *someone* forgot to)


Eirenliel (approved):

server/src/main/java/dev/slimevr/autobone/AutoBone.java
```diff
@@ -529,7 +529,7 @@ public AutoBoneResults processFrames(
 					int frameCursor2 = frameCursor + cursorOffset;
 
 					applyConfig(skeleton1.skeletonConfig);
-					skeleton2.skeletonConfig.setConfigs(skeleton1.skeletonConfig);
```
ButterscotchV, line 532: Not sure why I even used `setConfigs`, make sure this still works and lets you apply them properly (Or I can do that <3)

ButterscotchV (changes requested):
Should reset the `changedToggles` and `changedValues` arrays in `SkeletonConfig.resetConfigs()`, make sure to test AutoBone (this might actually even make it faster, as SkeletonConfig is the slowest part :P).

server/src/main/java/dev/slimevr/vr/processor/skeleton/SkeletonConfig.java
```diff
@@ -21,6 +21,9 @@ public class SkeletonConfig {
 		SkeletonConfigValues.class
 	);
 
+	protected boolean[] changedToggles = new boolean[SkeletonConfigToggles.values().length];
+	protected boolean[] changedValues = new boolean[SkeletonConfigValues.values().length];
```
ButterscotchV, line 25: Oh also these should use `values` and not `values()`
```suggestion
	protected boolean[] changedToggles = new boolean[SkeletonConfigToggles.values.length];
	protected boolean[] changedValues = new boolean[SkeletonConfigValues.values.length];
```

ButterscotchV (changes requested):

server/src/main/java/dev/slimevr/autobone/AutoBone.java
Louka3000: I’m at school rn, if you’re not busy you can do it :3 
else I’ll do it when I get home

Louka3000: Tested it, Autobone works.

ButterscotchV (approved):

Louka3000:
Converted to draft for now so it doesn’t get merged yet.
ResetOffset needs to call the skeleton in order to use the HMD for reset!!!
Also maybe add a “Reset settings” that clears values and toggles since we’ll need that anyway

ButterscotchV (changes requested):
I really have to resist complaining about adding Skeleton to SkeletonConfig ;~;
It's good 'cause it's honestly the best way to go about this right now, but I'm dying inside a little

server/src/main/java/dev/slimevr/vr/processor/skeleton/SkeletonConfig.java
```diff
@@ -345,27 +360,110 @@ public void setConfigs(
...
+
+		save();
+	}
+
+	public void resetToggles() {
+		configToggles.clear();
+
 		if (autoUpdateOffsets) {
 			computeAllNodeOffsets();
 		}
```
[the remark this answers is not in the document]
Louka3000: oops indeed

server/src/main/java/dev/slimevr/vr/processor/skeleton/SkeletonConfig.java
```diff
@@ -345,27 +360,110 @@ public void setConfigs(
...
+		// Copy skeletonConfig's nodeOffsets as the configs are all the same
+		skeletonConfig.nodeOffsets.forEach((key, value) -> {
+			setNodeOffset(key, value.x, value.y, value.z);
+		});
+	}
+
+	public void resetOffsets() {
+		for (SkeletonConfigOffsets config : SkeletonConfigOffsets.values) {
+			skeleton.resetSkeletonConfig(config);
+		}
```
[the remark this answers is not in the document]
Louka3000: Good catch thanks :3

PR #319, opened 2022-11-30
Add more readable errors for the launcher

### LINKED ISSUE #311: Proper error reporting in GUI
GUI should report when it can't start server, continuation of #310

ImUrX:
Lots of changes like:
- Use tauri's dialog lib
- Added a mobile and desktop ``#[cfg()]`` alias, will help if something ever happens in the future (?
- Check for missing WebView2 on Windows, kind of impossible to do on Linux because it will just fail in the linking stage
- Check for Java binaries if the bundled JRE is not found. It checks for the version of each through a small ``.class`` program

Requires testing in Windows, haven't done it myself yet. Mainly the registry check
Fixes #311


gui/src-tauri/src/main.rs
```diff
@@ -154,6 +214,58 @@ fn main() {
...
+	}
+
+	// Otherwise check if anything else is a supported version
+	let mut childs = vec![];
+	for java in which_all("java").unwrap() {
+		let res = spawn_java(java.as_os_str(), java_version.as_os_str());
+
+		match res {
+			Ok(child) => childs.push((java.into_os_string(), child)),
+			Err(e) => eprintln!("Error on trying to spawn a Java executable: {}", e),
```
TheButlah, line 257: I would use `log` here, as we often don't have a mechanism for seeing stderr of tauri. Log will allow us to log to a file in the future if we so choose.

If you want, you can use `pretty_env_logger` as the global logger, for now. Set the default log level to `info`.
You can do that in a followup PR though, I prefer small incremental merges and this PR is already large-ish.
Created #321 to track this.

TheButlah (changes requested):
Several style nits. The actual feature/enhancements I suggested I have opened separate issues to track, those can be addressed by @ImUrX or @TheButlah in a separate PR.

TheButlah:
Closes #311

gui/src-tauri/src/main.rs
```diff
@@ -103,6 +128,29 @@ fn main() {
...
+			.map(|r| r.get_value("pv").ok()).ok().flatten();
+		let mut exists = false;
+		if let Some(version) = machine {
+			exists = version.split('.').any(|x| x != "0");
+		}
+		// Then in the current user
+		if !exists {
+			let user: Option<String> = RegKey::predef(HKEY_CURRENT_USER)
+			.open_subkey(r"Software\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}")
+			.map(|r| r.get_value("pv").ok()).ok().flatten();
```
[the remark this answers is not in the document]
ImUrX: tauri provides it's own installer that should take care of this :p

gui/src-tauri/src/main.rs
```diff
@@ -103,6 +128,29 @@ fn main() {
...
+			if let Some(version) = user {
+				exists = version.split('.').any(|x| x != "0");
+			}
+		}
+		if !exists {
+			show_error("Couldn't find WebView 2 installed. You can install it from Microsoft's site or try reinstalling it from the SlimeVR installer");
```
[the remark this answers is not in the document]
ImUrX: I don't think something like ``thiserror`` should be needed for a launcher, I understand using it for libs and other stuff. But I will have to look at the PR tbh

gui/src-tauri/src/main.rs
```diff
@@ -154,6 +214,58 @@ fn main() {
 			Ok(())
 		})
 		//
-		.run(tauri::generate_context!())
-		.expect("error while running tauri application");
+		.run(tauri::generate_context!());
+	match res {
```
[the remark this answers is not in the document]
ImUrX: I added it for future convenience, I was going to add a check for a runtime error, but I would need ``tauri-runtime`` as a dependency

gui/src-tauri/src/main.rs
```diff
@@ -154,6 +214,58 @@ fn main() {
...
+	let mut file = Builder::new()
+		.suffix(".class")
+		.tempfile()
+		.expect("Couldn't generate .class file");
+	file.write_all(include_bytes!("JavaVersion.class"))
+		.expect("Couldn't write to .class file");
+	let java_version = file.into_temp_path();
+
+	// Check if main Java is a supported version
+	let main_java = if let Ok(java_home) = std::env::var("JAVA_HOME") {
```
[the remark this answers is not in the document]
ImUrX: I can't seem to find a way to implement ``let-else`` statements, I see that they are mostly used for early returns.

In this case how can I convert the ``JAVA_HOME`` var to an ``OsString``?
I could use ``.map`` but we are mixing functional with expressions again.

Then the other ``if let`` is okay as it is because it doesn't use an ``else``

gui/src-tauri/src/main.rs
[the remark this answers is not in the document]
TheButlah: No point preemptively doing something like that, just change it when you actually need it. For now since it makes it harder to read, let's revert that to how it was

gui/src-tauri/src/main.rs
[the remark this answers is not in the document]
TheButlah: Looks like I misunderstood how let else works. The else must return `!`. So you're right, a let else is not possible here.

gui/src-tauri/src/main.rs
```diff
@@ -111,20 +140,30 @@ fn main() {
...
-				.current_dir(p)
-				.args(["-Xmx512M", "-jar", "slimevr.jar", "--no-gui"])
-				.spawn()
-				.expect("Unable to start the server jar");
+		// Check if any Java already installed is compatible
+		let java_paths = valid_java_paths();
+		let jre = p.join("jre/bin/java");
+		let java_bin = jre
+			.exists()
+			.then(|| jre.to_string_lossy())
```
TheButlah, line 148: I'm on my phone so I wasn't able to test this, but isn't this only checking the first element of the Java path vec? Did you instead mean to iterate, map, and take the first one that was valid?

TheButlah (changes requested):

gui/src-tauri/src/main.rs
ImUrX: Yeah, this grabs the first found compatible version. If length is higher than 1, the rest will be ignored.

TheButlah: Oh I understand. So once youve gotten the valid paths, you only care about getting the first one, because the rest are redundant.

Resolving.

TheButlah:
Just needs the change for the unecessary match statement, then I will approve.

TheButlah (approved):

Eirenliel (approved):

gui/src-tauri/src/main.rs
```diff
@@ -103,6 +127,11 @@ fn main() {
 		// We don't do anything with the job anymore, but we shouldn't drop it because that would
 		// terminate our process tree. So we intentionally leak it instead.
 		std::mem::forget(job);
+
+		if !webview2_exists() {
+			show_error("Couldn't find WebView 2 installed. You can install it from Microsoft's site or try reinstalling it from the SlimeVR installer");
```
ButterscotchV, line 132: Maybe link to Microsoft's site for WebView2 and link to the installer, this doesn't guide users very well (Maybe just say to use the installer if we just want it simple for users? They might get confused with 2 "random" links)

gui/src-tauri/src/main.rs
```diff
@@ -111,13 +140,23 @@ fn main() {
...
-				.args(["-Xmx512M", "-jar", "slimevr.jar", "--no-gui"])
-				.spawn()
-				.expect("Unable to start the server jar");
+		// Check if any Java already installed is compatible
+		let java_paths = valid_java_paths();
+		let jre = p.join("jre/bin/java");
+		let java_bin = jre
+			.exists()
+			.then(|| jre.to_string_lossy())
+			.or_else(|| java_paths.first().map(|x| x.0.to_string_lossy()));
```
ButterscotchV, line 149: `let java_paths = valid_java_paths();` should be retrieved after checking the bundled Java version, as launching a bunch of child processes only to not use the result is kinda wasteful.

ButterscotchV (changes requested):
Are we supposed to include the JavaVersion class binary?

gui/src-tauri/src/main.rs
ImUrX: I cant have stuff like hyperlinks with names, I would be adding a whole URL

gui/src-tauri/src/main.rs
ImUrX: ehhhhh, I know. But java_bin is a reference ``Cow<str>``.
What do you suggest

TheButlah (approved):

ButterscotchV (approved):

PR #330, opened 2022-12-02
Fix OSC rotations, using OSC with SteamVR and reduce error logging spam

Erimelowo:
- I forgor to send HMD rotation, I just hardcoded what it was when no HMD was present. This worked fine when no SteamVR or when using standalone, but it breaks if using the steamvr driver. Fixed it.
- Reduce the logging amount of AsynchronousCloseException.


Louka3000:
Rotation is messed up…

Louka3000:
@ButterscotchV I'll E-Transfer you 20$ if you fix the coordinate system.
I just wasted my whole evening on this and I wanna punch my wall and never touch this again.

Louka3000:
Ready for review/merge
Zrock and I tested that it works in VRC as well.

ButterscotchV (approved):
> Zrock and I tested that it works in VRC as well.

Approved, thanks for confirming it's tested <3

PR #337, opened 2022-12-05
Fix launching Java from the launcher

ImUrX:
- Made Windows not make a cmd appear when locating Java versions
- Fixed current_dir being an empty string which made Windows not like it
- Made JavaVersion a .jar, making it a .class doesnt make things simple
- added some more logging around


gui/src-tauri/Cargo.toml
```diff
@@ -27,22 +27,7 @@ cfg_aliases = "0.1"
...
-    "window-close",
-    "window-maximize",
-    "window-minimize",
-    "window-set-resizable",
-    "window-set-title",
-    "window-start-dragging",
-    "window-unmaximize",
-    "window-unminimize"
-] }
+tauri = { version = "1.2", features = ["cli", "devtools", "dialog", "fs-all", "path-all", "shell-execute", "window-close", "window-maximize", "window-minimize", "window-set-resizable", "window-set-title", "window-start-dragging", "window-unmaximize", "window-unminimize"] }
```
ImUrX, line 30: This is provoked by tauri cli sadly

gui/src-tauri/src/main.rs
```diff
@@ -282,7 +294,19 @@ fn valid_java_paths() -> Vec<(OsString, i32)> {
 				.expect("Failed on executing a Java executable")
 				.code()
 				.map(|code| (p, code))
-				.filter(|(_p, code)| *code >= MINIMUM_JAVA_VERSION)
+				.filter(|(_p, code)| {println!("{}", code); *code >= MINIMUM_JAVA_VERSION})
 		})
 		.collect()
 }
+
+fn executable<W: AsRef<OsStr> + ?Sized>(bin: &W) -> Cow<OsStr> {
```
TheButlah, line 302: There is no need to 
```
\|/          (__)    
     `\------(oo)
       ||    (__)
       ||w--||     \|/
   \|/
```

We aren't writing performance critical code where a `.clone()` is too expensive. Just return an `OsStr`

TheButlah (changes requested):

gui/src-tauri/src/main.rs
ImUrX:
```
 __________________ 
< why the hate tho >
 ------------------ 
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

gui/src-tauri/src/main.rs
```diff
@@ -45,7 +51,8 @@ fn is_valid_path(path: &PathBuf) -> bool {
 
 fn get_launch_path(cli: Cli) -> Option<PathBuf> {
 	let mut path = cli.launch_from_path.unwrap_or_default();
-	if is_valid_path(&path) {
+	// The reason from all our problems was path being ""
```
ButterscotchV, line 54: Could this comment explain what the problem was a little? It's a bit odd mentioning it caused problems but not *what* the problems were.

ButterscotchV (commented):
I'll approve under the condition that Butlah approves as well.

gui/src-tauri/src/main.rs
```diff
@@ -282,7 +294,19 @@ fn valid_java_paths() -> Vec<(OsString, i32)> {
 				.expect("Failed on executing a Java executable")
 				.code()
 				.map(|code| (p, code))
-				.filter(|(_p, code)| *code >= MINIMUM_JAVA_VERSION)
+				.filter(|(_p, code)| {println!("{}", code); *code >= MINIMUM_JAVA_VERSION})
```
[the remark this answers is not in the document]
ImUrX: thanks for reminding me, i left this on accident

gui/src-tauri/src/main.rs
ImUrX: i ended up changing this with an if let

ImUrX:
this is ready for merge

TheButlah (changes requested):
Replacing the slimevr-ui.exe in my installation (everything except gui was installed from the installer), I got:
image

No logs, no popups, no anything. I did enable RUST_LOG=trace

gui/src-tauri/src/main.rs
```diff
@@ -286,3 +299,15 @@ fn valid_java_paths() -> Vec<(OsString, i32)> {
 		})
 		.collect()
 }
+
+fn executable<W: AsRef<OsStr> + ?Sized>(bin: &W) -> OsString {
+	let bin = bin.as_ref();
+	if cfg!(windows) {
+		let mut new = OsString::with_capacity(bin.len() + 4);
```
TheButlah, line 306: We are working in a codebase that doesn't have rust developers, please focus on keeping the code as simple and readable as possible. just do `OsString::new()` - reserving a particular capacity is not necessary here

TheButlah (changes requested):
I'll give a more thorough review and find the cause of the bug when I get home. I still have only glanced at this.

gui/src-tauri/src/main.rs
ImUrX: i would do this even in js and it would be less unreadable, ``OsString::with_capacity`` is very verbose tbh

TheButlah (commented):
It works! Reviewing code now to see if it can be simplified. Will approve soon otherwise

gui/src-tauri/src/main.rs
```diff
@@ -286,3 +300,15 @@ fn valid_java_paths() -> Vec<(OsString, i32)> {
 		})
 		.collect()
 }
+
+fn executable<W: AsRef<OsStr> + ?Sized>(bin: &W) -> OsString {
```
TheButlah, line 304: We don't need to be so generic here. Just accept an `&OsStr`. These bounds over complicate things.

TheButlah, line 304: There is no comment on this function. What does it do? How about naming it `maybe_exe` and say "appends .exe on windows".

Better yet, why is this even a function? Its not like its doing anything particularly reusable. Just have 
```rust
const JAVA_EXE: &OsStr = if cfg!(windows) {
  "java.exe"
} else {
  "java"
};
```

gui/src-tauri/src/main.rs
```diff
@@ -196,6 +206,7 @@ fn main() {
 							.emit_all("server-status", emit_me)
 							.expect("Failed to emit");
 					}
+					println!("Java server receiver died");
```
TheButlah, line 209:
```suggestion
					log::error!("Java server receiver died");
```

TheButlah (changes requested):

gui/src-tauri/src/main.rs
```diff
@@ -1,11 +1,12 @@
...
-use std::env;
 use std::ffi::{OsStr, OsString};
 use std::io::Write;
+#[cfg(windows)]
+use std::os::windows::process::CommandExt;
 use std::panic;
 use std::path::PathBuf;
 use std::process::{Child, Stdio};
-use std::str::FromStr;
+use std::{env, fs};
```
TheButlah, line 9:
```suggestion
use std::env;
```

gui/src-tauri/src/main.rs
ImUrX: didnt know if would be needed for smth else

TheButlah: dont add until you know you need it, when the alternative is so much simpler. You can always refactor into a new function

TheButlah (approved):

TheButlah (approved):

PR #346, opened 2022-12-08
Add keys and French translation for legtweaks+gesture-control

Stermere:


Eirenliel:
We should decided on what how to deal with new added strings in other translations. I think you should not add them.

Stermere:
> We should decided on what how to deal with new added strings in other translations. I think you should not add them.

Ok, I agree actually doing it the way I just did will lead to weird issues in the future. I only made this PR since Erimel wanted this done. I can just send him the fork

Louka3000:
> We should decided on what how to deal with new added strings in other translations. I think you should not add them.

Added french translation

PR #351, opened 2022-12-09
Show git commit in version

### LINKED ISSUE #317: Show git version in GUI (and cli) and from server CLI
In case it is helpful, I recently did this in rust: https://github.com/SlimeVR/SlimeVR-Rust/pull/75. Might be helpful for tauri, not sure about for the ts side though.

Probably for java you can find a library to do something similar.

ImUrX:
Closes #317


ImUrX:
I'm currently using ``shadow-rs`` for showing git commit, but I think it's a little overkill. I originally thought it had nice integration with clap but it shows too much information:
```
./slimevr-ui --version
slimevr_ui 
pkg_version:0.0.0
branch:git-version
commit_hash:580156b5
build_time:2022-12-09 07:54:12 -03:00
build_env:rustc 1.65.0 (897e37553 2022-11-02),1.65-x86_64-unknown-linux-gnu
```
So I ended up making my own constant and now it looks like this:
```
./slimevr-ui --version
slimevr_ui 0.5.0-580156b5
```
It has its benefits as a build info tool but idk, y'all be the judge.

ImUrX:
I removed clap integration with tauri, unless we actually think of using it from the typescript-side I don't think it's necessary

ImUrX:
need to fix the spotless error on windows but its ready for review

gui/src/components/TopBar.tsx
```diff
@@ -36,7 +36,7 @@ export function TopBar({
             className="mx-2 flex justify-around flex-col text-standard-bold text-status-success bg-status-success bg-opacity-20 rounded-lg px-3"
             data-tauri-drag-region
           >
-            v{packagejson.version}
+            v{packagejson.version}-{window.__COMMIT_HASH__}
```
ButterscotchV, line 39: Do we want a commit hash to be shown for stable releases? I'd like to get some other opinions on this.

server/src/main/java/dev/slimevr/Main.java
```diff
@@ -42,6 +44,10 @@ public static void main(String[] args) {
 			formatter.printHelp("slimevr.jar", options);
 			System.exit(0);
 		}
+		if (cmd.hasOption("version")) {
+			System.out.println("SlimeVR Server " + VERSION + "-" + BuildConfig.GIT_COMMIT_HASH);
+			System.exit(0);
+		}
```
ButterscotchV, line 50: I know it's small, but could we just make a quick little method for printing the version instead of duplicating? 🥺

ButterscotchV (changes requested):

server/src/main/java/dev/slimevr/Main.java
ImUrX: Both prints are different, that goes directly to stdout and the other uses a logger :/

ImUrX: Unless you are talking about a function for making that string, im gonna die XD

gui/src-tauri/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "slimevr"
-version = "0.0.0"
+version = "0.5.0"
```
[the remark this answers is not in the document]
ImUrX: I have an idea but let me test it first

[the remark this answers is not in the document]
TheButlah: I think you should be using version 0.0.0, for this exact reason.

gui/src/components/TopBar.tsx
TheButlah: If its tagged in git, it should use the tag. If its not tagged, it should use the hash.
We could *also* include the hash in the development view though.

gui/src-tauri/Cargo.toml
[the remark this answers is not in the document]
ImUrX: I need to make a const function for this then, I need to evaluate what to return :/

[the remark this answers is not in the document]
ImUrX: This isnt a newbie friendly issue let me tell you that XD

gui/src-tauri/Cargo.toml
```diff
@@ -23,19 +23,23 @@ custom-protocol = ["tauri/custom-protocol"]
 [build-dependencies]
 tauri-build = { version = "1.2", features = [] }
 cfg_aliases = "0.1"
+shadow-rs = "0.18"
```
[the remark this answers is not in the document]
ImUrX: thats bindings for libgit2 tho, idk how to show current tag with ``git describe``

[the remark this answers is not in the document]
TheButlah: it does not error out if there is no tag, I'm doing this already in the overlay. If there is no tag, it will use the commit hash

server/src/main/java/dev/slimevr/Main.java
ButterscotchV: Oh you're right, why isn't it using the logger? If it's intentionally not using the logger for some reason, then just resolve this

server/src/main/java/dev/slimevr/Main.java
```diff
@@ -13,7 +13,9 @@
 
 public class Main {
 
-	public static final String VERSION = "0.5.1";
+	public static final String VERSION = BuildConfig.GIT_VERSION_TAG.isEmpty()
+		? BuildConfig.GIT_VERSION_TAG
+		: BuildConfig.GIT_COMMIT_HASH;
```
ButterscotchV, line 18: This looks like the wrong order to me
```suggestion
	public static final String VERSION = !BuildConfig.GIT_VERSION_TAG.isEmpty()
		? BuildConfig.GIT_VERSION_TAG
		: BuildConfig.GIT_COMMIT_HASH;
```

ButterscotchV (changes requested):

gui/src-tauri/Cargo.toml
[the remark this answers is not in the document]
ImUrX: Ok I understand how ``git describe`` works, the default config of git_version is nice because it appends ``-modified`` if it's dirty. So ``git describe`` literally what it does is describe the current commit, it will use a tag apparently when it's used on a clean commit with a tag or if not it will use the hash.
My currenty implementation is by getting both separately and all versions on all parts look like this:
```
slimevr.exe -V   
slimevr git-8e5e142c
```
if a tag exists:
```
slimevr.exe -V   
slimevr TAG_NAME
```

The ``git_version!`` one will look like
```
slimevr.exe -V   
slimevr 8e5e142c-modified
```
if a tag exists (and it's not dirty):
```
slimevr.exe -V   
slimevr TAG_NAME
```
My version of the repo will always be dirty because of the submodules :sob:

TheButlah (changes requested):
Opinion:
git_version still does it better, because it shows if the code is modified, and doesn't use libgit2. The originally proposed version does not do that. Also, git_version is simple, simpler than shadow. It also has basically no dependencies.

The objective of the rust backend is to keep the code as simple as possible - its basically just a small script that launches the frontend. So any additional complexity here is really over engineering. Just use git_version with its default invocation.

gui/src-tauri/Cargo.toml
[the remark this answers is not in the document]
TheButlah: see review comment

gui/src-tauri/src/main.rs
```diff
@@ -33,16 +34,26 @@ static POSSIBLE_TITLES: &[&str] = &[
 	"never gonna let you down",
 	"uwu sowwy",
 ];
+shadow!(build);
+// Tauri has a way to return the package.json version, but it's not a constant...
+const VERSION: &str = if build::TAG.is_empty() {
+	formatcp!("git-{}", build::SHORT_COMMIT)
```
TheButlah, line 40: can't you just use `concat!()` here? Avoids the need for another dependency

gui/src-tauri/src/main.rs
```diff
@@ -33,16 +34,26 @@ static POSSIBLE_TITLES: &[&str] = &[
 	"never gonna let you down",
 	"uwu sowwy",
 ];
+shadow!(build);
+// Tauri has a way to return the package.json version, but it's not a constant...
+const VERSION: &str = if build::TAG.is_empty() {
```
TheButlah, line 39: Need to add logic for detecting if working tree is modified

gui/src-tauri/Cargo.toml
```diff
@@ -23,19 +23,23 @@ custom-protocol = ["tauri/custom-protocol"]
...
 log = "0.4"
-clap-verbosity-flag = "1"
+clap = { version = "4.0.29", features = ["derive"] }
+clap-verbosity-flag = "2"
 tauri-plugin-window-state = "0.1.0"
 rand = "0.8.5"
 tempfile = "3"
 which = "4.3.0"
 open = "3"
+shadow-rs = "0.18"
```
TheButlah, line 41: Need to disable the libgit2 feature. It should just call git from the PATH. Might turn off all default features tbh - tzdb probably is not necessary.

gui/src-tauri/Cargo.toml
```diff
@@ -23,19 +23,23 @@ custom-protocol = ["tauri/custom-protocol"]
...
+const_format = "0.2.30"
```
TheButlah, line 42: not necessary if you use `concat!()`

TheButlah (changes requested):

gui/src-tauri/Cargo.toml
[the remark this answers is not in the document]
TheButlah: resolving, as we decided to use shadow.

gui/src-tauri/src/main.rs
ImUrX: Nah concat doesnt work aside from literals

ImUrX:
Hellish PR, became more than expected because migrated from using separate versions for each part of the codebase to using git tags

Eirenliel (approved):

PR #353, opened 2022-12-09
Create a mapping for remote sensor id to internal.

carl-anders:
This is an up-to-date version of #271 .

I will keep this branch updated to support those with trackers that do not have a contiguous list of sensor id's.
The reason why this is a pull request is for users with this issue to easier find this build.

.jar build as of 2023-01-06:
[slimevr.zip](https://github.com/SlimeVR/SlimeVR-Server/files/10361904/slimevr.zip)


carl-anders:
I don't think this should be marked as a draft, because any user finding this merge request will assume that the solution doesn't work.

Eirenliel:
we should just finish it and merge

carl-anders:
It is finished?

Eirenliel:
Then we should merge it

Eirenliel:
Is it superceeded by #464?

carl-anders:
#464 is a better solution if it tests out okay.

Eirenliel:
If they solve the same problem, #464 is better. Closing this.

PR #359, opened 2022-12-10
Fix page not opening

### LINKED ISSUE #345: ”I'm having trouble connecting!” button has trouble
In v0.5.0, ”I'm having trouble connecting!” button in "Connect trackers"(in setup wizard) causes the window filled with dull blue and doesn't respond.
When I used v0.4.0, this problem doesn't happen.

macluxHD:
The page opens now when pressing "I'm having trouble connecting!"  Button, even if typedState is undefiend.

Fixes #345

### DIFF
--- a/gui/src/components/commons/ArrowLink.tsx
+++ b/gui/src/components/commons/ArrowLink.tsx
@@ -6,11 +6,13 @@ import { ArrowLeftIcon, ArrowRightIcon } from './icon/ArrowIcons';
 export function ArrowLink({
   to,
   children,
+  state,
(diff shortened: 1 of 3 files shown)


loucass003 (changes requested):
This is not enough. We need the button in the wifi window to set the state with a port
Auto for now will be best

macluxHD:
The Button now sets the SerialPort State to Auto

Eirenliel (approved):

loucass003 (approved):

PR #366, opened 2022-12-12
Rename upper/lower legs to thigh/ankle

Louka3000:
I believe using upper/lower_leg is better in the code, but can be very confusing for users.
This
1. Makes it clear where a tracker should be (e.g: users should mount on ankles, not calves)
2. Makes the text less verbose and easier to understand.
3. Make translating into other languages easier without weird or long formulations (e.g: "left upper leg" = "haut de la jambe gauche"...)

Please give your thoughts if you disagree to rename these.

### DIFF
--- a/gui/src/i18n/en/translation.json
+++ b/gui/src/i18n/en/translation.json
@@ -15,8 +15,8 @@
     "RIGHT_UPPER_ARM": "Right upper arm",
     "RIGHT_LOWER_ARM": "Right lower arm",
     "RIGHT_HAND": "Right hand",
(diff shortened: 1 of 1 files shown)


ZRock35 (approved):
makes sense to me!

ButterscotchV (approved):
Makes sense, I'd guess it's either easier to understand or about the same, and if it makes translating easier or more consistent then I think it'll be good to make the switch.

TheButlah:
yes pls

Eirenliel (approved):

PR #367, opened 2022-12-12
Fix language config for pl, zh & pt-BR languages

carl-anders:
Misspelling was corrected for other languages already, but not these.


Louka3000 (approved):
Indeed. Tested, works :3

PR #374, opened 2022-12-12
Create new Japanese translation

tamakiti3010:
A Japanese translation of the GUI was created.


Eirenliel (approved):

gui/src/i18n/config.ts
```diff
@@ -44,6 +45,10 @@ export const langs = [
     name: '🥺 Engwish~ OwO',
     key: 'owo',
   },
+  {
+    name: '日本語',
```
TheDevMinerTV, line 49: Could you please add the flag as well?

tamakiti3010: OK
I have added it.

Eirenliel (approved):

PR #377, opened 2022-12-13
Include font for colour flags on Windows

carl-anders:
Twemoji font is CC-BY Twitter - https://twitter.github.io/twemoji/

Open questions:

1. ~~Where to put CC-BY attribution? Is it enough to have it in the commit message?~~
2. ~~I just chose a high up div in App.tsx, there's other options to include it:~~
    1. Put it as a script tag in index.html
    2. Extend the Dropdown component to take a classname (doing check in LangSelector to send to Dropdown)
    3. **Chosen:** Use the colour flags font for all platforms


carl-anders:
CC BY 4 license and resource information added to ThirdPartyNotices.txt file, striking open question 1.

gui/src/App.tsx
```diff
@@ -152,14 +153,15 @@ function App() {
     };
   }, []);
 
+  const winClass = !/windows/i.test(navigator.userAgent) ? 'win_flags' : '';
```
TheDevMinerTV, line 156: can't we just use the twitter emojis all the time?

carl-anders: That's an option too, do people on linux/mac care much about having their native OS flags? I don't know if there's even much of a difference.

carl-anders:
Changed to use the font on all platforms, simplified logic a lot. Letting the reviewer decide if that's the right way :)

Eirenliel (approved):

PR #378, opened 2022-12-13
Filtering amount check

Louka3000:
Some users experienced a bug (totally not my fault uwu~) that put their filtering amount to 1000% in an older server version, and won't go back on newer versions unless they manually lower it in GUI. This fixes that and adds a nice check just in case.

### DIFF
--- a/server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
+++ b/server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
@@ -1,10 +1,7 @@
 package dev.slimevr.config;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.fasterxml.jackson.databind.node.DoubleNode;
-import com.fasterxml.jackson.databind.node.JsonNodeFactory;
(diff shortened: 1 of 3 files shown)


ButterscotchV (approved):

loucass003 (changes requested):
use the versioning of the config to fix it
it would be cleaner and used only once
in the config logic
it makes more sense that way
than having a check every time 
you add a new branch here like
https://github.com/SlimeVR/SlimeVR-Server/blob/main/server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java#L26
```
if (version == 2) {
   ObjectNode filtersNode = (ObjectNode) modelData.get("filters");
   if (filtersNode != null && filtersNode.get("amount").asFloat() > 2f)
     filtersNode.set('amount', 0.2);

} else if (version < 2) {
   // old code
}
```
 
not tested but that should brobably be like that 
and you should also update
https://github.com/SlimeVR/SlimeVR-Server/blob/main/server/src/main/java/dev/slimevr/config/VRConfig.java#L16
i think you need to set both currentVersion = "2", defaultDeserializeToVersion = "1" to 3

Eirenliel (approved):

PR #383, opened 2022-12-13
Basic workaround for using multiple sets on one network

ButterscotchV:
I don't exactly like how this is done and it's kinda hacky, but an increasing number of people depend on it and the alternative still doesn't exist after at least 6 months now. It's really small, but I don't want to keep individually maintaining this for forever.


loucass003:
i think it is a good alternative until we implement the better way

Eirenliel (approved):

PR #390, opened 2022-12-14
Make port binding error in onError less scary

TheButlah:
It looks like the server crashes, when actually having this sort of error is expected (I think). We should be retrying the connection in that even (which we do, I think. someone fact check me.

Someone fact check that we actually retry the websocket when it fails to bind. If we don't, I would rather fix that before doing this.


TheButlah:
Edit: I think I'm going to mark this as draft and open a new PR, as we *do not* retry the bind.

Sacred-Shadow:
This thing is already over 2 years old. Completely covered in dust.
Is this art, or can it go? 🤔

PR #394, opened 2022-12-15
Remove ReferenceAdjustedTracker

Louka3000:
Merge ReferenceAdjustedTracker into IMUTracker.

This improves the code and 0. and I kinda need this to be able to implement stuff without doing cursed code.


Eirenliel (commented):
There is an important reason it wasn't part of IMU Tracker: because it's not. It could be applied to any tracker, potentially, including trackers that report their absolute rotation and position, that's why it was "layered" like this...

Louka3000:
I know that was the purpose, but, for the example of Vive trackers, we don’t need it, as they’re already going the right direction… all it needs is 1 offset when resetting which is like 2 lines of code…
We don’t have any example currently that would require ReferenceAdjustedTracker afaik.

Eirenliel:
~~here's my example: :nya_gun:~~

Louka3000:
Unless we have a concrete example of Referenceadjustedtracker being useful separate from IMUTracker, there’s no reason to keep it, and for what 0. and I wanna so, it’s much simpler to merge it (and it makes the whole code simpler in general! @loucass003 can back me up on that one :p)

PR #396, opened 2022-12-16
Draft Latin American Spanish Translation

Fowels:
This is the ongoing Spanish LatinAmerican translation.

### DIFF
--- a/gui/src/i18n/config.ts
+++ b/gui/src/i18n/config.ts
@@ -1,6 +1,7 @@
 import i18next from 'i18next';
 import { initReactI18next } from 'react-i18next';
 import en from './en/translation.json';
+import es419 from './es-419/translation.json';
 import fr from './fr/translation.json';
 import ko from './ko/translation.json';
 import pl from './pl/translation.json';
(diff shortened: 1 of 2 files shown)


ReadyPlayerEmma:
I've been discussing with @Fowels about this. I think the appropriate way to do translation for Latin America is to use es-419 as the actual code and folder for the translations. Then, a fallback function can be configured such that Spanish for Spain uses one translation, and all other Spanish countries will use es-419 which is the correct code to summarize Latin America.

How to implement this is described here, and an example is in the last comment: https://github.com/i18next/i18next/issues/1506

You can also see on Wikipedia they make mention of es-419 here: https://en.wikipedia.org/wiki/Language_code
image

This code is referenced by Microsoft and also Unreal Engine as the way to handle Latin American Spanish, and then mapping the individual countries to this code.

ImUrX:
Yes, we know about IETF language tags, I don't see anything wrong from using the ``UN M49`` code for Latin America and the Caribbean xd.
And yes languages fallbacks are a good idea, they are not simple programatically in cases like this but yeah.

btw I wouldnt have give unreal engine and microsoft as an example of usage tho...

Eirenliel:
If this is not relevant after #415, please close it.

Issue #398, opened 2022-12-16
Switching the codebase to Kotlin


ButterscotchV:
This is a proposal to switch the server codebase to Kotlin! I'd like to gauge the general opinion on this and what might be any pros/cons about switching.

I propose for the switch, we can do each part of the server without an active PR or active development first, then slowly as PRs clear up, we can switch those remaining classes to Kotlin without loss of development time or huge merge conflicts. This switch should be fairly easy and for the most part should be fully automated.

Pros:
 - Non-nullable by default. This provides way better protection against null pointer exceptions.
 - Less verbose, more hinting. This reduces the overall size of the code and makes it much easier to look through and review. When there's less to look at, the more likely you are to spot problems.
 - Loads of shortcuts for ease of use. For example, replacing a full line of a null check to just `?.`.

Cons:
 - Would be a fairly long transition from Java to Kotlin for the codebase.
 - Everyone would need to learn how to use Kotlin, though it is fairly similar to Java.
 - Reviews may be a bit slower as people get used to the change.
 - May not work with all the tools that we currently use (ex. Spotless maybe?).

You can read more about Kotlin here: https://kotlinlang.org/docs/home.html

The relevant Discord post can be found here: https://discord.com/channels/817184208525983775/1053164760071417946

TheButlah:
We discussed this in discord for a while. Here is a summary:

So we discussed this in discord more. Here is a summary, which I ran past the main participants of the conversation which were @TheButlah @Louka3000 @ButterscotchV @ImUrX to make sure I didn't miss something.

**Motivation for using Kotlin:**
*  **The primary motivator was the null safety of kotlin.**
* We observed numerous NPE issues in the last month alone that broke the server. Theory is that kotlin would eliminate or reduce these.
* It was also discussed that kotlin was less verbose than java, and had nice operators like ? and did var by default.
* There was some discussion of kotlin having an easier concurrency story than java with the use of coroutines and promises.
* There was less discussion about other reasons for use of kotlin, but it seemed that most people who participated in the conversation were more interested in kotlin than java or at least curious/open to the idea.


**Ability to incrementally port to Kotlin:**
* Kotlin lets you port the code on a per-file basis.
* You get to call java from kotlin and vice versa as a first class citizen. So no FFI boilerplate or anything is needed.
* Both can be compiled into same library, same jar, etc.
* There are tools to autoconvert from java to kotlin. This makes it really easy as most of the work can be automated.
[...]

PR #401, opened 2022-12-16
Kotlin revolution

ImUrX:


TheButlah:
I don't think we want the `src/main/java` vs `src/main/kotlin` style. I'd like it to be `.kt` and `.java` files in the same folders preferably.

ImUrX:
Yeah I'm not against the first part.

I didnt touch VRServer tho, only touched Main in purpose to give an example Kotlin file

ImUrX:
how is this?

TheButlah:
This looks good to me.

As an aside: I know things like syntax don't truly matter. But I look at this code and the `var my_var: Type = asdf` is really nice to look at :)

TheButlah (approved):

Eirenliel (approved):

ImUrX:
Related to #398

PR #402, opened 2022-12-16
Improve French translation

Louka3000:
Greatly improve French translation.
Note: we changed the tracker status.ok from "connected" to "ok", since that's less ambiguous (in english/furry as well) (note: Futura said they approve that change via Discord)


HeySora (approved):
image

PR #408, opened 2022-12-19
Fix lang select not fitting again

ImUrX:
thought the app autoupdated when I ctrl+s but no. fit-content doesnt work and max needs to beused


ButterscotchV (approved):
😔

Eirenliel (approved):

PR #414, opened 2022-12-21
Use fluent format for translations

ImUrX:
Some languages have to make worse translations because of the current limitation that we use JSONs, I propose using [fluent](https://projectfluent.org/) will make translation a lot easier for these languages that have a limitation for such a thing.

The actual issue is that we reuse terms in different sentences and just remaking the key 60 times wont be nice for translators either. There is more info on cases like [locative in here](https://en.wikipedia.org/wiki/Locative_case).

Namings were changed from ``prop.prop-name`` to ``prop-prop_name`` as ``.`` means an attribute in fluent


ImUrX:
By the way, this PR will only include the conversion to ``fluent``, I will add the actual fluent syntaxing in another PR

ImUrX:
The JSONs should get removed when it's merged or right before merging so translators don't get mixed up

TheDevMinerTV (approved):

Eirenliel (approved):

Eirenliel:
Please work with other translators to move to the new format :firBless:

PR #435, opened 2022-12-30
Switch rotation to orientation

Louka3000:
Both orientation and rotation were used for mounting. 
Orientation being the end result rather than the process, it makes more sense to use that rather than rotation.
This switches all uses of "mounting rotation" to "mounting orientation"

Protocol PR: https://github.com/SlimeVR/SolarXR-Protocol/pull/84

### DIFF
--- a/gui/src/components/onboarding/pages/mounting/ManualMounting.tsx
+++ b/gui/src/components/onboarding/pages/mounting/ManualMounting.tsx
@@ -45,7 +45,7 @@ export function ManualMountingPage() {
       const assignreq = new AssignTrackerRequestT();
 
       assignreq.bodyPosition = td.tracker.info?.bodyPart || BodyPart.NONE;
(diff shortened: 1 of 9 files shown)


ImUrX (changes requested):
make the submodule point to the main branch on commit [``d96bf21``](https://github.com/SlimeVR/SolarXR-Protocol/commit/d96bf2136fd50b653799c9f3a925309798924272)

ImUrX (approved):
asked for permission first :p

ButterscotchV (approved):
Approved as long as `CurrentVRConfigConverter.java` has been tested.

Louka3000:
Yes I tested on my side extensively, but anyone else feel free to double test

ButterscotchV (approved):

Eirenliel (approved):

PR #439, opened 2022-12-31
Fix bug on window state save

### LINKED ISSUE #433: GUI stays on second monitor even if it doesn't exist
The GUI saves the monitor it's on, but it should check if that monitor is connected and fall back to the main monitor if it doesn't exist.

- Example 1: GUI was on main monitor, main monitor is disconnected so second monitor becomes the main monitor: everything is fine :3
- Example 2: GUI was on second monitor, main monitor is disconnected so second monitor becomes main monitor: GUI window is invisible and inaccessible as it's on the second monitor, but it doesn't exist.

@loucass003

ImUrX:
Fixes #433

### DIFF
--- a/gui/src-tauri/tauri.conf.json
+++ b/gui/src-tauri/tauri.conf.json
@@ -78,7 +78,7 @@
         "decorations": false,
         "transparent": false,
         "fileDropEnabled": false,
-        "visible": true
+        "visible": false
       }
     ],
     "security": {
--- a/gui/src-tauri/tauri.conf.json
+++ b/gui/src-tauri/tauri.conf.json
@@ -78,7 +78,7 @@
         "decorations": false,
         "transparent": false,
         "fileDropEnabled": false,
-        "visible": false
+        "visible": true
       }
     ],
     "security": {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.66"
+version = "1.0.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


gui/src-tauri/tauri.conf.json
```diff
@@ -78,7 +78,7 @@
         "decorations": false,
         "transparent": false,
         "fileDropEnabled": false,
-        "visible": true
+        "visible": false
```
Louka3000, line 81:
```suggestion
        "visible": true
```
This needs to be true or else the window is invisible .-.

Louka3000 (changes requested):
Doesn't work for me

gui/src-tauri/tauri.conf.json
ImUrX: That’s not what visible means, and weird I tested it in my side

Louka3000: The window is invisible if false and visible if true tho…

Louka3000: Nvm, after further testing, ig that doesn't matter.
This still doesn't fix the window bug for me tho

Louka3000 (approved):
Works now OWO

Eirenliel (approved):

Beanow:
Please note the PR you're referencing was closed: https://github.com/tauri-apps/tauri-plugin-window-state/pull/52#issuecomment-1369853832

> Hey :wave: Thank you for your contributions! We don't accept PRs for the individual plugin repos anymore since we're moving all of them to https://github.com/tauri-apps/plugins-workspace. It would be much appreciated if you could re-open it against that repo.

ImUrX:
Yeah. I know, I got the notification XD

PR #445, opened 2023-01-03
Only CI check on main push instead of all pushes

ImUrX:


Eirenliel:
Will it run if PR is updated?

ImUrX:
Yes, that's handled by the `pull_request` event

PR #447, opened 2023-01-03
Fix some fluent stuff

ImUrX:
- ``check-missing.js`` can be ran anywhere now
- add check on wrong language tags
- add english fallback for missing keys


Louka3000 (approved):
Thanks, works! <3

PR #449, opened 2023-01-04
server: Send accel and ref-adjusted rotations

0forks:
**Requires:**
- [x] https://github.com/SlimeVR/SolarXR-Protocol/pull/77
- [x] https://github.com/SlimeVR/SlimeVR-Server/pull/443

### DIFF
--- a/server/src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
+++ b/server/src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
@@ -276,9 +276,9 @@ public class IMUTracker
 	}
 
 	/**
-	 * Calculates zero-reference-adjusted rotation (with full/quick reset).
-	 * Same as {@link #getRotation(Quaternion)}, except rotation is aligned to
-	 * an identity quaternion instead of HMD and does not include mounting reset
+	 * Calculates zero-reference-adjusted rotation (with full/quick reset). Same
+	 * as {@link #getRotation(Quaternion)}, except rotation is aligned to an
+	 * identity quaternion instead of HMD and does not include mounting reset
 	 * and mounting orientation adjustments. Does not take drift compensation
 	 * into account.
 	 *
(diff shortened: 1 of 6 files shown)


server/src/main/java/dev/slimevr/vr/trackers/IMUTracker.java
```diff
@@ -396,6 +474,11 @@ public boolean hasPosition() {
 		return false;
 	}
 
+	@Override
+	public boolean hasAcceleration() {
+		return true;
```
Louka3000, line 479: What if an IMU doesn't have accel data? Would that cause any problem, or would the accel vector being (0, 0, 0) suffice?

0forks: With new firmware acceleration is always sent, unless explicitly disabled by the user in defines. For trackers with old firmware this will be a zero vector, yes.

Louka3000: Alright

Louka3000 (approved):

ButterscotchV (approved):
Seems good, the "reset" stuff in `IMUTracker.java` is a bit confusing to read, so I'd like it to be tested to be sure it still functions the same (and maybe others can check the code better)

Eirenliel (approved):
This changes rotation code, needs to be beta-tested. @Louka3000

Eirenliel (changes requested):
This changes rotation code, needs to be beta-tested. @Louka3000 

Also it doesn't build @0forks.

Eirenliel (approved):
Needs testing

Louka3000:
> Needs testing

Works fine on my side, and believe 0. made a lot of people on Discord test it as well (especially people using BMIs in the BMI thread)

0forks:
> > Needs testing
> 
> Works fine on my side, and believe 0. made a lot of people on Discord test it as well (especially people using BMIs in the BMI thread)

That was before server code changes, old code was calculating all that in the GUI and didn't touch `IMUTracker`.

Louka3000:
Anyway, I just tested in VR and both manual mounting and using mounting reset work just fine.

ButterscotchV (approved):

PR #450, opened 2023-01-04
gui: More info in tracker table and IMU visualizer

0forks:
**Requires:**
- [x] https://github.com/SlimeVR/SlimeVR-Server/pull/449
- [x] https://github.com/SlimeVR/SlimeVR-Server/pull/463
---

- Adds an IMU visualizer;

  image

- Adds a widget with toggles to the side: <details><summary>Show image</summary>
  image
</details>

- Extends tracker table with TPS, temperature and linear acceleration columns, less useful data columns are hidden behind a toggle;
- Temperature column is automatically hidden if there is no temperature data for any device;
- Adds a developer mode widget with toggles to set **high contrast** for better accessibility, increasing table update rate, basic filtering and sorting;
- Fixes a case when ping = 0 displays as an empty cell in the tracker table (wrangler, owotrack, anything local);
- Improves reaction to shaking when highlighting trackers, since the server can now send linear acceleration in addition to rotation data;
- Adds a warning highlight/icon for unassigned trackers, since reference/identity-adjusted rotations (after reset) only apply to assigned trackers:

  image

owo

### DIFF
--- a/gui/package.json
+++ b/gui/package.json
@@ -7,6 +7,7 @@
     "@fluent/react": "^0.14.1",
     "@fontsource/poppins": "^4.5.8",
     "@formatjs/intl-localematcher": "^0.2.32",
+    "@react-three/fiber": "^8.10.0",
     "@tauri-apps/api": "^1.2.0",
     "@vitejs/plugin-react": "^3.0.0",
     "browserslist": "^4.18.1",
@@ -25,7 +26,6 @@
     "postcss-normalize": "^10.0.1",
     "postcss-preset-env": "^7.0.1",
     "prompts": "^2.4.2",
-    "quaternion": "^1.4.0",
     "react": "^18.0.0",
     "react-dev-utils": "^12.0.0",
     "react-dom": "^18.0.0",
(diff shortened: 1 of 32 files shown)


Eirenliel:
> Adds a developer mode widget with toggles to set high contrast for better accessibility

is this for developer mode only? would be useful for everyone tbh

0forks:
> > Adds a developer mode widget with toggles to set high contrast for better accessibility
> 
> is this for developer mode only? would be useful for everyone tbh

For now the high contrast toggle affects only the numbers in the tracker table, which is accessible only with developer mode enabled. I had personally found the numbers hard to look at for extended periods of time with the text color set as "secondary", so it's an optional toggle. It's easy to remove the toggle and make it a default, but I'm not sure if many people would find it preferable.

Here's before and after comparison:
image

Adding high contrast mode globally is somewhat out of scope for this PR and I think general accessibility should be discussed separately, since it requires tweaking the whole UI and design guidelines.

TheButlah:
needs rebase

Eirenliel (approved):

PR #457, opened 2023-01-05
Add User Action packet. Allows for reset buttons on trackers and not just tap detection to reset.

carl-anders:
This is needed to support resetting by pressing a button on a joycon, as the normal tap detection system doesn't work well on joycons.

### DIFF
--- a/server/src/main/java/dev/slimevr/vr/trackers/udp/UDPPacket21Reset.java
+++ b/server/src/main/java/dev/slimevr/vr/trackers/udp/UDPPacket21Reset.java
@@ -1,0 +1,32 @@
+package dev.slimevr.vr.trackers.udp;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+
+public class UDPPacket21Reset extends UDPPacket {
+
+	public static final int RESET = 0;
+	public static final int RESET_YAW = 1;
+	public static final int RESET_MOUNTING = 2;
+
+	public int type;
+
+	public UDPPacket21Reset() {
+	}
+
+	@Override
+	public int getPacketId() {
+		return 21;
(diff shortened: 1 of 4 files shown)


Louka3000:
Would another solution be to make the g-force required configurable? Altho this allows individual devices to have their own threshold set by the manufacturer/software dev, so it's good as well.

carl-anders:
One problem with that is for a user with many phone trackers and joycons for ankles. The tap reset system doesn't allow ankle trackers to be used for tapping. It's nice to be able to use the actual buttons to reset too. By default in wrangler I will use the Up and B buttons as they are shielded from accidental presses by the thumbsticks.

TheButlah:
From discord, Eiren requested to turn this into a generic "user action" packet , so that the action can be remapped in the server later

TheButlah (changes requested):
.

server/src/main/java/dev/slimevr/vr/trackers/udp/UDPPacket21Reset.java
```diff
@@ -0,0 +1,32 @@
+package dev.slimevr.vr.trackers.udp;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+
+public class UDPPacket21Reset extends UDPPacket {
```
Eirenliel, line 7: Can you please rename the packet to UDPPacket21UserAction and update action IDs so that RESET starts from 2 insted of 0. This will allow the system to be expanded in the future, I have a few ideas for it and will create a ticket later.

Eirenliel (changes requested):

Eirenliel (approved):

PR #458, opened 2023-01-06
Link to the slimevr.dev installer link instead of GitHub directly

ButterscotchV:


Eirenliel:
I think it's better if github links to github without any third-party dependencies, so i don't want to update the readme. We should also remove run.bat, it's not supplied with the build anymore.

PR #464, opened 2023-01-08
Support arbitrary sensor id's for devices

carl-anders:
So I wanted to check how much work it was to change the device tracker storage to use a hashmap instead of a list, enabling arbitrary sensor id's. Turns out SolarXR and the GUI already supported this. I think I have gone through the UI code regarding this and found no place where the this would break things. So the needed changes were very minimal.

The only minor concern is in DataFeedBuilder, where the first tracker is used, however, it only seems to get the hardware info like temperature, voltage etc, and these are all the same for every tracker. I'm not 100% sure on this however.

I have tested it in VR, sending normal sensor id's and arbitrary id's. Seems to work fine.

Further testing is needed though, making sure that both AutoBone and the SlimeVR Overlay still works. Would love some help testing that. :)

### DIFF
--- a/server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
+++ b/server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
@@ -103,7 +103,7 @@ public abstract class SteamVRBridge extends ProtobufBridge<VRTracker> implements
 			device
 		);
 
-		device.getTrackers().add(tracker);
+		device.getTrackers().put(0, tracker);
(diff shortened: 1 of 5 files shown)


Eirenliel (approved):
Should work, good to merge after some testing.

Louka3000:
Autobone and overlay work fine, as they don't use tracker IDs anyway

Louka3000 (approved):

ButterscotchV (approved):

PR #465, opened 2023-01-08
Dutch translation

Vyolex:
Rough draft for the Dutch translation, tried to fix as much as possible but some translations still need work.

**Added:**
- Dutch translation file

**Changed:**
- `config.tsx`


ImUrX (commented):
seems to have some missing translations like L155

Vyolex:
Looks like I forgot that one, will add it along with the new keys tonight

Eirenliel (approved):

PR #468, opened 2023-01-09
Rename "wifi" to "Wi-Fi" and "SSID" to "Wi-Fi" name

Louka3000:
- Wi-Fi is the official spelling, see https://english.stackexchange.com/questions/41180/how-do-you-spell-wifi-wi-fi-wifi
- SSID is causing a lot of confusion amongst users who don't know what it means. Wi-Fi name is easier to understand.

### DIFF
--- a/gui/public/i18n/en/translation.ftl
+++ b/gui/public/i18n/en/translation.ftl
@@ -67,9 +67,9 @@ reset-quick = Quick Reset
 
 ## Serial detection stuff
 serial_detection-new_device-p0 = New serial device detected!
-serial_detection-new_device-p1 = Enter your WiFi credentials!
+serial_detection-new_device-p1 = Enter your Wi-Fi credentials!
 serial_detection-new_device-p2 = Please select what you want to do with it
-serial_detection-open_wifi = Connect to WiFi
+serial_detection-open_wifi = Connect to Wi-Fi
 serial_detection-open_serial = Open Serial Console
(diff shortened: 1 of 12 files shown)


gui/public/i18n/en/translation.ftl
```diff
@@ -366,18 +366,18 @@ onboarding-skip = Skip setup
...
 
-## WiFi setup
+## Wi-Fi setup
 onboarding-wifi_creds-back = Go Back to introduction
-onboarding-wifi_creds = Input WiFi credentials
+onboarding-wifi_creds = Input Wi-Fi credentials
 # This cares about multilines
 onboarding-wifi_creds-description =
-    The Trackers will use these credentials to connect wirelessly.
+    The Trackers will use these credentials to connect wireless.
```
ImUrX, line 374: The Trackers will use these credentials to connect ``wireless``?

ImUrX (changes requested):

gui/public/i18n/en/translation.ftl
Louka3000: ugh, I assumed IDEA caught an error and didn't double check smh 
image
Thanks

ImUrX (approved):

ButterscotchV (approved):
Could add SSID in brackets, but if that makes it too long or unecessary, then it's fine without too

PR #469, opened 2023-01-09
Add .editorconfig

ImUrX:
This will suggest the editor to use the configurations specified on the file depending on the file patterns

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+root = true
+
+[*]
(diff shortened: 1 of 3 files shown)


TheButlah (approved):

ButterscotchV (approved):
Might we also want a rule for encoding to UTF-8? This is fine though.

TheButlah (approved):

Eirenliel (approved):

PR #493, opened 2023-01-19
Adding some missing deutsch translations

unlogisch04:


Louka3000:
For body proportions, is capitalizing each word intentional?

unlogisch04:
good question. I don't know. We start each sentence with a capital letter.
But left right is not a noun, it is an adjective. Since it is at the beginning it was probably capitalized. So I think that's intentional.
you could write left and right small in that case too, because it is not a sentence.

TheDevMinerTV:
Yes, this is fine in the ui

TheDevMinerTV (approved):

Louka3000 (approved):

Philip6294 (approved):

ImUrX (approved):

ImUrX (approved):

PR #494, opened 2023-01-20
Better JVM detection

ImUrX:
Added conditionals depending on OS, now Linux does it's own JVM detection based on ``/usr/lib/jvm``.
Also left a macOS one because it makes the ``#[cfg()]`` a lot cleaner and I just need to actually test it (I know where macOS saves them but I still want to test it).
I prioritized the ``APPDIR`` env just in case, because if we are using AppImage, we should trust the jar we have instead of a relative dir one.
I also kind of crept in some optimizations on the release build for less bin size... :eyes:

### DIFF
--- a/gui/src-tauri/Cargo.toml
+++ b/gui/src-tauri/Cargo.toml
@@ -11,9 +11,6 @@ rust-version = "1.65"
 default-run = "slimevr"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-[profile.release]
-lto = true
-codegen-units = 1
 
 [features]
 # by default Tauri runs in production mode
--- a/gui/src-tauri/Cargo.toml
+++ b/gui/src-tauri/Cargo.toml
@@ -11,6 +11,9 @@ rust-version = "1.65"
 default-run = "slimevr"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[profile.release]
+lto = true
+codegen-units = 1
 
 [features]
 # by default Tauri runs in production mode
@@ -23,7 +26,7 @@ custom-protocol = ["tauri/custom-protocol"]
 [build-dependencies]
 tauri-build = { version = "1.2", features = [] }
 cfg_aliases = "0.1"
-shadow-rs = "0.19"
+shadow-rs = "0.20"
 
 [dependencies]
 serde_json = "1"
@@ -36,10 +39,12 @@ clap-verbosity-flag = "2"
 tauri-plugin-window-state = { git = "https://github.com/tauri-apps/tauri-plugin-window-state", rev = "refs/pull/52/head" }
 rand = "0.8.5"
 tempfile = "3"
-which = "4.3.0"
+which = "4.3"
+glob = "0.3"
 open = "3"
-shadow-rs = { version = "0.19", default-features = false }
+shadow-rs = { version = "0.20", default-features = false }
 const_format = "0.2.30"
+cfg-if = "1.0.0"
 
 [target.'cfg(windows)'.dependencies]
 win32job = "1"
--- a/gui/src-tauri/Cargo.toml
+++ b/gui/src-tauri/Cargo.toml
@@ -41,7 +41,7 @@ glob = "0.3"
 open = "3"
 shadow-rs = { version = "0.20", default-features = false }
 const_format = "0.2.30"
-cfg-if = "1.0.0"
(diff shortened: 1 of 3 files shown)


ImUrX:
nvm, gonna delete the profile thing. It's getting ignored :sob:

Eirenliel (approved):

TheButlah (approved):
haven't tested it, but seems fine. Double check that the comments on line 301 are intended to be commented out @ImUrX

kitlith (approved):
LGTM, though you might want to consider inspecting/respecting the JAVA_HOME environment variable at some point.

gui/src-tauri/Cargo.toml
```diff
@@ -36,10 +36,12 @@ clap-verbosity-flag = "2"
...
 rand = "0.8.5"
 tempfile = "3"
-which = "4.3.0"
+which = "4.3"
+glob = "0.3"
 open = "3"
-shadow-rs = { version = "0.19", default-features = false }
+shadow-rs = { version = "0.20", default-features = false }
 const_format = "0.2.30"
+cfg-if = "1.0.0"
```
TheButlah, line 44:
```suggestion
cfg-if = "1"
```

gui/src-tauri/src/main.rs
```diff
@@ -64,6 +63,14 @@ fn get_launch_path(cli: Cli) -> Option<PathBuf> {
 		}
 	}
 
+	// This is only for AppImage
+	if let Some(appimage) = env::var_os("APPDIR") {
+		let path = PathBuf::from(appimage);
+		if is_valid_path(&path) {
```
TheButlah, line 69: change `is_valid_path(path: &PathBuf)` to `is_valid_path(path: &Path)`. No need to have a reference to an owned type when the ref type is better.

gui/src-tauri/src/main.rs
```diff
@@ -64,6 +63,14 @@ fn get_launch_path(cli: Cli) -> Option<PathBuf> {
 		}
 	}
 
+	// This is only for AppImage
```
TheButlah, line 66:
```suggestion
	// AppImage passes the path in `APPDIR` env var.
```

gui/src-tauri/src/main.rs
TheButlah, line 69: you call the same code 4 times. Why not just construct a list from each possible path source, loop on it, and return on first `Some`.

gui/src-tauri/src/main.rs
```diff
@@ -306,7 +305,23 @@ fn valid_java_paths() -> Vec<(OsString, i32)> {
 
 	// Otherwise check if anything else is a supported version
 	let mut childs = vec![];
-	for java in which_all(JAVA_BIN).unwrap() {
+	cfg_if::cfg_if! {
+		if #[cfg(target_os = "macos")] {
```
TheButlah, line 309: Use `cfg!()` bool instead of feature gating, to avoid complexity of nesting introduced by `cfg_if` as well as avoid separate code paths. The compiler is going to treat the `if true {}` as a const anyway and compile out the dead branch anyway.

TheButlah (changes requested):

gui/src-tauri/src/main.rs
ImUrX: Not related to the current PR

gui/src-tauri/src/main.rs
ImUrX: just tried it, not worth it. Different typings because one of them uses a filter_map which means it has the Filter type, and another one has some unknown type that impl's Iterator

gui/src-tauri/src/main.rs
ImUrX: ``is_valid_path`` returns the path itself, we need to own it.

TheButlah: It is because you added another thing to this section of code. Its a simple fix, should be easy to do in under 3 minutes.

TheButlah: oh I understand now - the appimage path stuff was already added in a prior PR. In that case, resolving.

gui/src-tauri/src/main.rs
TheButlah: I would just make them collect into a vector then iterate on the vector but like...who really cares anyway. Resolving.

TheButlah (approved):

ImUrX:
this is ready for merge

TheButlah (approved):

PR #496, opened 2023-01-20
Some text was translated into Japanese.

tamakiti3010:


Louka3000:
"FK" can be replaced by whatever Japanese word is used for "Tracking"
Apart from that seems good, but there's still missing translation, do you wanna add them in this PR? (if not it's fine I'll still approve for the current changed lines! :p)
image
image
image
image
image

Also not sure if it's intentional, but "router" hasn't been translated (altho a lot of languages seem to keep it as is...)

tamakiti3010:
It was difficult to find a good translation for "Extended," so we left it as is.
FK", "SSID", and "Password" were left as they are because we thought they would look more natural in the UI if they were written in English.
Other than that, I fixed them!

gui/public/i18n/ja/translation.ftl
```diff
@@ -174,29 +174,29 @@ body_assignment_menu-unassign_tracker = トラッカーの割り当て解除
...
-tracker_selection_menu-HIP = { -tracker_selection-part } hip?
-tracker_selection_menu-LEFT_SHOULDER = { -tracker_selection-part } left shoulder?
-tracker_selection_menu-LEFT_UPPER_ARM = { -tracker_selection-part } left upper arm?
-tracker_selection_menu-LEFT_LOWER_ARM = { -tracker_selection-part } left lower arm?
-tracker_selection_menu-LEFT_HAND = { -tracker_selection-part } left hand?
-tracker_selection_menu-LEFT_UPPER_LEG = { -tracker_selection-part } left thigh?
-tracker_selection_menu-LEFT_LOWER_LEG = { -tracker_selection-part } left ankle?
-tracker_selection_menu-LEFT_FOOT = { -tracker_selection-part } left foot?
-tracker_selection_menu-LEFT_CONTROLLER = { -tracker_selection-part } left controller?
+-tracker_selection-part = どのトラッカーに割り当てるか？
```
ImUrX, line 177: Ok, let me explain this one.
``-tracker_selection-part`` is a term, we can't use it in the slimevr code. It's made so translators don't need to repeat themselves a lot of times for the same thing.

Currently your text looks like this: ``どのトラッカーに割り当てるか？ 右手?``
An alternative would be to do something like this:
```ftl
-tracker_selection-part = どのトラッカーを{ $body-part }に割り当てるか？
tracker_selection_menu-HEAD = { -tracker_selection-part(body-part: "頭") }
tracker_selection_menu-RIGHT_HAND = { -tracker_selection-part(body-part: "右手") }
```
and it would look like this:
``どのトラッカーを頭に割り当てるか？``
``どのトラッカーを右手に割り当てるか？``

You don't need to do it this way or use the same text I just gave you. Just giving you an example.
You could also remove ``-tracker_selection-part`` and do ``tracker_selection_menu-HEAD=どのトラッカーを頭に割り当てるか？``

ImUrX (changes requested):

ImUrX (approved):

Louka3000:
Please at least translate “FK” to “Tracking”, that’s how it is in the english translation now

Louka3000 (changes requested):
Please at least translate “FK” to “Tracking”, that’s how it is in the english translation now
Also same for “SSID” which got renamed to “Wi-Fi name”

Louka3000 (approved):

Louka3000 (approved):

Louka3000 (changes requested):
Oops, you forgot `onboarding-home` and `onboarding-home-description`, can you do that 😅

Louka3000 (approved):

ImUrX (approved):

PR #504, opened 2023-01-24
Fix switching between HMD and controller arm tracking

Louka3000:
Fix switching between HMD and controller arm tracking and improve some math and clear up some code in HumanSkeleton
Also, make the arms 0.26cm+0.26cm instead of 0.25cm+0.25cm


server/src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
```diff
@@ -1210,19 +1211,20 @@ protected void updateLocalTransforms() {
 	 * @param ankle the second Quaternion
 	 * @return the rotated Quaternion
 	 */
-	Quaternion extendedKneeYawRoll(Quaternion knee, Quaternion ankle) {
+	private Quaternion extendedKneeYawRoll(Quaternion knee, Quaternion ankle) {
+		knee = knee.clone();
```
ButterscotchV, line 1215: I myself was confused why this is being cloned, based off that, I'd like to have a comment so nobody removes it without understanding why it's done
```suggestion
		// Clone the knee as we're returning a separate quaternion
		knee = knee.clone();
```

server/src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
```diff
@@ -1234,7 +1236,15 @@ Quaternion extendedKneeYawRoll(Quaternion knee, Quaternion ankle) {
...
 	 */
-	Quaternion extendedPelvisYawRoll(Quaternion leftKnee, Quaternion rightKnee, Quaternion hip) {
+	private Quaternion extendedPelvisYawRoll(
+		Quaternion leftKnee,
+		Quaternion rightKnee,
+		Quaternion hip
+	) {
+		leftKnee = leftKnee.clone();
+		rightKnee = rightKnee.clone();
+		hip = hip.clone();
```
ButterscotchV, line 1246: Sorta same as last comment, but it shouldn't be necessary to clone all three, only hip since it's the only one being output.

ButterscotchV (changes requested):
Good ^w^
Look into these and I'll approve

server/src/main/java/dev/slimevr/vr/processor/skeleton/HumanSkeleton.java
Louka3000: I'm modifying the other twos at line 1254 and 1257

Eirenliel (approved):

ButterscotchV (approved):
I'm happier with less unecessary duplication, but I still think `rotBuf3` and `rotBuf4` could be used in place of `leftKnee` and `rightKnee` to reduce instantiation. If Eiren thinks this is fine then I'm okay with it, but I want it to at least be made aware.

Louka3000:
I'd rather not, as this is how the method is called
image

ButterscotchV:
> I'd rather not, as this is how the method is called
> image
> 

Wtff... Well then that's fine lmao, not your problem

PR #507, opened 2023-01-24
Korean: Improved untranslated parts

Kamilake:
Improved untranslated parts and corrected typos!


gui/public/i18n/ko/translation.ftl
```diff
@@ -397,10 +396,10 @@ onboarding-wifi_creds-skip = Wi-Fi 설정 건너뛰기
 onboarding-wifi_creds-submit = 저장!
 onboarding-wifi_creds-ssid =
     .label = SSID
```
ImUrX, line 398: Can you change SSID to Wifi name?

ImUrX (changes requested):

gui/public/i18n/ko/translation.ftl
Kamilake: Not a big deal, but would it be easier for users to understand?

ImUrX: We are forcing the change so we don't use a technical name like SSID

ImUrX (approved):

Louka3000 (approved):

Eirenliel (approved):

PR #517, opened 2023-01-25
Added package io.github.axisangles.ktmath

AxisAngles:
Added new kotlin math library, io.github.axisangles.ktmath, to replace jME3.math
includes Quaternion, Matrix3, Vector3, and EulerAngles.

- Contains mathematically propper operator overloads for matrices, vectors and quaternions.
- Has a small focus on numerical stability, trying to guarantee at least 6 sig figs on most operations.
- Contains several improvements over standard algorithms for converting to eulerangles.
- Provides 6 orders of eulerangles.
- Provides for homomorphic conversion from quaternion to matrix.
- Provides for matrix orthonormalization.


Eirenliel:
Why not include as a lib or a fork?

TheDevMinerTV:
(through gradle)

ImUrX:
We discussed it and it's worth keeping it in here XD. Gonna start reading :P

server/src/main/java/io/github/axisangles/ktmath/Vector3.kt
```diff
@@ -0,0 +1,92 @@
+@file:Suppress("unused")
+
+package io.github.axisangles.ktmath
+
+import kotlin.math.atan2
+import kotlin.math.sqrt
+
+data class Vector3(val x: Float, val y: Float, val z: Float) {
```
ButterscotchV, line 8: No "distance" method? It's pretty standard for most implementations to have one.

ButterscotchV (approved):
Well done on the unit tests, I'm really happy with how this looks.

server/src/main/java/io/github/axisangles/ktmath/Vector3.kt
AxisAngles: There might be some methods missing. I felt (b - a).len() was sufficient but maybe not. Same goes for some methods in the matrix library like angleTo, and angle. if you think of any features you want to see, let me know and I'll add them shortly.

ButterscotchV: `(b - a).len()` is definitely sufficient, I just figure shortcuts like `.distance(Vector3 other)` are good to have as it'll be used really frequently

ImUrX: I personally coming from a math side of things prefer ``(b - a).len()`` for representing distance. They are as short and they are more explicit.

AxisAngles: I think it's probably worth it though I suppose someone or I can always add these additions in the future.

Eirenliel (approved):
Besides the fact that unit tests are not done properly (usually you need to check all edge cases) :D

PR #525, opened 2023-01-28
zh-Hant translation

yenmax:
zh-Hant translation
Use zh-Hans as a template, special thanks to zh-Hans authors


ImUrX:
Translations uploaded to platform

PR #534, opened 2023-01-30
Remove the mounting reset button on the homepage

ButterscotchV:
Use the mounting calibration tab!!

### DIFF
--- a/gui/src/components/BVHButton.tsx
+++ b/gui/src/components/BVHButton.tsx
@@ -9,7 +9,7 @@ import { useWebsocketAPI } from '../hooks/websocket-api';
 import { BigButton } from './commons/BigButton';
 import { RecordIcon } from './commons/icon/RecordIcon';
 
-export function BVHButton() {
+export function BVHButton(props: React.HTMLAttributes<HTMLButtonElement>) {
   const { l10n } = useLocalization();
   const { useRPCPacket, sendRPCPacket } = useWebsocketAPI();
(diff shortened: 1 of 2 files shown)


ZRock35:
OSC users will either need to go through the mounting calibration tab or use tap to mount reset only now?

ButterscotchV:
> OSC users will either need to go through the mounting calibration tab or use tap to mount reset only now?

Yes, everyone should use the mounting calibration tab. idk how I feel about tap to mount reset, but it's the only way so I won't complain. If there's any problems with removing this, let me know.

Louka3000:
Why remove it? It’s convenient and fast

ButterscotchV:
> Why remove it? It’s convenient and fast

Because people just randomly press it expecting it to do magic, plus it's no quicker than the tab for it because you're supposed to do those steps anyways (and you only do it once per session!)

Louka3000:
I’d like to keep it for dev mode then at least, because it’d be a pain to do it via the tab each time I rebuild the server .-.

ImUrX:
Add a warning on non dev mode maybe?

Louka3000 (approved):

Rosdayle:
Just a comment on the steps on mounting calibration it doesn't include a notice on keeping your heels in place. I would be shifting around to go from standing into a squat/skii pose without knowing to keep my heels in place

ButterscotchV:
You don't necessarily need to keep your heels in the same position, but maybe it could be recommended

PR #541, opened 2023-02-01
Make mounting reset for arms forward

Louka3000:


Eirenliel:
That's a really hard position to get in 😢

Louka3000:
Frick you’re right… I was testing with arms-only. Doing the rest of the movement kinda breaks the balance ;-;

Louka3000:
Marking as draft. I think a toggle for this would be nice, on the mounting tab or in settings.

Louka3000:
Scratching this PR. I instead want a toggle for T-Pose (vrchat) and another for arms forward (v-tubing). To elaborate, it's not immediately here on my todo list

PR #543, opened 2023-02-02
Add a way to override current locale

ImUrX:
Current behavior is simple, if [config dir](https://docs.rs/tauri/1.2.4/tauri/api/path/fn.config_dir.html) contains a file called ``override.ftl``, it will be used instead of the selected locale in the language menu. It's loaded as the current type of locale which means number formatting, plural rules and etc won't work correctly...

The main usage for this is for testing locales without having to compile the GUI.


Louka3000 (approved):
works :+1:

PR #549, opened 2023-02-04
Twemoji emojis

ImUrX:
Add emoji font from twemoji so we can use all of them!


loucass003 (approved):

ImUrX:
works on chromium but not webkit

Eirenliel (approved):

ImUrX:
Should work for now, webkitgtk cant be fixed. But I tell it to use native emoji font if nothing works :p

ButterscotchV (approved):
Looks fine to me, not really too sure about it, so take my review with a grain of salt

Issue #515, opened 2023-01-25
Server fails to bind to address when the gui is still running


TheButlah:
The websocket address previously used seems to fail to bind to, when the gui is still running. This happens because by default, the OS will reject any attempts to re-bind to the socket until a somewhat long timeout. This behaviour can be disabled by the server when it creates the websocket, I believe.

Here is the stack trace:
```
00:57:13 [INFO] Running version 0043c30-dirty
00:57:13 [INFO] Starting the tracker server on port 6969...
00:57:18 [INFO] [VRCOSCHandler] Listening to port 9001
00:57:18 [INFO] [VRCOSCHandler] Sending to port 9000 at address /192.168.1.50
00:57:18 [INFO] [Keybinding] Currently only supported on Windows. Keybindings will be disabled.
00:57:18 [SEVERE] log4j:WARN No appenders could be found for logger (org.java_websocket.server.WebSocketServer).
00:57:18 [SEVERE] log4j:WARN Please initialize the log4j system properly.
00:57:18 [SEVERE] log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
00:57:18 [SEVERE] [WebSocket] Exception on connection null
java.net.BindException: Address already in use
        at java.base/sun.nio.ch.Net.bind0(Native Method)
        at java.base/sun.nio.ch.Net.bind(Net.java:555)
        at java.base/sun.nio.ch.ServerSocketChannelImpl.netBind(ServerSocketChannelImpl.java:337)
        at java.base/sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:294)
        at java.base/sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:89)
        at org.java_websocket.server.WebSocketServer.doSetupSelectorAndServerThread(WebSocketServer.java:567)
        at org.java_websocket.server.WebSocketServer.run(WebSocketServer.java:367)
        at java.base/java.lang.Thread.run(Thread.java:833)
00:57:19 [INFO] [TrackerServer] Handshake received from /192.168.51.34:6969
00:57:19 [INFO] [TrackerServer] Tracker 0 added with address /192.168.51.34:6969. Board type: 10, imu type: 4, firmwar
e: 0.3.2 (15), mac: 34:85:18:03:83:C0, name: udp://34:85:18:03:83:C0
[...]

PR #554, opened 2023-02-05
fix: websocket fail to bind to address due to time_wait

marcozzxx810:
Related to: #515 

setReuseAddr(true) is the Setter for soReuseAddr Enable/disable SO_REUSEADDR for the socket

SO_REUSEADDR allows the server to bind to an address which is in a TIME_WAIT state

https://www.javadoc.io/doc/org.java-websocket/Java-WebSocket/1.3.5/org/java_websocket/server/WebSocketServer.html


TheButlah:
Great, glad someone tackled this :)

TheButlah (approved):

ButterscotchV (approved):
It seems implied that this option can forcibly bind to an occupied socket. I'm unsure of the ideal behaviour for the websocket, but this should in theory resolve the linked issue.

Eirenliel (approved):

Issue #497, opened 2023-01-21
Fix server logging


Eirenliel:
We don't need to keep logs indefinetely. Should just remove full log and daily log and keep only last log.

Eirenliel:
Fixed in #555.

PR #555, opened 2023-02-05
refactor: remove full log and daily log

marcozzxx810:
related to #497 

- [x] remove the daily log (logs in logs folder)
- [x] remove full log (log_main.log)
- [x] remove the logs variable in LoggerManager as daily logs are not needed any more 

Please let me know if there is any other improvement

### DIFF
--- a/server/src/main/java/dev/slimevr/Main.kt
+++ b/server/src/main/java/dev/slimevr/Main.kt
@@ -49,7 +49,7 @@ fun main(args: Array<String>) {
 	}
 	val dir = File("").absoluteFile
 	try {
-		LogManager.initialize(File(dir, "logs/"), dir)
+		LogManager.initialize(dir)
 	} catch (e1: java.lang.Exception) {
 		e1.printStackTrace()
 	}
--- a/server/src/main/java/io/eiren/util/logging/LogManager.java
+++ b/server/src/main/java/io/eiren/util/logging/LogManager.java
@@ -3,7 +3,6 @@ package io.eiren.util.logging;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
-import java.text.SimpleDateFormat;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.ConsoleHandler;
 import java.util.logging.FileHandler;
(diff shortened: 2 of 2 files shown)


ButterscotchV (approved):
Looks good. If you're looking to add an additional feature, it would be quite nice to have a file size limit for the log file. Something like 50 MB or so? Anything bigger is just excessive and would waste space on people's computers.

Note: We could probably delete these old log files with the SlimeVR installer, maybe we should consider that?

marcozzxx810:
@ButterscotchV It is my pleasure to add this feature, but I may need more clarification and this feature should be regarded as another PR.

What is the expected behavior when it exceeds 50MB?

Proposal:
1. setup two logs (log_last_0.log, log_last_1.log) with 25MB limit
2. rotate the log when the output approximately is equal 25MB size 

Ref:
```
public FileHandler(pattern,
           int limit,
           int count)
     
Initialize a FileHandler to write to a set of files. When (approximately) the given limit has been written to one file, another file will be opened. The output will cycle through a set of count files.
The FileHandler is configured based on LogManager properties (or their default values) except that the given pattern argument is used as the filename pattern, the file limit is set to the limit argument, and the file count is set to the given count argument.

The count must be at least 1.
```
https://docs.oracle.com/javase/7/docs/api/java/util/logging/FileHandler.html#FileHandler(java.lang.String,%20int,%20int)

ButterscotchV:
I was thinking of just cutting off the logging at that point. I really don't think it should be expected to get that large at all, so if it reaches that size, then something really bad has happened and will likely continue to write to it constantly. I'd like to hear Eiren's thoughts on this though. Your proposal does seem like a viable solution if we value the latest logs rather than valuing the earliest logs.

marcozzxx810:
For the removal of the old logs, I guess they will be removed already 

https://github.com/SlimeVR/SlimeVR-Installer/blob/main/windows/web/slimevr_web_installer.nsi#L602

Eirenliel (approved):

Eirenliel:
> Looks good. If you're looking to add an additional feature, it would be quite nice to have a file size limit for the log file. Something like 50 MB or so? Anything bigger is just excessive and would waste space on people's computers.
> 
> Note: We could probably delete these old log files with the SlimeVR installer, maybe we should consider that?

Plz move to the #497 or another issue to discuss features out of the PRs scope 😛

PR #556, opened 2023-02-05
docs: include java version

marcozzxx810:
When I follow the CONTRIBUTING.md, I encounter a build error because of the old version of JDK

An error like below. 

```bash
Unrecognized option: --add-exports
```

Therefore, I suggest adding a Java version inside  CONTRIBUTING.md to make sure the guideline works as expected

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,15 @@ cd SlimeVR-Server
 
 Now you can open the codebase in your favorite IDE or text editor.
 
+### Installing Java
+The codebase is required to build with Java version 11 or higher.
+
+```bash
+# Check java version
+java --version
+```
+
+
 ### Building the code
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,6 +17,15 @@ cd SlimeVR-Server
 
 Now you can open the codebase in your favorite IDE or text editor.
 
+### Installing Java
+The codebase is required to build with Java version 11 or higher.
```
Eirenliel, line 21:
```suggestion
The codebase is required to build with Java version 17 or higher.
```

Eirenliel: We're targeting Java 17 already.

Eirenliel (changes requested):

Eirenliel (approved):

Issue #558, opened 2023-02-05
Add log file size limiter


marcozzxx810:
By @ButterscotchV from https://github.com/SlimeVR/SlimeVR-Server/pull/555., it would be quite nice to have a file size limit for the log file. Something like 50 MB or so? Anything bigger is just excessive and would waste space on people's computers.

My thoughts:
1. setup two logs (log_last_0.log, log_last_1.log) with 25MB limit
2. rotate the log when the output approximately is equal 25MB size 

Ref:
```
public FileHandler(pattern,
           int limit,
           int count)
     
Initialize a FileHandler to write to a set of files. When (approximately) the given limit has been written to one file, another file will be opened. The output will cycle through a set of count files.
The FileHandler is configured based on LogManager properties (or their default values) except that the given pattern argument is used as the filename pattern, the file limit is set to the limit argument, and the file count is set to the given count argument.

The count must be at least 1.
```
https://docs.oracle.com/javase/7/docs/api/java/util/logging/FileHandler.html#FileHandler(java.lang.String,%20int,%20int)

marcozzxx810:
@Eirenliel, any opinion about this?

Eirenliel:
sounds good

marcozzxx810:
will issue a PR for that, thanks @Eirenliel 😄

Eirenliel:
Fixed in #560.

PR #560, opened 2023-02-06
feat: log file size limiter

marcozzxx810:
related to #558

### DIFF
--- a/server/src/main/java/io/eiren/util/logging/LogManager.java
+++ b/server/src/main/java/io/eiren/util/logging/LogManager.java
@@ -27,10 +27,11 @@ public class LogManager {
 		if (mainLogDir != null) {
 			if (!mainLogDir.exists())
 				mainLogDir.mkdirs();
-			File lastLogFile = new File(mainLogDir, "log_last.log");
-			if (lastLogFile.exists())
-				lastLogFile.delete();
-			FileHandler filehandler = new FileHandler(lastLogFile.getPath(), true);
+			for (File f : mainLogDir.listFiles()) {
+				if (f.getName().startsWith("log_last"))
+					f.delete();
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/eiren/util/logging/LogManager.java
```diff
@@ -27,10 +27,11 @@ public static void initialize(File mainLogDir)
...
 				mainLogDir.mkdirs();
-			File lastLogFile = new File(mainLogDir, "log_last.log");
-			if (lastLogFile.exists())
-				lastLogFile.delete();
-			FileHandler filehandler = new FileHandler(lastLogFile.getPath(), true);
+			for (File f : mainLogDir.listFiles()) {
+				if (f.getName().startsWith("log_last"))
+					f.delete();
+			}
+			FileHandler filehandler = new FileHandler("log_last_%g.log", 25 * 1000000, 2);
```
ButterscotchV, line 34: `mainLogDir` should not be ignored for this file path. (This suggestion would also require an additional import for `import java.nio.file.Path;`)
```suggestion
			String lastLogPattern = Path.of(mainLogDir.getPath(), "log_last_%g.log").toString();
			FileHandler filehandler = new FileHandler(lastLogPattern, 25 * 1000000, 2);
```

ButterscotchV (changes requested):

server/src/main/java/io/eiren/util/logging/LogManager.java
marcozzxx810: added the mainLogDir path!

ButterscotchV (approved):

Eirenliel (approved):

PR #566, opened 2023-02-09
Support VRTrackers for protocol rotations

Louka3000:
- Use getRawRotation for all trackers since they all support it.
- Use VRTracker getRotation for getRotationReferenceAdjusted (that's adjusted to the HMD)
- use VRTracker getRawRotation for addRotationIdentityAdjusted (that's normally adjusted to the identity Quaternion, which I believe is already how we get rotations from OpenVR)


Louka3000:
> sending them also for playspace-aligned trackers and HMD, which come through the feeder app, feels wrong.

VRTrackers are affected by resets now, so they should have their adjusted rotations sent just like IMU trackers.

0forks (approved):

Eirenliel (approved):

PR #573, opened 2023-02-09
Add macOS builds

ImUrX:
Added JVM detection and macOS `.app` artifacts!

### DIFF
--- a/.github/workflows/gradle.yaml
+++ b/.github/workflows/gradle.yaml
@@ -130,3 +130,40 @@ jobs:
         with:
           name: SlimeVR-GUI-AppImage
           path: target/release/bundle/appimage/slimevr*.AppImage
+
+  bundle:
+    runs-on: macos-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: "SlimeVR-Server"
+          path: server/build/libs/
+
+      - name: Cache cargo dependencies
+        uses: Swatinem/rust-cache@v2
+
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.node-version'
+          cache: 'npm'
+
+      - name: Build
+        run: |
+          npm ci
+          npm run tauri build -- "--target universal-apple-darwin"
+
+      - name: Modify Application
+        run: |
+          cd target/release/bundle/macos/slimevr.app/Contents/MacOS
+          cp $( git rev-parse --show-toplevel )/server/build/libs/slimevr.jar ./
+
+      - uses: actions/upload-artifact@v3.1.0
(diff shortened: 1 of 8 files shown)


Eirenliel (approved):

ImUrX:
it works now, ``.dmg`` file should be published. It didn't let me run the `.app` one.

.github/workflows/gradle.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
       - name: Cache cargo dependencies
         uses: Swatinem/rust-cache@v2
         with:
-          shared-key: "tauri"
+          shared-key: "ubuntu-20.04"
```
TheButlah, line 96: Pretty sure the cache already checks for OS

TheButlah: https://github.com/Swatinem/rust-cache#cache-details

ImUrX: > the github [job_id](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id),

It also does it per job_id, caches aren’t shared between different actions

Louka3000 (approved):

.github/workflows/gradle.yaml
TheButlah: I don't think its a good idea to share the cache between different actions - to me it seems like its just asking for cache thrashing. Also, there will be a race condition on which cache gets uploaded, since the jobs run in parallel

ImUrX: It's just sharing cache so both don't have to maintain different caches, they are not different systems (promised by Github).
Cache locking is something that already exists, only one can upload on it if both actions are being ran concurrently, although I'm sure the build GUI one will finish before the bundling one starts. The main idea is to save time and space.

TheButlah: so if I understand you correctly:
* Github jobs run on the same machine, always. This means we are not sharing a cache between different runners that could thrash?
* These jobs share the same cache now, so we should see improvements

I didn't think github jobs worked that way - I thought jobs will run on any runner if its available, so they aren't necessarily running on the same machine.

If they did run on the same machine, you're right the lockfile in cargo will prevent concurrent modification. I wasn't really talking about that though -I was saying that one job will overwrite the cache of the other (thats if they *dont* run on the same machine though, if they run on the same the lockfile makes it not concurrent).

I think the only optimization I really see here is the space optimization not time. Does that make sense, or am I overlooking something?

If the only benefit is space optimization, I'm still not a big fan because I'm wary of cache thrashing (for example, running clippy before cargo build [thrashes cache](https://www.reillywood.com/blog/rust-faster-ci/). But maybe its not an issue?

TheButlah: I'm gonna stamp it and we can deal with it later

TheButlah (approved):

PR #575, opened 2023-02-10
Fix trackers overflow in dev mode

Louka3000:


gui/src/components/home/Home.tsx
```diff
@@ -44,7 +44,7 @@ export function Home() {
         </div>
       )}
       {config?.debug && trackers.length > 0 && (
-        <div className="flex px-2 pt-5 justify-center  overflow-x-auto">
+        <div className="px-2 pt-5 justify-center overflow-y-scroll overflow-x-auto">
```
ImUrX, line 47:
```suggestion
        <div className="px-2 pt-5 overflow-y-scroll overflow-x-auto">
```
Seeing the child element, it's just another div that is also a flex box.

ImUrX (changes requested):

ImUrX (approved):

ButterscotchV (approved):

Issue #422, opened 2022-12-23
No feedback when triggering tap gestures.


nullstalgia:
Howdy howdy.

When using the new and (wonderful) tap reset, it's hard to tell if my taps actually went through and if I should stay in the pose for the duration of time I set for the delay, or if I should tap again, or when I can put my arms down if I have arm trackers.

A beep of some kind when the gesture is recognized, and another when the reset has occurred would be ideal.

Eirenliel:
The feature that plays a sound on reset was requested long ago, we even have a meow sound recorded for this. If someone goes to implement it, please ask me for the audio file!

marcozzxx810:
I will take this!

Louka3000:
Closed by https://github.com/SlimeVR/SlimeVR-Server/pull/576

PR #576, opened 2023-02-11
feat: feedback sound when reset triggered

marcozzxx810:
https://github.com/SlimeVR/SlimeVR-Server/issues/422

Borrow some idea from [Philip6294](https://github.com/Philip6294)

- [x] add gui setting
- [x] configure i18n 
- [x] enable the play sound in server


Philip6294:
First, great someone is actually fully implementing this because I always wanted this feature and I didn't know how to modify the GUI. 
One thing I would add tho is the option to have the sound play when the tap is detected and not when the reset is executed (so before the reset delay). 
My version played the sound when the tap was detected.

ButterscotchV:
Could also be nice to have this for the reset shortcuts! Through the keybinds and controller bindings

marcozzxx810:
> First, great someone is actually fully implementing this because I always wanted this feature and I didn't know how to modify the GUI. One thing I would add tho is the option to have the sound play when the tap is detected and not when the reset is executed (so before the reset delay). My version played the sound when the tap was detected.

I have taken a quick look at your version, and now should work as expected.

Do you guys know how can I trigger tap with owoTracker? seem the tap based is not working for all reset now

marcozzxx810:
For the controller binding, do I need to do anything inside the codebase? ( seem we need to do sth in feeder-app )

server/src/main/java/dev/slimevr/Keybinding.java
```diff
@@ -69,6 +75,16 @@ public void onHotKey(int identifier) {
 				LogManager.info("[Keybinding] Reset mounting pressed");
 				server.scheduleResetTrackersMounting(this.config.getResetMountingDelay());
 			}
+			case FEEDBACK_SOUND -> {
```
Eirenliel, line 78: Wait what is going on here, why is there a keybind to play the sound?

Eirenliel (changes requested):
The sound should play when actions are triggered no matter the source - tap, keybind, controller, or GUI button. It should be played before the action and and after.

Philip6294:
@Eirenliel 
> The sound should play when actions are triggered no matter the source - tap, keybind, controller, or GUI button. It should be played before the action and and after.

I think playing the sound before the action is only needed when the source is a tap or maybe GUI Button. Because AFAIK those are the only cases where there's a delay between reset trigger and actual reset.

Eirenliel:
True, @Philip6294.

server/src/main/java/dev/slimevr/Keybinding.java
marcozzxx810: This is the keybind for enable or disable the feedback sound for tap only

marcozzxx810:
@Philip6294 @Eirenliel, I would like to clarify do we need to have GUI button feedback sound?

Since GUI button will already start the countdown, which is a visual feedback.

For tapping, it is better with a sound feedback

Philip6294:
I would only do sound Feedback for Tapping Resets. 
Other Reset could maybe have Feedback when the reset actually happens, but GUI probably does not need that because it already shows a countdown as feedback.

server/src/main/java/dev/slimevr/Keybinding.java
Eirenliel: That's a weird feature 😅 It should be done via settings, not keybind.

Eirenliel:
I think feedback when reset happens should be from any source. It's UX consistency.

Philip6294:
Jup, for UX consistency it's definetly better when it happens on every reset

server/src/main/java/dev/slimevr/Keybinding.java
marcozzxx810: We have that toggle with setting, make sense to me for removing the toggle in keybind

marcozzxx810:
Thanks @Philip6294 @Eirenliel 

I think these is the changes needed:

- [ ] Make feedback sound for all reset action regardless of the source 
- [ ] remove the keybind
- [ ] New config for feedback sound

For the feedback config, do we need to have the sound toggle for each action type( GUI, Tab and keybind)

marcozzxx810:
Final Decision with @Eirenliel:

- [ ] implement the feedback sound for tap first, when the tap reset is triggered, play a sound
- [ ] put this setting in interface

will create another PR for on reset sound

Louka3000:
Tap sound toggle should be below tap settings in gesture settings imo

loucass003:
@marcozzxx810 Please your PR so the gui play the sound. (or any sound for that matter). Add a RPCMessage to the protocol to so. The sound should not be handled by the server imo

TheButlah:
Hi, just saw this because I was requested for review in solarxr. In my opinion, it is violating separation of concerns for the server to control UI such as sounds.

@loucass003 I believe it makes more sense for the server to notify the GUI of concepts like "reset complete" or "reset triggered" rather than dictate UX like "go play a sound". The GUI should be in charge of UX not the server.

Do you agree? What are your thoughts.

loucass003:
Yes i do agree it would be better. Also that means that this PR that was merged recently is probably not necessary anymore. if there is no concept of "Sound" at all in the server. https://github.com/SlimeVR/SolarXR-Protocol/pull/97/files

marcozzxx810:
> Hi, just saw this because I was requested for review in solarxr. In my opinion, it is violating separation of concerns for the server to control UI such as sounds.
> 
> @loucass003 I believe it makes more sense for the server to notify the GUI of concepts like "reset complete" or "reset triggered" rather than dictate UX like "go play a sound". The GUI should be in charge of UX not the server.
> 
> Do you agree? What are your thoughts.

Make sense to me, coz we don't want sound event couple with server. I guess the following changes is needed:

- revert the [TabDection](https://github.com/SlimeVR/SolarXR-Protocol/pull/97/files)
- add a ResetResponse to the RPC message in the protocol
- add an RPCResetHandler, ResetHandler, ResetListener
- emit reset rpc message when triggered or completed

Eirenliel:
This PR started so simple and change so much... but I agree, we can go for good structure!

marcozzxx810:
> > Hi, just saw this because I was requested for review in solarxr. In my opinion, it is violating separation of concerns for the server to control UI such as sounds.
> > @loucass003 I believe it makes more sense for the server to notify the GUI of concepts like "reset complete" or "reset triggered" rather than dictate UX like "go play a sound". The GUI should be in charge of UX not the server.
> > Do you agree? What are your thoughts.
> 
> Make sense to me, coz we don't want sound event couple with server. I guess the following changes is needed:
> 
> * revert the [TabDection](https://github.com/SlimeVR/SolarXR-Protocol/pull/97/files)
> * add a ResetResponse to the RPC message in the protocol
> * add an RPCResetHandler, ResetHandler, ResetListener
> * emit reset rpc message when triggered or completed

@loucass003 , @TheButlah , can I get some advice for this to-do?

TheButlah:
I lack the context to really help, I just chimed in because the current approach seemed like an antipattern. 

@loucass003 said he would provide some advice once he gets home, :nya_gun: him if he doesn't.

gui/src/sounds/sounds.tsx
```diff
@@ -0,0 +1,18 @@
...
+    case ResetType.Full: {
+      new Audio('/sounds/double_beep.wav').play();
+      break;
+    }
+    case ResetType.Mounting: {
+      new Audio('/sounds/triple_beep.wav').play();
+      break;
+    }
+  }
+}
```
ImUrX, line 18: Could you put the audio elements in a const outside the function like
```js
const singleBeep = new Audio('/sounds/single_beep.wav')
// ...
singleBeep.play()
```

Audio elements shouldn't be played the moment they are created. This is mostly for browsers tho, in where there is an actual latency of buffering the audio.

ImUrX (changes requested):
You should make this a draft, if it isn't finished yet :p.

gui/src/sounds/sounds.tsx
ImUrX: Do it anyways though, because we do have a browser version that people can use!

marcozzxx810: > You should make this a draft, if it isn't finished yet :p.

I think I completed the PR ytd and it is ready for review.

Sorry for the notification.

marcozzxx810: fixed!

marcozzxx810:
need:
1. https://github.com/SlimeVR/SolarXR-Protocol/pull/101
2. https://github.com/SlimeVR/SlimeVR-Server/pull/611 ( this need to merge first as protocol pr has changed)

gui/src/hooks/config.ts
```diff
@@ -29,6 +30,7 @@ const initialConfig = {
   doneOnboarding: false,
   watchNewDevices: true,
   lang: 'en',
+  feedbackSound: true,
```
Louka3000, line 33:
```suggestion
  feedbackSound: false,
```
I personally would have any sound setting default to false. What do you think @loucass003 ?

Louka3000 (changes requested):
Resets only happen on GUI buttons... I can't get it to trigger on taps nor keybinds.
Also it only triggers before the reset when it does, not before and after...

gui/src/hooks/config.ts
Eirenliel: I would not.

Louka3000: Why?

marcozzxx810:
> Resets only happen on GUI buttons... I can't get it to trigger on taps nor keybinds. Also it only triggers before the reset when it does, not before and after...

I noticed there is some weird behavior but I find out when I launch GUI first then server, then toggle on and off the feedback setting in the setting. The key bind will work. 


I am not too sure about the implementation here:  https://github.com/SlimeVR/SlimeVR-Server/pull/576/files#diff-2da1b5d9816c92ddc68fdcb4f223336f0adcf2816a006f090ff21c82b9f7c45aR36

So there are two issue:
1. ResetResponse is not received when server start first then client
2. the INIT GUI config doesn't works? how do I update this file https://github.com/SlimeVR/SlimeVR-Server/blob/main/gui/src/hooks/config.ts#L66

marcozzxx810:
Resolved the https://github.com/SlimeVR/SlimeVR-Server/issues/621

Not the ws.current. Some other code can mutate the ws.current and the old websocket instance will not be closed by cleanup. The connection will remain open and cause bugs

Therefore we should use another variable wsCurrent to call close()

marcozzxx810:
> Resets only happen on GUI buttons... I can't get it to trigger on taps nor keybinds. Also it only triggers before the reset when it does, not before and after...

I don't think we need the sound after yet.

Eirenliel:
did this issue blow up out of proportion? what's going on?

marcozzxx810:
> did this issue blow up out of proportion? what's going on?

The first issue is related to https://github.com/SlimeVR/SlimeVR-Server/issues/621 , then the cleanup function is not properly setup resulting bug in receiving the rpc message

The second one is that the GUI init config is stored in %appdata%, just we need to override the config. I overthink

Eirenliel:
> > Resets only happen on GUI buttons... I can't get it to trigger on taps nor keybinds. Also it only triggers before the reset when it does, not before and after...
> 
> I don't think we need the sound after yet, since this PR started with the initiative to play a sound when triggered.
> 
> With both started and ended, it will be quite noisy

Yes. This PR is for triggering on the start. We can add after later, since it requires modifying different parts of the code, and not as important. Though I would like to have an option in the future.

gui/src/hooks/config.ts
Eirenliel: Because it's an important UX feedback.

marcozzxx810:
The only remaining issue left is linux cannot play the sound automatically by useEffect hook.

marcozzxx810:
> The feedback sound being on by default doesn't work. I have to toggle them off and on again (reproduced by deleting the key in localstorage). Also is there a less aggressive beep please. The pitch and popping of this one hurts me ;-; The sound itself works tho, with keybind, GUI and taps at least :3

@Louka3000, thank you for the review. I have successfully reproduced the issue by removing all keys. 

Now, if the local storage is completely empty, the feedbacksound setting should be true and able to play sound

For the sound file, @Eirenliel will provide some meow sound to replace the beep ( I hate too )

marcozzxx810:
@Louka3000 fixed, thank you for the help

Louka3000 (approved):
LGTM

marcozzxx810:
@Louka3000 updated!

Eirenliel (approved):

0forks (commented):
Which bug did this fix? React hooks are called multiple times, so you can't just use a simple local variable as shared state here, it leads to unexpected behavior. This commit needs to be reverted and the issue should be fixed in another way.

Currently this breaks the `reconnect()` call from the fast data feed toggle in developer mode.

marcozzxx810:
@0forks  https://github.com/SlimeVR/SlimeVR-Server/issues/621 

Reverting the fix will resulting the sound features will break, Also, reverting will make the problem worst. As the old implementation will not disconnect properly coz the old way used a mutable state to store the connection. The mutable state will overwrite when it initializes so when calling was. close with the user will not close the connection.

As I know reconnecting only happens when development. maybe you can find an issue a PR to properly fix it instead of reverting the feature that works. 

The previous approach lead to unexpected behavior coz ref means to be mutable but local variable now will not overwriting easily. Maybe you can take a look : https://stackoverflow.com/questions/60152922/proper-way-of-using-react-hooks-websockets#comment120898768_60161181

0forks:
Will look into it, thanks for context.

PR #581, opened 2023-02-12
Convert AutoBone to Kotlin and extend PoseFrame data functionality

ButterscotchV:
- Converts all of AutoBone classes to Kotlin
- Adds PoseFrame support for:
  - Designation as an enum value rather than a string
  - Acceleration
  - Raw rotation
- Prints out a representation of every data flag when listing trackers in AutoBone
- Adds `add` and `remove` utilities to TrackerFrameData
- Automatically serialize designation as an enum rather than a string


ButterscotchV:
Tested lightly, seems to work really well

Louka3000 (approved):

ButterscotchV:
Have tested this more thoroughly now, looks to be working the same as before

Issue #544, opened 2023-02-03
GUI doesn't have a way to assign controller TrackerPositions


Louka3000:
Feeder app will assign controllers' tracker roles to controllers, but if user changes it, there's no way to change it back
bruh moment

PR #588, opened 2023-02-14
Use hands everywhere instead of controller

Louka3000:
Just like HMD doesn't exist and we use Head instead, we should use Hand instead of Controller, as we don't care about button input, and a non-controller tracker can be worn on the hand and have positional data, and thus should be used the same way as a controller inside SlimeVR.

~~Needs https://github.com/SlimeVR/SolarXR-Protocol/pull/99~~

Closes https://github.com/SlimeVR/SlimeVR-Server/issues/544


ButterscotchV (approved):
Looks fine

Eirenliel (approved):

PR #598, opened 2023-02-18
Start using localStorage for config

ImUrX:
This will let the config work on web and we can just stop having filesystem operations.

I added a migration script which is pretty simple also. The `config.json` localStorage code is based on @ButterscotchV's!

### DIFF
--- a/gui/src/hooks/config.ts
+++ b/gui/src/hooks/config.ts
@@ -1,10 +1,4 @@
-import {
-  BaseDirectory,
-  createDir,
-  readTextFile,
-  renameFile,
-  writeFile,
-} from '@tauri-apps/api/fs';
+import { BaseDirectory, readTextFile } from '@tauri-apps/api/fs';
 
 import { createContext, useContext, useRef, useState } from 'react';
 import { DeveloperModeWidgetForm } from '../components/widgets/DeveloperModeWidget';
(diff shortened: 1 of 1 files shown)


loucass003 (approved):

gui/src/hooks/config.ts
```diff
@@ -73,9 +60,21 @@ export function useConfigProvider(): ConfigContext {
...
+          }).catch(() => null);
+
+          if (oldConfig) localStorage.setItem('config.json', oldConfig);
+
+          localStorage.setItem('configMigrated', 'true');
+        }
+
+        const json = localStorage.getItem('config.json');
+
+        if (!json) throw new Error('Config has ceased existing for some reason');
```
ButterscotchV, line 76: This error is a little strange, as you'd get it before any config is written, so it's not necessarily happening when the config vanishes

ButterscotchV (approved):

gui/src/hooks/config.ts
ImUrX: it can also appear if you remove the config while its gonna get read :P. But it's an error that shouldn't happen at all unless user error or weird behaviour

ButterscotchV: eh it's fine whatever

PR #601, opened 2023-02-19
fix row height

loucass003:


ImUrX (changes requested):
fix the eslint problem, but seems fine

loucass003:
No ? as long as the git lint config does not match the local files gui config. i wont follow it. And i wont fix it in this pr this is not the purpose of it

gui/src/components/tracker/TrackersTable.tsx
```diff
@@ -1,21 +1,21 @@
+import { useLocalization } from '@fluent/react';
 import classNames from 'classnames';
 import { IPv4 } from 'ip-num/IPNumber';
-import { MouseEventHandler, ReactNode, useState, useMemo } from 'react';
+import { MouseEventHandler, ReactNode, useMemo, useState } from 'react';
 import {
   TrackerDataT,
   TrackerIdT,
-  TrackerStatus as TrackerStatusEnum,
+  TrackerStatus as TrackerStatusEnum
```
ImUrX, line 8:
```suggestion
  TrackerStatus as TrackerStatusEnum,
```

ImUrX (changes requested):

gui/src/components/tracker/TrackersTable.tsx
loucass003: no

ImUrX (approved):

ButterscotchV (approved):

PR #603, opened 2023-02-20
Fix hip interpolation

Louka3000:
*Maybe* make sure the squash commit isn't named "wtf java"

Uriel and I have literally zero idea what causes the bug. Nothing modifies left and right knee outside of the negateLocal... we even logged the values: nothing.
It's cursed
We can afford 2 object creations more tho :p


ButterscotchV (approved):
=w=

Issue #3, opened 2021-06-12
VMC support


Eirenliel:
Add support for Virtual Motion Capture trackers server for both reading and writing data to support motion capture and vtubing software.

Eirenliel:
![](https://github.trello.services/images/mini-trello-icon.png) [VMC support for VSeeFace](https://trello.com/c/7pHYCHIo/19-vmc-support-for-vseeface)

PR #606, opened 2023-02-23
VMC Support

Louka3000:
- Send shareable trackers and all UnityBones (that we have) through VMC.
- Receive VMC trackers and bones
- Read VRM

~~Needs: https://github.com/SlimeVR/SolarXR-Protocol/pull/102~~
Closes: https://github.com/SlimeVR/SlimeVR-Server/issues/3


server/src/main/java/dev/slimevr/tracking/trackers/ComputedTracker.java
```diff
@@ -135,10 +142,15 @@ public boolean userEditable() {
...
 	@Override
 	public void dataTick() {
+		timeAtLastUpdate = System.currentTimeMillis();
 	}
 
 	@Override
 	public void tick() {
+		if (useTimeout) {
+			if (System.currentTimeMillis() - timeAtLastUpdate > TIMEOUT_MS)
+				setStatus(TrackerStatus.DISCONNECTED);
```
Eirenliel, line 152: add the reverse of it? it'll always be disconnected

Eirenliel (changes requested):

server/src/main/java/dev/slimevr/tracking/trackers/ComputedTracker.java
Eirenliel: The tracker will never stop being disconnected and will be set to DISCONNECTED again on each tick.

Louka3000: timeAtLastUpdate gets updated to System.currentTimeMillis, so if it's receiving data it'll be for example:
19348 - 19348 > 3000
0 > 3000
false

If not, it'll be, for example:
29348 - 19348 > 3000
10000 > 3000
true

Eirenliel: Okay but where does it get out of the disconnected state?

Louka3000: Trackers received from VMC and feeder app are changed to OK on each data update, but I'll double check HMD later

Louka3000: Yes HMD sets it to OK... I remembered which part of my code did that and saw you reviewed it asking for a comment :p

Louka3000: Resolving this, will rewrite trackers to be better as my next big PR anyway...

Eirenliel (approved):

PR #607, opened 2023-02-24
server: Fix frame drops in SteamVR/Windows

0forks:
The named pipe thread currently has a 5ms sleep, which leads to inconsistent tracker position updates (drops about half of the frames). This PR improves message timing consistency for OpenVR trackers on Windows by waiting for a data ready event, and runs an update as soon as the HMD position frame is received from the driver.

Comparison before and after:

[frametimes.webm](https://user-images.githubusercontent.com/114709761/221206830-a52744c8-bd63-41aa-a93e-101bdb441e17.webm)


Eirenliel (approved):

0forks:
Thank you for pointing this out. This will be fixed by #629.

Issue #11, opened 2021-07-13
Implement feet touching floor


Eirenliel:
There can be an option added for feet always trying to touch floor when users raise legs, even if they don't have feet trackers. If the ankle is slightly above the floor, foot can snap down emulating user standing on their toes. This should be an opt-in option, since it can mess up some user's inentions.

Louka3000:
Implemented by https://github.com/SlimeVR/SlimeVR-Server/pull/609

PR #609, opened 2023-02-26
Toesnap and Footplant

Stermere:
This PR adds the feature's toe snap and foot plant, both components are focused on improving tracking without the presence of foot trackers. This addresses issue #11  and cannot be merged until https://github.com/SlimeVR/SolarXR-Protocol/pull/104 is merged.


server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
```diff
@@ -932,6 +1148,31 @@ private void updateLockStateCounters() {
...
 
+	// remove the x and z components of the given quaternion
+	private Quaternion isolateYaw(Quaternion quaternion) {
+		return new Quaternion(
+			0,
+			quaternion.getY(),
+			0,
+			quaternion.getW()
+		);
+	}
```
ButterscotchV, line 1159: Is this safe? I'm not quite sure that this is fully reliable, particulatly at a rotation of 180 degrees on the yaw axis? I need someone else's input on this or if you can explain it, I would appreciate that a lot.

server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
```diff
@@ -713,6 +757,178 @@ private void correctUnlockedRightFootTracker() {
...
+		skeleton.computedLeftFootTracker.rotation.set(leftFootRotation);
+		skeleton.computedRightFootTracker.rotation.set(rightFootRotation);
+	}
+
+	// returns the length of the xz components of the normalized difference
+	// between two vectors
+	public float getXZAmount(Vector3f vec1, Vector3f vec2) {
+		return vec1
+			.subtract(vec2)
+			.normalize()
```
ButterscotchV, line 888: `normalizeLocal()` can be used here instead, as `subtract()` is creating a new `Vector3f` object. I also see potential for using a buffer `Vector3f` as a method argument for this calculation, especially since it isn't ending up returning a `Vector3f` in the end.
```suggestion
			.normalizeLocal()
```

ButterscotchV (changes requested):

server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
Stermere: I'm not sure that I know what using a buffer  `Vector3f` would look like, I'd be happy to do that though if you could explain it

gui/public/i18n/en/translation.ftl
```diff
@@ -273,7 +273,7 @@ settings-general-tracker_mechanics-drift_compensation-max_resets-label = Use up
 ## FK/Tracking settings
 settings-general-fk_settings = Tracking settings
 settings-general-fk_settings-leg_tweak = Leg tweaks
-settings-general-fk_settings-leg_tweak-description = Floor-clip can Reduce or even eliminates clipping with the floor but may cause problems when on your knees. Skating-correction corrects for ice skating, but can decrease accuracy in certain movement patterns.
+settings-general-fk_settings-leg_tweak-description = Floor-clip can Reduce or even eliminates clipping with the floor but may cause problems when on your knees. Skating-correction corrects for ice skating, but can decrease accuracy in certain movement patterns. Toe snap and foot plant make a guess at the rotation of your foot and are only active when no trackers are assigned to the feet.
```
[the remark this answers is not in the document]
Stermere: I think that is a great idea!

server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
Stermere: I wish I could tell you but I'm not sure if it is or is not safe. If it isn't I have yet to run in to that case in testing

Eirenliel: we need @AxisAngles in the thread :D

AxisAngles: When the yaw is at 180 degrees, y will be exactly 1 or -1. So that case is completely safe.
The only time it is not safe is when the quaternion's up direction is facing exactly down. At that point, y and w will both be 0, and the resulting quaternion is null. However, this does not tend to happen in reality, so this is pretty safe.
When the bad case is encountered, there are infinitely many rotations which satisfy the "isolateYaw" constraint, and so it is reasonable and acceptable to return the identity quaternion.

Eirenliel: thank you for your wisdom

ButterscotchV: Sounds good, thanks for clarifying!

server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
ButterscotchV: We can chat on Discord later or i can make a commit for you if you want

Stermere: I'm happy to chat about it!

server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
```diff
@@ -300,13 +334,20 @@ private boolean preUpdate() {
...
 			initialized = true;
 		}
 
-		// if not enabled do nothing and return false
+		// update the foot length
+		footLength = skeleton.leftAnkleNode.worldTransform
+			.getTranslation()
+			.distance(
+				skeleton.leftFootNode.worldTransform.getTranslation()
+			);
```
Louka3000, line 348: skeleton.leftFootNode.localTransform.getTranslation().length should work for that

gui/public/i18n/en/translation.ftl
```diff
@@ -282,10 +281,20 @@ settings-general-fk_settings-leg_tweak-floor_clip = Floor clip
 # why the name - without this enabled the feet will often slide across the ground as if your skating across the ground,
 # since this largely prevents this it corrects for it hence skating correction (note this may be renamed to sliding correction)
 # definition - Guesses when each foot is in contact with the ground and uses that information to improve tracking
+
+
+
```
Louka3000, line 286: Why skip those lines?

Louka3000 (changes requested):

gui/public/i18n/en/translation.ftl
Stermere: making room for myself and forgetting to remove it...

Louka3000 (approved):

ButterscotchV (approved):

ButterscotchV (approved):

Eirenliel (approved):

PR #611, opened 2023-02-27
Rename resets

Louka3000:
Renamed "reset" to "full reset", and "quick reset" to "yaw reset".

We should check that most people agree with that renaming before merging ^^

Needs: https://github.com/SlimeVR/SolarXR-Protocol/pull/105


Louka3000:
Those were the ones that won the poll btw
https://i.gyazo.com/thumb/1200/f432d9789c48a0493428b56e65f0adf9-png.jpg

ButterscotchV (approved):

Eirenliel (approved):

PR #613, opened 2023-02-27
Fix vrconfig crashing on load

Louka3000:
- Check if chestOffset existed before negating it since it's new.
- Log exception on config load fail instead of throwing it.


ButterscotchV (approved):
This is an alright workaround for now, but it sucks to not notify the user aside from just logs... But keeping it working for the user is better than it crashing without any notifying

TheButlah (approved):

Eirenliel (approved):

PR #615, opened 2023-03-01
Add a ratio mode for manual offsets

### LINKED ISSUE #478: Simple manual offsets GUI
Now that offsets are linear in GUI, I think we should simplify setting body proportions manually.
I propose we have 2 modes: simplified and advanced.

Advanced being what we already have.
Simplified replacing certain offsets into ratios. For example, upperLeg-to-lowerLeg ratio, chest-to-torso ratio, etc. And use normal offsets for the rest.

This would allow changing the upper leg’s length while keeping the feet at the same y position if the user already has that much calibrated.

ImUrX:
Closes #478

This requires cleaning some of the code but I'm using functional code that react likes... ***sees @loucass003***

output


loucass003:
damn it looks pretty good

gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
```diff
@@ -40,156 +32,233 @@ function IncrementButton({
...
+                      }
+                    >
+                      {configFormat.format(-0.5)}
+                    </IncrementButton>
+                  )}
+                </div>
+                <div
+                  className="flex flex-grow flex-col px-2"
+                  onClick={() => {
+                    switch (type) {
```
loucass003, line 146: c'mon do a function you lazy rust programmer

ImUrX:
i have become a react developer :sob:

Louka3000 (approved):

TheButlah (approved):
Approving because I am needed for merge. I did not review anything but the 3 lines of rust that changed.

PR #616, opened 2023-03-01
[GUI] Improve window resizing behavior

TheButlah:
Helps mitigate https://github.com/tauri-apps/tauri/issues/4012

See https://github.com/tauri-apps/tauri/issues/4012#issuecomment-1449499149

### DIFF
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -13,7 +13,7 @@ use const_format::concatcp;
 use rand::{seq::SliceRandom, thread_rng};
 use shadow_rs::shadow;
 use tauri::api::process::Command;
-use tauri::Manager;
+use tauri::{Manager, WindowEvent};
 use tempfile::Builder;
 
 #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


gui/src-tauri/src/main.rs
```diff
@@ -229,9 +229,15 @@ fn main() {
 			}
 			Ok(())
 		})
-		//
+		.on_window_event(|e| match e.event() {
```
ImUrX, line 232: Can you ``#[cfg(windows)]`` this, or test it in other OSes?

ImUrX (changes requested):

gui/src-tauri/src/main.rs
TheButlah: you can if you wish. But this is a chromium problem presumably, so I think its safer to keep it unconditional. Who knows what the OS is using as their webview.

ImUrX: I have had 0 issues in Linux in my experience btw

FabianLars: > Who knows what the OS is using as their webview.

both linux and macos currently use a webkit based webview. It shouldn't have any (positive) effect on these platforms. Buuut maybe it helps on Linux where some users see white flickering when resizing (i don't have that problem on my system so i can't test it)
tldr: `#[cfg(windows)]` should be fine, or even recommended.

ImUrX (approved):

Issue #610, opened 2023-02-26
Defaults should not be initialized by VRConfig


ButterscotchV:
https://github.com/SlimeVR/SlimeVR-Server/blob/a0f3515981be7ae1b6fab2ed3de1726cb75c6f8a/server/src/main/java/dev/slimevr/config/VRConfig.java#L53-L80

These defaults should be set by the classes themselves, it feels really wrong to be doing it by VRConfig.

PR #617, opened 2023-03-01
fix: move default config to class 

marcozzxx810:
related to https://github.com/SlimeVR/SlimeVR-Server/issues/610

### DIFF
--- a/server/src/main/java/dev/slimevr/config/OSCConfig.java
+++ b/server/src/main/java/dev/slimevr/config/OSCConfig.java
@@ -7,10 +7,10 @@ public class OSCConfig {
 	private boolean enabled = false;
 
 	// Port to receive OSC messages from
-	private int portIn;
+	private int portIn = 9002;
 
 	// Port to send out OSC messages at
-	private int portOut;
+	private int portOut = 9000;
 
 	// Address to send out OSC messages at
 	private String address = "127.0.0.1";
--- a/server/src/main/java/dev/slimevr/config/VMCConfig.kt
+++ b/server/src/main/java/dev/slimevr/config/VMCConfig.kt
@@ -1,8 +1,14 @@
 package dev.slimevr.config
 
 class VMCConfig : // Address of the VRM to be used
-	OSCConfig() {
+	OSCConfig {
 	// Anchor the tracking at the hip?
+
+	constructor() : super() {
+		super.setPortIn(39540)
+		super.setPortOut(39539)
+	}
+
 	var anchorHip = true
 
 	var vrmJson: String? = null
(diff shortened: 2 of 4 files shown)


ImUrX:
there should be a config rewrite tbh

ImUrX (changes requested):
override the fields instead of making a constructor in kotlin please

marcozzxx810:
> override the fields instead of making a constructor in kotlin please

How should I use Kotlin to override java class? seem not possible 

Ref: https://stackoverflow.com/questions/64397076/how-to-override-a-java-class-property-variable-in-kotlin

Louka3000:
I did it like how it is because it didn't work properly otherwise. Is this tested?

server/src/main/java/dev/slimevr/config/VMCConfig.kt
```diff
@@ -3,7 +3,27 @@ package dev.slimevr.config
...
+		this.portIn = portIn
+	}
+
+	override fun getPortOut(): Int {
+		return portOut
+	}
+
+	override fun setPortOut(portOut: Int) {
+		this.portOut = portOut
+	}
```
ButterscotchV, line 28: Can't this just use override properties?Can't these getters and setters just be override properties?
https://kotlinlang.org/docs/inheritance.html#overriding-properties
```suggestion
	override var portIn: Int = 39540
	override var portOut: Int = 39539

	var anchorHip = true

	var vrmJson: String? = null
```

ButterscotchV (changes requested):

Louka3000:
boop

Eirenliel:
@Louka3000 resolve the conversations and conflicts, what's the boop about?

server/src/main/java/dev/slimevr/config/VMCConfig.kt
Eirenliel: Why does VMCConfig overrides all the same from OSCConfig... also just rewrite it all in kotlin without getters and setters.

server/src/main/java/dev/slimevr/config/OSCConfig.java
```diff
@@ -7,10 +7,10 @@ public class OSCConfig {
 	private boolean enabled = false;
 
 	// Port to receive OSC messages from
-	private int portIn;
+	private int portIn = 9002;
```
Eirenliel, line 10: The class gets extended and the fields get used in children, make it protected or rewrite in kotlin and make public.

Eirenliel (approved):

Louka3000:
> @Louka3000 resolve the conversations and conflicts, what's the boop about?

The boop is about Butter’s review not being taken care of

PR #620, opened 2023-03-04
Fix replace pitch

Stermere:
Just the fix to the replace pitch function in leg tweaks. Currently untested but I will leave a comment here when it has been if not merged already

### DIFF
--- a/server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
+++ b/server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
@@ -1168,18 +1168,10 @@ public class LegTweaks {
 
 	// return a quaternion that has been rotated by the new pitch amount
 	private Quaternion replacePitch(Quaternion quaternion, float newPitch) {
-		// first get the axis of the current pitch
-		Vector3 currentPitchAxis = quaternion.toMatrix().getX().unit();
-
-		// then get the current pitch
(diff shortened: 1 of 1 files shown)


server/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.java
```diff
@@ -1167,19 +1167,12 @@ private Quaternion isolateYaw(Quaternion quaternion) {
...
-
-		// then add the new pitch
-		Quaternion newQuat = Quaternion.Companion
-			.fromRotationVector(currentPitchAxis.times(newPitch + currentPitch));
-
-		return newQuat.times(quaternion);
-	}
+        EulerAngles curAngs = quaternion.toEulerAngles(EulerOrder.YXZ);
+        EulerAngles newAngs = EulerAngles(EulerOrder.YXZ, newPitch, curAngs.y, curAngs.z);
+        return newAngs.toQuaternion();
```
Louka3000, line 1174:
```suggestion
        EulerAngles angles = EulerAngles(EulerOrder.YZX, newPitch, 0f, 0f);
        return angles .toQuaternion();
```

Louka3000 (changes requested):

Louka3000:
oh wait I'm dumb

Louka3000 (approved):

Issue #364, opened 2022-12-11
Automatically disable Leg tweaks when using Autobone or Manual Body proportions


ZRock35:
Proportions have problems if it is active when doing those things

loucass003:
![image](https://user-images.githubusercontent.com/3948904/206927695-1b62ff41-061d-462c-9fb8-44a4fd2a9744.png)

ButterscotchV:
> Proportions have problems if it is active when doing those things

AutoBone is not affected by this as it already internally disables it, though when you try out your proportions after it'll affect that but idk that's just user error? We could add a "proportions mode" where it just disables any enhancements and whatnot for testing them or demonstrating them.

Spazznyan:
This was added yeh? close this one plx

Louka3000:
Implemented in https://github.com/SlimeVR/SlimeVR-Server/pull/627

PR #627, opened 2023-03-12
Disable leg tweaks on proportions

Stermere:
Turns off any leg corrections when the user is in the proportions tab of the GUI and addresses https://github.com/SlimeVR/SlimeVR-Server/issues/364. Not sure if the typescript I added is the right way to go about doing what it does so please check that :)

https://github.com/SlimeVR/SolarXR-Protocol/pull/106 Also needs to be fixed before this is merged


server/src/main/java/dev/slimevr/protocol/rpc/RPCHandler.java
```diff
@@ -257,6 +259,17 @@ public void onClearDriftCompensationRequest(
...
 
+	public void onGUIInfosResponse(
+		GenericConnection conn,
+		RpcMessageHeader messageHeader
+	) {
+		GUIInfosResponse req = (GUIInfosResponse) messageHeader.message(new GUIInfosResponse());
+		if (req == null)
+			return;
+
+		this.api.server.setLegTweaksEnabled(!req.inProportions());
```
ButterscotchV, line 270: What happens if you have legtweaks disabled in the config? Does this enable it even if you disabled it manually?

Stermere: nope! setLegTweaksEnabled is the master on/off for all 4 options (skating correction, floor clip, foot plant, and toe snap), It is the same way Autobone turns them off

ButterscotchV: All good, just wanted to double check

ButterscotchV (approved):

Eirenliel (approved):

PR #629, opened 2023-03-13
server: Implement non-blocking IO for VR bridges

0forks:
I forgot `FILE_FLAG_OVERLAPPED` and it was actually blocking at `ReadFile` on Windows.

Now it implements non-blocking IO for both receiving and sending, for the `UnixSocketBridge` too. It was tested on Linux, but only on synthetic data as I don't own a Linux-compatible VR headset. On Windows it felt slightly snappier and reduced hiccups, compared to the #607.

Latency became more consistent on Linux across different update rates. On Windows it's clear that it was receiving data with a delay of 1 frame.

Time delta measurements between sending headset position and receiving a new tracker position:

| Platform / Update rate / Num frames | Old (ms) | New (ms) |
| -------- | ---- | --- |
|                        | `   p1   avg    p99`  | `   p1   avg   p99` |
| Windows / 90hz / 5000  | `0.089 0.134  0.203`  | `0.538 1.746 4.103` |
| Windows / 100hz / 5000 | `0.086 0.130  0.213`  | `0.538 1.721 4.094` |
| Windows / 120hz / 5000 | `0.076 0.192  0.253`  | `0.550 1.844 4.123` |
| Linux / 90hz / 5000    | `0.117 1.685 11.799`  | `0.172 1.305 2.196` |
| Linux / 100hz / 5000   | `0.102 0.764 10.710`  | `0.157 1.086 1.815` |
| Linux / 120hz / 5000   | `0.087 0.622  8.701`  | `0.161 1.288 1.835` |

image

Please review and test.


kitlith (approved):
SlimeVR-Latency concurs, it's no longer a frame behind. I should really write code to make it give statistics like the ones you're collecting. (I'm curious, what are you using to collect those statistics currently?)

the difference between windows old and windows new looking really bad, but feeling better sounds like there's something going on with the measurement there (are the newer measurements more accurate?)

I suspect this might still be handling some details w/ overlapped io incorrectly (i.e. we should be inserting explicit waits/getoverlappedresult anywhere we want a blocking-style call) but if it works (which it appears to do so) i'm not going to nitpick about it.

0forks:
> what are you using to collect those statistics

100 lines of JS and plotly, it was faster to write a crossplatform tester in JS. Though it's always async and buffered (and slower than native), so doesn't represent the driver latency exactly. But should be close enough.

> the difference between windows old and windows

In the old version it's a delayed tracker frame, which comes instantly when the thread unblocks after receiving a headset frame. The new latency is what it should have been instead.

> I suspect this might still be handling some details w/ overlapped io incorrectly (i.e. we should be inserting explicit waits/getoverlappedresult anywhere we want a blocking-style call)

I'm calling `ReadFile` with 0 bytes to read, and `GetOverlappedResult` is only used to get amount of bytes read. `ReadFile` should always instantly return `false` with a `GetLastError` either 0 or IO pending. We're doing a select with `WaitForMultipleObjects` on auto-resetting events, so technically we should be fine here.

Indeed, there could be something I missed, so I want more people to review this to make sure all of it is implemented correctly, as I haven't touched Java and WinAPI in years.

ButterscotchV (approved):

Eirenliel (approved):

Eirenliel:
So does it make it worse, I don't get it...

ButterscotchV:
> So does it make it worse, I don't get it...

It looks to me like it's taking about 1 ms to get the update now, but the old one was taking the time to get the update *and* it was one frame behind, so for 100 Hz, that's 10 ms + the extra time. I think this is how to interpret it? Please let me know if I got something wrong.

0forks:
> I think this is how to interpret it?

Yes, this should reduce latency in Windows SteamVR by ~8-11ms or more (1 frame) depending on your headset's update rate. Because the bridge thread was waking up only after receiving data, the new tracker positions were stuck until the next frame. This implementation signals the thread when to send too, so it no longer happens.

The previous implementation works and is okay because frame time consistency is more important than latency when presenting frames.

The lower half of the bar chart is the new version, now it takes 1-2ms on average to receive a response by the (simulated) driver on both platforms, if I'm measuring it right. Please do test if there is any perceivable difference.

ImUrX:
yeah, I havent been able to test this yet, but I will try it tomorrow

0forks:
> we should be inserting explicit waits/getoverlappedresult anywhere we want a blocking-style call

Ah, I just realized you were talking about the part where we read the data, not the waits. Indeed the docs say:

> If hFile is opened with FILE_FLAG_OVERLAPPED, the lpOverlapped parameter must point to a valid and unique [OVERLAPPED](https://learn.microsoft.com/en-us/windows/desktop/api/minwinbase/ns-minwinbase-overlapped) structure, otherwise the function can incorrectly report that the read operation is complete.

As we're peeking and know how many bytes are available, the read operation returns true immediately with data. If not - pipe will close. Maybe we should handle the async case indeed, but JNA doesn't seem to expose `GetOverlappedResult`.

ButterscotchV (approved):
Looks good, I'd be interested to see the difference in performance from these changes?

0forks:
No measurable difference after the last 2 commits:

| Platform / Update rate / Num frames  | 2cc1201 / before (ms) | 9f45d9d / after (ms) |
| ---------------------- | ------------------- | ------------------- |
|                        | `   p1   avg   p99` | `   p1   avg   p99` |
| Windows / 90hz / 5000  | `0.539 1.835 4.110` | `0.543 1.803 4.111` |
| Windows / 100hz / 5000 | `0.536 1.764 4.098` | `0.543 1.825 4.103` |
| Windows / 120hz / 5000 | `0.540 1.806 4.116` | `0.545 1.791 4.108` |

image

0forks:
Now comparing to v0.6.3:

| Platform / Update rate / Num frames  | v0.6.3 (ms) | 9f45d9d (ms) |
| ---------------------- | ------------------- | ------------------- |
|                        | `   p1   avg    p99` | `   p1   avg   p99` |
| Windows / 90hz / 5000  | `0.083 9.452 20.208` | `0.546 1.774 4.134` |
| Windows / 100hz / 5000 | `0.071 9.857 19.148` | `0.541 1.773 4.122` |
| Windows / 120hz / 5000 | `0.059 4.196 14.086` | `0.541 1.786 4.115` |

image

kitlith:
> JNA doesn't seem to expose `GetOverlappedResult`.

I ran into that when I was doing my own try at overlapped. This was the approach I took to gain access to `GetOverlappedResult`: https://github.com/kitlith/SlimeVR-Server/blob/unpoll_openvr_thread/server/src/main/java/dev/slimevr/platform/windows/Kernel32.java

Should really get around to submitting an upstream issue.

0forks:
Looking deeper, it appears that the sleep precision in the VRServer thread is the limiting factor now.
Replacing `Thread.sleep(1)` with `LockSupport.parkNanos(1000000)` improves the latency on Windows to 0.6-1.6ms, but also results in slightly higher CPU usage.

As of now, I believe the latency is already acceptable.

This is how long it sleeps for:
| OS      | Method                         |      p1     |     avg     |     p99     |
|---------|--------------------------------|-------------|-------------|-------------|
| Windows | Thread.sleep(1)                | 1.008800 ms | 2.511393 ms | 4.000000 ms |
| Windows | LockSupport.parkNanos(1000000) | 0.862300 ms | 0.985986 ms | 1.477500 ms |
| Linux   | Thread.sleep(1)                | 1.078178 ms | 1.456425 ms | 1.627606 ms |
| Linux   | LockSupport.parkNanos(1000000) | 1.073479 ms | 1.468993 ms | 1.597515 ms |

0forks:
Now with VRServer waking on data the p99 latency is under 1ms and it looks smoother under load in VR. Still only properly tested on Windows.

0forks:
Implemented async IPC and pose updates in the driver, this nuked the remaining 1 frame of latency - body is perfectly in sync with the head now. Will make another PR after cleaning up and more code validation.

ImUrX:
This seems working nicely, only problem is that apparently the Linux driver has some issues :sob:

ImUrX (approved):

Louka3000 (changes requested):
This prevents sending data to OSC when SteamVR is open

Louka3000 (approved):

ButterscotchV (approved):
Using `Object.wait()` and `Object.notify()` looks like a much better solution than `Thread.sleep()` and `Thread.interrupt()`.
It's worth noting that `Object.notify()` will only wake a single thread waiting on the object, whereas `Object.notifyAll()` will wake all threads. It won't have any difference in effect for this implementation and I can't really foresee any future where multiple threads would be waiting on `wakeupLock`, but I figure I'd just kinda mention it. Maybe someone else can think of a case where that would be a problem?

PR #632, opened 2023-03-19
feat: VRServer is gracefully shutdown

marcozzxx810:
- [ ] add log cleanup function to shutdown Hook
- [ ] setting up SIGTERM for window and Linux system

### DIFF
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -12,10 +12,10 @@ use clap::Parser;
 use const_format::concatcp;
 use rand::{seq::SliceRandom, thread_rng};
 use shadow_rs::shadow;
-use tauri::api::process::Command;
-use tauri::Manager;
+use tauri::api::process::{Command, CommandChild};
 #[cfg(windows)]
 use tauri::WindowEvent;
+use tauri::{Manager, RunEvent};
 use tempfile::Builder;
 
 #[cfg(windows)]
@@ -31,6 +31,10 @@ static POSSIBLE_TITLES: &[&str] = &[
 	"never gonna let you down",
 	"uwu sowwy",
 ];
+
+#[derive(Default)]
+struct Backend(Option<CommandChild>);
+
 shadow!(build);
 // Tauri has a way to return the package.json version, but it's not a constant...
 const VERSION: &str = if build::TAG.is_empty() {
@@ -173,6 +177,7 @@ fn main() {
 	}
 
 	// Spawn server process
+	let mut backend = Backend::default();
 	let run_path = get_launch_path(cli);
 
 	let stdout_recv = if let Some(p) = run_path {
@@ -190,18 +195,19 @@ fn main() {
 		};
 
 		log::info!("Using Java binary: {:?}", java_bin);
-		let (recv, _child) = Command::new(java_bin.to_str().unwrap())
+		let (recv, child) = Command::new(java_bin.to_str().unwrap())
 			.current_dir(p)
 			.args(["-Xmx512M", "-jar", "slimevr.jar", "--no-gui"])
 			.spawn()
 			.expect("Unable to start the server jar");
+		_ = backend.0.insert(child);
 		Some(recv)
 	} else {
 		log::warn!("No server found. We will not start the server.");
 		None
 	};
 
-	let run_result = tauri::Builder::default()
+	let build_result = tauri::Builder::default()
 		.plugin(tauri_plugin_window_state::Builder::default().build())
 		.setup(|app| {
 			if let Some(mut recv) = stdout_recv {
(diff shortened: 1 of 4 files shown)


gui/src-tauri/src/main.rs
```diff
@@ -31,6 +32,10 @@ static POSSIBLE_TITLES: &[&str] = &[
 	"never gonna let you down",
 	"uwu sowwy",
 ];
+
+#[derive(Default)]
```
TheButlah, line 36: Add a doc comment explaining what this struct is for.

gui/src-tauri/src/main.rs
```diff
@@ -237,8 +244,23 @@ fn main() {
...
-	match run_result {
+		.build(tauri::generate_context!());
+	match build_result {
+		Ok(app) => {
+			app.run(move |_app_handle, event| match event {
+				RunEvent::ExitRequested { .. } => {
+					if let Some(mut child) = backend.0.take() {
+						let write_result = child.write(b"exit\n");
+						match write_result {
+							Ok(()) => log::info!("send exist to backend"),
```
TheButlah, line 255:
```suggestion
							Ok(()) => log::info!("send exit to backend"),
```

gui/src-tauri/src/main.rs
```diff
@@ -237,8 +244,23 @@ fn main() {
...
+        					Err(_) => log::info!("fail to send exist to backend"),
```
TheButlah, line 256:
```suggestion
        					Err(_) => log::info!("fail to send exit to backend"),
```

gui/src-tauri/src/main.rs
```diff
@@ -258,7 +280,9 @@ fn main() {
 			}
 			return;
 		}
-		_ => run_result.expect("error while running tauri application"),
+		Err(_) => {
+			println!("Unknown Error")
```
TheButlah, line 284: This should be a tauri message dialog, right? println is not useful for an app that is double clicked like this.

gui/src-tauri/src/main.rs
```diff
@@ -237,8 +244,23 @@ fn main() {
...
+		Ok(app) => {
+			app.run(move |_app_handle, event| match event {
+				RunEvent::ExitRequested { .. } => {
+					if let Some(mut child) = backend.0.take() {
+						let write_result = child.write(b"exit\n");
+						match write_result {
+							Ok(()) => log::info!("send exist to backend"),
+        					Err(_) => log::info!("fail to send exist to backend"),
+						}
+						thread::sleep(Duration::from_millis(10000));
```
TheButlah, line 258:
```suggestion
```
I would assume that sleeping is unecessary, you don't want to block the closure that is processing events.

TheButlah (changes requested):

gui/src-tauri/src/main.rs
marcozzxx810: added

gui/src-tauri/src/main.rs
marcozzxx810: changed

gui/src-tauri/src/main.rs
marcozzxx810: sleeping is necessary, since after sending exit to server. it takes some time for the server to clean up. 

If remove the sleep, the server will receive a kill signal which cleanup may not finish.

Why sleep? I think this just gives 10s to clean up or else kill the signal.

TheButlah: what happens if the server cleans up before the 10s finishes?

gui/src-tauri/src/main.rs
```diff
@@ -190,18 +194,19 @@ fn main() {
 		};
 
 		log::info!("Using Java binary: {:?}", java_bin);
-		let (recv, _child) = Command::new(java_bin.to_str().unwrap())
+		let (recv, child) = Command::new(java_bin.to_str().unwrap())
 			.current_dir(p)
 			.args(["-Xmx512M", "-jar", "slimevr.jar", "--no-gui"])
 			.spawn()
 			.expect("Unable to start the server jar");
+		_ = backend.insert(child);
```
TheButlah, line 202: I didn't actually know this was possible, but I tried it in rust playground and it works.
@ImUrX do you have a style opinion on whether we should to it this way or instead use `let _ = `?

ImUrX: huh, thats cool. Still I have some ideas on how to do this, I dont like the use of stdin for exiting but I understand it's mainly because of Windows limitations in signals :/

gui/src-tauri/src/main.rs
marcozzxx810: it just wait 10s before closing, the window is closed. User doesn't feel anything.

@ImUrX has add some comment about this. There is some limitation about windows.

ImUrX: https://github.com/tauri-apps/tauri/issues/6634

FabianLars: (i'm coming from the tauri issue you opened :P )
Another workaround would be to use something like an AtomicBool which you then change in your `CommandEvent::Terminated` and check for its value in ExitRequested in a loop - probably want to add a max duration as a fallback idk.

ImUrX: haha, yeah seems like an idea lmao

TheButlah (approved):
Deferring my review to @ImUrX since they are making changes.

gui/src-tauri/src/main.rs
```diff
@@ -148,12 +156,14 @@ fn main() {
...
-					// We can't wait the process to die because Tauri doesn't expose
-					// such thing, so the only way is sleeping and hope for the best
-					// for now...
-					// (We can wait pids in POSIX but not in Windows if we wanted
-					// to workaround this)
-					thread::sleep(Duration::from_secs(10))
+					let ten_seconds = Duration::from_secs(10);
+					let start_time = Instant::now();
+					while start_time.elapsed() < ten_seconds {
+						if exit_flag_requested.load(Ordering::Relaxed) {
```
ImUrX, line 162: please just use ``exit_flag`` directly instead of making a new clone to the ``Arc``

PR #641, opened 2023-03-25
Mounting reset improvements

ButterscotchV:
- Clears mounting reset when a mounting rotation is set manually
- Bases mounting reset off of the current HMD rotation rather than the last full reset


Eirenliel (approved):

ButterscotchV:
Needs testing, there's a released beta with instructions. Will report back whenever anyone has tested.

PR #644, opened 2023-03-29
Revert VRServer tick interval

0forks:
There seems to be an issue with skating correction, it can't handle consistent ticks at 1ms, for now we need to partially revert #629 until we understand why it breaks.

It is possible that skating correction is broken on Linux, as sleeps there are faster.


Louka3000 (approved):

0forks:
I can retrigger CI with a force push if necessary, [there was an outage](https://www.githubstatus.com/incidents/z3c6q056q332), CI [did build it in my fork](https://github.com/0forks/SlimeVR-Server/actions/runs/4555635505) eventually, after almost an hour.

ButterscotchV (approved):
Looks good, but I don't see why most of this code can't just be left in and left unused for now.

PR #651, opened 2023-04-05
Fix default tracker mounting state to be front correctly

Eirenliel:
By default GUI shows the trackers are mounted to front, as they should, when the mounting orientation is null. But server loads identity for orientation when it's null, which is wrong, because FRONT is 180deg. This leads to wrong default state for frount moutned trackers. It can be fixed if orentation is changed to any other side and then to front again. This PR fixes this.

Found with Poly during testing of official slimes.


Louka3000 (approved):
lgtm

PR #663, opened 2023-04-07
Theme support

ImUrX:
Because of @ButterscotchV's trans day of visibility SlimeVR version. I present to all of you, theme support (with trans theme included!)
output


TheDevMinerTV:
Thank you for adding flashbang mode. :)

gui/src/components/commons/BaseModal.tsx
```diff
@@ -17,14 +17,17 @@ export function BaseModal({
       overlayClassName={
         props.overlayClassName ||
         classNames(
-          'fixed top-0 right-0 left-0 bottom-0 flex flex-col justify-center items-center w-full h-full bg-background-90 bg-opacity-60 z-20'
+          'fixed top-0 right-0 left-0 bottom-0 flex flex-col justify-center',
```
TheDevMinerTV, line 20: Please use multiline strings (with backticks), classnames has to run which might result in flicker.

gui/src/components/commons/BaseModal.tsx
```diff
@@ -17,14 +17,17 @@ export function BaseModal({
...
+          'items-center w-full h-full bg-background-90 bg-opacity-60 z-20'
         )
       }
       className={
         props.className ||
         classNames(
           props.className as string,
-          'items-center focus:ring-transparent focus:ring-offset-transparent focus:outline-transparent outline-none bg-background-60 p-6 rounded-lg text-white'
+          'items-center focus:ring-transparent focus:ring-offset-transparent',
```
TheDevMinerTV, line 28: Please use multiline strings, not multiple, `classNames` has a tendency to not apply the 2nd string

gui/src/components/commons/icon/MaximiseIcon.tsx
```diff
@@ -1,17 +1,16 @@
+import classNames from 'classnames';
+
 export function MaximiseIcon({ className }: { className?: string }) {
   return (
     <svg
       width="35"
       height="35"
-      className={className}
+      className={classNames('stroke-window-icon', className)}
```
TheDevMinerTV, line 8: It would be more ideal to use
```js
className={`stroke-window-icon ${className}`}
```
since it doesn't run any/less code at runtime.

gui/src/components/commons/icon/MinimiseIcon.tsx
```diff
@@ -1,13 +1,15 @@
+import classNames from 'classnames';
+
 export function MinimiseIcon({ className }: { className?: string }) {
   return (
     <svg
       width="35"
       height="35"
-      className={className}
+      className={classNames('stroke-window-icon', className)}
```
TheDevMinerTV, line 8: It would be more ideal to use
```js
className={`stroke-window-icon ${className}`}
```
since it doesn't run any/less code at runtime.

gui/src/components/commons/icon/CloseIcon.tsx
```diff
@@ -9,21 +11,13 @@ export function CloseIcon({
     <svg
       width={size}
       height={size}
-      className={className}
+      className={classNames('stroke-window-icon', className)}
```
TheDevMinerTV, line 14: It would be more ideal to use
```js
className={`stroke-window-icon ${className}`}
```
since it doesn't run any/less code at runtime.

gui/src/components/commons/Radio.tsx
```diff
@@ -35,7 +35,10 @@ export function Radio({
         >
           <input
             type="radio"
-            className="text-accent-background-30 focus:ring-transparent focus:ring-offset-transparent focus:outline-transparent"
+            className={classNames(
```
TheDevMinerTV, line 38: Please use multiline strings (with backticks), classnames has to run which might result in flicker.

gui/src/components/commons/ThemeSelector.tsx
```diff
@@ -0,0 +1,46 @@
+import classNames from 'classnames';
+import { Control, Controller } from 'react-hook-form';
+import { Typography } from './Typography';
```
TheDevMinerTV, line 3: Unused import here.

gui/src/components/commons/ThemeSelector.tsx
```diff
@@ -0,0 +1,46 @@
...
+            WebkitTextFillColor: 'transparent',
+          }}
+          name={name}
+          ref={ref}
+          onChange={onChange}
+          value={value}
+          checked={value == checked}
+          disabled={disabled}
+          {...props}
+        ></input>
```
TheDevMinerTV, line 42: Please just `/>`, otherwise someone might get tempted to insert something inside of `<input>` -> this will render the HTML invalid.

TheDevMinerTV (changes requested):

gui/src/components/commons/ThemeSelector.tsx
ImUrX: this is copied from `Radio`, I will edit there too

gui/src/components/commons/BaseModal.tsx
ImUrX: I find this kind of optimization unnecesary, mainly to put in this PR. Check because there is a lot of components that do this, make an issue for this :angry:

TheDevMinerTV: Might as well just start right now here in the components you touch?

gui/src/components/commons/BaseModal.tsx
loucass003: Wrong it does work with multiple string. Also Using multiline string create wierd class definition in the dom. It does not remove the \n for some reasons. this is why we are not using it. I havent checked in a while if it is still like that tho, so might be worth checking again.

gui/src/components/commons/BaseModal.tsx
loucass003: i do not agree with this. Keep classNames please

gui/src/components/commons/BaseModal.tsx
loucass003: Also multiline string does not allow for dynamic classes without ending up with really ugly code or having a string with a bunch of empty spaces

gui/src/components/commons/BaseModal.tsx
loucass003: ``classnames has to run which might result in flicker`` hello no wtf

gui/src/components/commons/BaseModal.tsx
TheDevMinerTV: Okay, fair

gui/src/components/commons/BaseModal.tsx
TheDevMinerTV: pff, fineeeeee

TheDevMinerTV (approved):

PR #664, opened 2023-04-08
Repair icons for linux

ImUrX:
linux builds currently dont run because of an optimization from imgbot


TheButlah (approved):

Eirenliel:
how can one ruin a png?? 😭

PR #647, opened 2023-04-01
Trackers rewrite & Switch to new math lib

Louka3000:
- Rewrites the trackers to a single class with flags and logic handlers instead of abstractions.
- Use a new immutable 3d math lib by AxisAngle.

Improvements (use for 0.7.2 changelog):
- Makes default tracker settings consistent (individual drift comp. always enabled)
- Improve drift compensation a lot! (Works lying down)
- Disconnected or errored trackers won’t affect skeleton until they go back to OK

PR #668, opened 2023-04-11
Send Tracker Battery Info to SteamVR

NWBx01:
Adds support for sending tracker battery info from the SlimeVR Server to SteamVR.
- Trackers that do not have battery info are not affected.
- Trackers that do have battery info will report their battery percentage, as well as whether they're charging or not.
- Works for owoTrack devices (However, these will not display whether they're charging, as owoTrack does not send battery voltage info)

This depends on another pull request. (https://github.com/SlimeVR/SlimeVR-OpenVR-Driver/pull/41)
Requires that trackers-rewrite be merged first (https://github.com/SlimeVR/SlimeVR-Server/pull/647)


Louka3000:
Please target main directly, and precise that this PR requires trackers-rewrite to be merged

NWBx01:
Changing to main artificially added a bunch of commits because it was a fork of trackers-rewrite, but I guess that's okay?...

Louka3000:
Ye, doesn’t matter, it’ll squash on merge. (And diff will be fine when trackers-rewrite gets merged into main)

Louka3000 (changes requested):
writeBatteryUpdate() is a whole lot of unnecessary code that makes it hard to read.
- Parentheses don't really help IMO
- Why do trackers that are alone (not waist or feet) check for lowestBattery?
- Waist and feet can use the same code. Get the trackers (tracker1 and tracker2) in an if or switch case, then handle them all the same outside. Everything else can also use the same code. Use `TrackerPosition.getByTrackerRole(TrackerRole)`

NWBx01:
I'm not sure what exactly you meant by using `TrackerPosition.getByTrackerRole(TrackerRole)`. I don't understand what would be returned for ankle/foot, waist/hip, or upper/low Arm. If it can return two TrackerPosition values I can see how that could eliminate the switch altogether.

server/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -125,6 +130,132 @@ protected void writeTrackerUpdate(Tracker localTracker) {
...
+
+		List<Tracker> allTrackers = Main.getVrServer().getAllTrackers();
+		TrackerRole role = localTracker.getTrackerPosition().getTrackerRole();
+
+		Tracker primaryTracker = null;
+		Tracker secondaryTracker = null;
+		Tracker tertiaryTracker = null;
+
+		// Given what the role is of localTracker, the tracker positions that
+		// make up that role are set to primaryTracker and secondaryTracker
```
Louka3000, line 153: forgor comment for third tracker

server/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -125,6 +130,132 @@ protected void writeTrackerUpdate(Tracker localTracker) {
...
+				&& (tertiaryTracker.getBatteryLevel() < lowestLevel)
+		) {
+			lowestLevel = tertiaryTracker.getBatteryLevel();
+
+			if (tertiaryTracker.getBatteryVoltage() != null) {
+				trackerVoltage = tertiaryTracker.getBatteryVoltage();
+			}
+		}
+
+		if (lowestLevel == 200) {
```
ButterscotchV, line 240:
```suggestion
		if (lowestLevel >= 200) {
```

server/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -125,6 +130,132 @@ protected void writeTrackerUpdate(Tracker localTracker) {
...
+		primaryTracker = TrackerUtils
+			.getNonInternalTrackerForBodyPosition(
+				allTrackers,
+				TrackerPosition.getByTrackerRole(role)
+			);
+		switch (role) {
+			case WAIST:
+				secondaryTracker = TrackerUtils
+					.getNonInternalTrackerForBodyPosition(allTrackers, TrackerPosition.WAIST);
+				break;
```
ButterscotchV, line 164: Why is this getting the waist tracker again? Wouldn't both `primaryTracker` and `secondaryTracker` be the same tracker in this case?
It seems like you'd be wanting to do this instead:
```suggestion
			case WAIST:
				secondaryTracker = TrackerUtils
					.getNonInternalTrackerForBodyPosition(allTrackers, TrackerPosition.CHEST);
				tertiaryTracker = TrackerUtils
					.getNonInternalTrackerForBodyPosition(allTrackers, TrackerPosition.HIP);
				break;
```

ButterscotchV (changes requested):
I'd like to see this small change just because, otherwise I think there could be an improvement by taking into account which SteamVR trackers are enabled, as the chest tracker being enabled could display more information (not aggregating the percentages from the other spine trackers), same with knees.

ButterscotchV:
> I'd like to see this small change just because, otherwise I think there could be an improvement by taking into account which SteamVR trackers are enabled, as the chest tracker being enabled could display more information (not aggregating the percentages from the other spine trackers), same with knees.

Just gonna quickly note, I'm not requiring this improvement, but we could have a TODO for it and a GitHub Issue for it once this is merged. It would be nicer to have this done than to have it held back over an unecessary complexity.

server/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
NWBx01: lowestLevel here is initialized to 200. Practically speaking, it realistically shouldn't go above 100% unless there's some error with battery percentage reporting. So, if after going through every check it's still "200%", the trackers in question don't report any battery level. There's probably a nicer way to go about doing this, like setting it as null I suppose.

ButterscotchV: I understand the purpose, this reduces the number of checks required within the code. I just simply propose this, as *any* value over 200 is unreasonable and can be entirely assumed to be invalid.

server/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
NWBx01: Chest should be its own tracking point in this case, as no other SlimeVR tracking points merge to a single SteamVR tracker for it. It should always be caught by:
`primaryTracker = TrackerUtils.getNonInternalTrackerForBodyPosition(allTrackers, TrackerPosition.getByTrackerRole(role))`.
Likewise, because there is no case for chest for `secondaryTracker`, it will never be part of multiple trackers for checking battery life.

> Wouldn't both `primaryTracker `and `secondaryTracker` be the same tracker in this case?

I think this is a result of confusion between `TrackerRole` and `BodyPart`. Hip is  `TrackerRole.WAIST`, and `BodyPart.HIP` whereas waist is null and `BodyPart.WAIST`. As I understand it, `TrackerPosition.getByTrackerRole(role)` should only return a tracker which is of `TrackerRole`, this should naturally exclude: waist, lower leg, and lower arm from being `primaryTracker` (because their `TrackerRole` is null), then leaving these aforementioned body parts to only be set as `secondaryTracker`. It is a bit confusing, I admit...

Your intuition is sort of correct though. This _could_ cause an issue with shoulder trackers in particular because they do not have a `TrackerRole` of null, however, uniquely this tracker position does not get a unique SteamVR tracking point, and is instead merged with elbow, so -- I believe -- there shouldn't be any issues with shoulder trackers overwriting the battery level of lower/upper arm trackers. I could be wrong on this, but as it is, shoulder + upper/lower arm is a very rare configuration and I haven't had anyone mention it as an issue; I should test this to see if it causes any issues.

NWBx01: Chest should be its own tracking point in this case, as no other SlimeVR tracking points merge to a single SteamVR tracker for it. It should always be caught by:
`primaryTracker = TrackerUtils.getNonInternalTrackerForBodyPosition(allTrackers, TrackerPosition.getByTrackerRole(role))`.
Likewise, because there is no case for chest for `secondaryTracker`, it will never be part of multiple trackers for checking battery life.

NWBx01:
> I think there could be an improvement by taking into account which SteamVR trackers are enabled, as the chest tracker being enabled could display more information (not aggregating the percentages from the other spine trackers), same with knees.

If the SteamVR tracking point for chest is enabled, the SlimeVR tracker of `TrackerRole.CHEST` should have its battery level sent to its respective SteamVR tracker. Likewise for knees. The only tracking points that aggregate their battery level are those trackers that have a `TrackerRole` of null: WAIST, LEFT_LOWER_LEG, RIGHT_LOWER_LEG, LEFT_LOWER_ARM, and RIGHT_LOWER_ARM. Despite not having a `TrackerRole` of null, LEFT_SHOULDER, RIGHT_SHOULDER also aggregate their battery level into a single SteamVR tracking point for elbows.

I had asked about how and which battery level should be sent and the consensus was to send the battery level of the lowest SlimeVR tracker; if foot < ankle; foot gets sent, and vice versa. Upon the SlimeVR tracker dying, it will therefore no longer be the tracker with the lowest battery level for a given SteamVR tracking point, and the battery level will then update to the battery level of whichever tracker now has the lowest battery level. In practice that might look like the the tracker having a battery level of 5%, dying, and then updating to 35%. From what I understand, now when trackers die or are disconnected, they should now also be removed from SteamVR so this process of trackers dying and their battery life updating should be rather seamless when multiple SlimeVR trackers combine to a single SteamVR tracking point. It's obviously not a 1:1 mapping of battery level, but it's the best that can be done without having each SlimeVR tracking point map to its own SteamVR tracking point.

NWBx01:
> I'd like to see this small change just because, otherwise I think there could be an improvement by taking into account which SteamVR trackers are enabled, as the chest tracker being enabled could display more information (not aggregating the percentages from the other spine trackers), same with knees.

The recent changes I made should take care of this.

server/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
NWBx01: Ah, I see what you mean.

ButterscotchV (approved):

server/src/main/java/dev/slimevr/bridge/ProtobufMessages.java
Eirenliel, line 1: Why are there 10k changes?

Eirenliel (approved):

Louka3000 (approved):

server/src/main/java/dev/slimevr/bridge/ProtobufMessages.java
NWBx01: ProtobufMessages is autogenerated by the Protocol Buffer compiler and based on ProtobufMessages.proto in the SlimeVR-OpenVR-Driver.

You can see the changes here:
https://github.com/SlimeVR/SlimeVR-OpenVR-Driver/pull/41

PR #670, opened 2023-04-13
Make config migration more robust

Louka3000:
Please release 0.7.1 along with this

### DIFF
--- a/server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
+++ b/server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
@@ -3,6 +3,7 @@ package dev.slimevr.config;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.*;
 import com.github.jonpeterson.jackson.module.versioning.VersionedModelConverter;
+import io.eiren.util.logging.LogManager;
 
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
```diff
@@ -18,188 +19,212 @@ public ObjectNode convert(
...
+			}
+			if (version < 5) {
+				// Migrate old skeleton offsets to new ones
+				ObjectNode skeletonNode = (ObjectNode) modelData.get("skeleton");
+				if (skeletonNode != null) {
+					ObjectNode offsetsNode = (ObjectNode) skeletonNode.get("offsets");
+					if (offsetsNode != null) {
+						// torsoLength, chestDistance and waistDistance become
+						// chestLength, waistLength and hipLength.
+						float torsoLength = 0, chestDistance = 0, waistDistance = 0;
```
ButterscotchV, line 116: This isn't very desirable to set missing values to `0`, can we do default lengths here? Neither solution is perfect, but if anything goes wrong, I'd rather it goes the least wrong.

server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
```diff
@@ -18,188 +19,212 @@ public ObjectNode convert(
...
+					if (skeletonNode == null) {
+						skeletonNode = nodeFactory.objectNode();
+					}
 
+					ObjectNode offsetsNode = nodeFactory.objectNode();
+					while (bodyIter.hasNext()) {
+						Map.Entry<String, JsonNode> node = bodyIter.next();
+						// Filter only number values because other types would
+						// be
+						// stuff that didn't get migrated correctly before
```
ButterscotchV, line 67: Weird comment splitting, should be easy to fix

server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
```diff
@@ -18,188 +19,212 @@ public ObjectNode convert(
...
+								new FloatNode(torsoLength - chestDistance - waistDistance)
 							);
+						offsetsNode.set("hipLength", offsetsNode.get("waistDistance"));
+						offsetsNode.remove("torsoLength");
+						offsetsNode.remove("chestDistance");
+						offsetsNode.remove("waistDistance");
+
+						// legsLength and kneeHeight become
+						// upperLegLength and lowerLegLength
+						float legsLength = 0, kneeHeight = 0;
```
ButterscotchV, line 139: Same as the comment before about using the default length values, is that possible?

server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
```diff
@@ -18,188 +19,212 @@ public ObjectNode convert(
...
+					tapDetectionNode
+						.set("yawResetEnabled", tapDetectionNode.get("quickResetEnabled"));
+					tapDetectionNode.set("fullResetEnabled", tapDetectionNode.get("resetEnabled"));
+					tapDetectionNode.set("yawResetTaps", tapDetectionNode.get("quickResetTaps"));
+					tapDetectionNode.set("fullResetTaps", tapDetectionNode.get("resetTaps"));
+				}
 			}
+		} catch (Exception e) {
+			LogManager.severe("Error during config migration: " + e);
 		}
```
ButterscotchV, line 228: If anything goes wrong, do we make a backup of the original config? We most definitely should. If it's not viable to do it on an error, then do it before any config migration is done.

ButterscotchV (changes requested):

server/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
Louka3000: Not sure that’s possible with how it is currently set, but Uriel wants to rewrite the whole config system or something anyways…

ImUrX: I think butter is saying to just copy the config file and put it into a .old one

Louka3000: I know, but does the class have access to the file itself? All I see is that it has access to the content… not even the file path, and we can’t access vrServer because it’s in lateinit

ImUrX: sigh, yeah lets leave it for the rewrite

ButterscotchV: Unfortunate, probably "okay" then as long as it's fairly reliable

ButterscotchV (approved):

ImUrX (approved):

Eirenliel (approved):

PR #671, opened 2023-04-14
Additional Theme Options

NWBx01:
Added simple hue shifted themes:
- Green
SlimeVR 4_13_2023 11_56_08 PM
- Yellow
SlimeVR 4_13_2023 11_56_11 PM
- Orange
SlimeVR 4_13_2023 11_56_22 PM
- Red
SlimeVR 4_13_2023 11_56_28 PM
- Dark
SlimeVR 4_13_2023 11_56_31 PM
- Light
SlimeVR 4_13_2023 11_56_34 PM


Louka3000:
I'm not a fan of just stuffing a bunch of different hues as themes for no reasons except "hey we got more themes and colours!". At this rate we'll have too many themes. Open to counter-arguments tho if people do enjoy being able to choose their favourite colours as themes or something tho.
Bland dark and light themes are a good idea tho, but they need better contrast (and maybe make the dark theme a tiny bit darker?)

NWBx01:
> I'm not a fan of just stuffing a bunch of different hues as themes for no reasons except "hey we got more themes and colours!". At this rate we'll have too many themes.

At the moment, they're called "Color theme". There's not really much room for in-depth customization other than changing the color. To be quite honest, I'm not sure what else you would do with themes otherwise.

> Bland dark and light themes... need better contrast (and maybe make the dark theme a tiny bit darker?)

The contrast of all the themes should be the same since they're a hue shift. It's hard to explain without seeing it yourself, but trying to do themes that have different levels of contrast or color saturation looks "off" when switching between themes. Personally, I think it's ideal to keep the contrast the same between themes, but that's just a personal preference.

ImUrX:
If I add custom themes support and a share feature, would everyone be happy?

NWBx01:
A share feature would be nice. Themes are easy enough to make, but having to build the GUI to add them is a bit of a hurdle for most people.

ImUrX:
i need to make a file format, there is no nice way of "loading css files"

TheDevMinerTV:
There once was a pretty easy to parse format called "JSON". I think you should use it.

ImUrX:
I think im going to use yml... /jk

TheDevMinerTV:
Here's a list of reason why YAML is terrible: https://noyaml.com/

Mystium-YT:
Can you add support to add custom css themes?

Mystium-YT:
Just like import a css file of the colours, no like gui or anything difficult

PR #675, opened 2023-04-16
Improve vrchat osctrackers text

Louka3000:


ImUrX (approved):

gui/public/i18n/en/translation.ftl
```diff
@@ -391,7 +392,7 @@ settings-osc-vrchat = VRChat OSC Trackers
 # This cares about multilines
 settings-osc-vrchat-description =
     Change VRChat-specific settings to receive HMD data and send
-    trackers data for FBT (works on Quest standalone).
+    tracker data for FBT without SteamVR (Quest standalone).
```
ButterscotchV, line 395:
```suggestion
    tracker data for FBT without SteamVR (ex. Quest standalone).
```

ButterscotchV (changes requested):
Looks good, but I think it could be specified that you don't *need* to be using Quest Standalone.

ButterscotchV (approved):

PR #682, opened 2023-04-21
Log yaw drift and reset all trackers

Louka3000:
This logs tracker yaw difference between every full or yaw reset in the format `[INFO] [HumanPoseManager] x seconds since last reset. Tracker yaw drifts: tracker mac address, x deg (x deg/min) | etc `

Example with 3 BNO085: `[INFO] [HumanPoseManager] 1162 seconds since last reset. Tracker yaw drifts: udp://F4:CF:A2:4C:B5:FB/0, 0.5888 deg (0.0304 deg/min) | udp://E8:DB:84:98:84:C2/0, 0.3599 deg (0.0186 deg/min) | udp://98:CD:AC:31:AE:B9/0, 0.5923 deg (0.0306 deg/min) | `

Also reset every tracker to prevent the need to use logic where it doesn't belong. This is good anyways for visualizing trackers in GUI and should've been done a while ago.

Needs https://github.com/SlimeVR/SlimeVR-Server/pull/647

### DIFF
--- a/server/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.java
+++ b/server/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.java
@@ -18,6 +18,7 @@ import io.eiren.util.collections.FastList;
 import io.eiren.util.logging.LogManager;
 import io.github.axisangles.ktmath.Quaternion;
 import io.github.axisangles.ktmath.Vector3;
+import org.apache.commons.math3.util.Precision;
 
 import java.util.List;
 import java.util.Map;
@@ -685,6 +686,10 @@ public class HumanPoseManager {
 				tracker.getNeedsReset()
 					&& tracker.getResetsHandler().getLastResetQuaternion() != null
 			) {
+				if (!trackersDriftText.isEmpty()) {
+					trackersDriftText.append(" | ");
+				}
+
 				// Get the difference between last reset and now
 				Quaternion difference = tracker
 					.getRotation()
@@ -699,13 +704,9 @@ public class HumanPoseManager {
 				// Fix for polarity or something
 				if (trackerDriftAngle > 180)
 					trackerDriftAngle = Math.abs(trackerDriftAngle - 360);
-				// Round it to 4 decimal places
-				trackerDriftAngle = Math.round(trackerDriftAngle * 10000f) / 10000f;
 
 				// Calculate drift per minute
 				float driftPerMin = trackerDriftAngle / (timeSinceLastReset / 60f);
-				// Round it to 4 decimal places
-				driftPerMin = Math.round(driftPerMin * 10000f) / 10000f;
 
 				trackersDriftText.append(tracker.getName());
 				TrackerPosition trackerPosition = tracker.getTrackerPosition();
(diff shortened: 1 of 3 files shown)


ImUrX:
Ready for review and merge!

server/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.java
```diff
@@ -653,10 +665,78 @@ public void resetTrackersYaw(String resetSourceName) {
...
+					.abs(
+						FastMath.atan2(difference.getY(), difference.getW())
+							* 2
+							* FastMath.RAD_TO_DEG
+					);
+				// Fix for polarity or something
+				if (trackerDriftAngle > 180)
+					trackerDriftAngle = Math.abs(trackerDriftAngle - 360);
+				// Round it to 4 decimal places
+				trackerDriftAngle = Math.round(trackerDriftAngle * 10000f) / 10000f;
```
ButterscotchV, line 703: Isn't there a utility for this functionality?

server/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.java
```diff
@@ -1380,33 +1380,34 @@ public boolean isTrackingRightArmFromController() {
...
+				trackerPreUpdate(this.rightLowerLegTracker),
+				trackerPreUpdate(this.rightFootTracker),
+				trackerPreUpdate(this.leftLowerArmTracker),
+				trackerPreUpdate(this.rightLowerArmTracker),
+				trackerPreUpdate(this.leftUpperArmTracker),
+				trackerPreUpdate(this.rightUpperArmTracker),
+				trackerPreUpdate(this.leftHandTracker),
+				trackerPreUpdate(this.rightHandTracker),
+				trackerPreUpdate(this.leftShoulderTracker),
+				trackerPreUpdate(this.rightShoulderTracker)
```
ButterscotchV, line 1403: Worth noting, `trackerPreUpdate` is totally useless now and can be completely removed. Doesn't need to be done in this PR, but could be convenient since it's already being modified.

server/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.java
```diff
@@ -653,10 +665,78 @@ public void resetTrackersYaw(String resetSourceName) {
...
+				TrackerPosition trackerPosition = tracker.getTrackerPosition();
+				if (trackerPosition != null)
+					trackersDriftText.append(" (").append(trackerPosition.name()).append(")");
+
+				trackersDriftText
+					.append(", ")
+					.append(trackerDriftAngle)
+					.append(" deg (")
+					.append(driftPerMin)
+					.append(" deg/min) | ");
```
ButterscotchV, line 720: The trailing `|` could be prevented by checking the StringBuilder for any content at the start of the loop, and only appending it if there is content.

server/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.java
ImUrX: like to remove decimal places?

Eirenliel (approved):

PR #684, opened 2023-04-22
Make AutoBone more modular and readable

ButterscotchV:
This PR is done in preparation for further work on AutoBone, but just generally improves the structuring.

- Makes the main AutoBone adjustment method take a config as an input, so each run can choose its own config separate of the default config. This will facilitate custom AutoBone runs through the GUI.
- Splits the AutoBone loop into multiple different, one for each level of iteration to reduce nesting and huge chunks of hard to understand code.
- Moves most state variables to `AutoBoneTrainingStep`.
- Reduces height bias in the AutoBone bone adjustment order, as the `currentHmdHeight` was being adjusted based on offset order.
- Rename `AutoBoneTrainingStep` to `AutoBoneStep`.
- Remove `HumanPoseManager` thingy from `AutoBone` (why was that still there?).
- Fixes a small bug in the order of `targetHeight` initialization stuff that caused it to not be set properly in `AutoBoneStep`. This should fix the functionality of height error.


Louka3000 (approved):

ImUrX:
Ready for review & merge!

Eirenliel (approved):

Eirenliel:
Was this beta'd? I don't want to put this in RC if it wasn't beta'd with more ppl

ButterscotchV:
No, I've been somewhat hesitant to make a beta for it, but I could later if you want

ImUrX (approved):

ButterscotchV:
Smol update, beta has been running at https://discord.com/channels/817184208525983775/1122631706269192212 and has generally received either good or neutral feedback. I also did thorough testing myself and found the values were slightly different- It turns out I had actually unintentionally also fixed another bug with having a minus on the wrong line... But further testing, I confirmed the behaviour is retained and results are improved by the aformentioned bug fix and bias reduction.

Issue #653, opened 2023-04-05
[Feature Request] Notify User That SlimeVR Addon is Disabled in SteamVR


NWBx01:
Oftentimes when SteamVR crashes it will disable addons, such as SlimeVR. For a less experienced user this can be rather confusing as to why their trackers would suddenly stop working.

Ideally, then, the SlimeVR Server should notify the user that the SlimeVR addon is not currently enabled in SteamVR.

Eirenliel:
Good feature request. We should add it to the list of warnings on the main screen. It should have an option disable the warning if the user doesn't use SteamVR.

Issue #655, opened 2023-04-05
[GUI] Home page improvements


Eirenliel:
Bunch of improvements for the home page of the server.

* [ ] Settings section in the sidebar should always be displayed, the title "Developer Mode" should be removed. Unless in Developer/Advanced mode, only following settings should be visible: "More Info". Other options might be added there in the future.
* [x] New section for warnings should be added between the buttons and settings section in the sidebar. It will display errors and warnings that user should address. The section is not visible if there is no content in it. Things that will be displayed there, maybe more in the future:
  * [x] SlimeVR Addon is Disabled in SteamVR https://github.com/SlimeVR/SlimeVR-Server/issues/653
  * [x] There are trackers that require reset https://github.com/SlimeVR/SlimeVR-Server/issues/623#issuecomment-1481870719
  * [x] There are trackers in the error state
  * [x] There is more than one tracker assigned to the same body part
* [ ] Extension trackers should be grouped together with the main trackers. Need to come up with a good way to accomodate both modes (normal and more info), and joycons. Assume that there might be any amount of extensions in the future. #453 
* [ ] Trackers should be sorted top to bottom of the skeleton, left to right (for example, first left thigh, then right thigh), except extensions should be grouped together.
* [x] Trackers should show warning icons if there are warnings in them, like the tracker requires reset (https://github.com/SlimeVR/SlimeVR-Server/issues/623#issuecomment-1481870719) or is in the error state

Eirenliel:
This will be reworked with new goals.

PR #689, opened 2023-05-09
Status system

ImUrX:
This will let the server notify of problems that in the future might be autofixable, but for now just will pass to the GUI and show them as a problem requiring fixing.

Requires #647 and SlimeVR/SolarXR-Protocol#110


ImUrX:
Fixes #653 and partially #655

server/src/main/java/dev/slimevr/status/StatusSystem.kt
```diff
@@ -0,0 +1,66 @@
...
+		listeners.add(listener)
+	}
+
+	fun removeListener(listener: StatusListener) {
+		listeners.remove(listener)
+	}
+
+	fun getStatuses(): Array<StatusMessageT> = statuses.map { (id, message) ->
+		val status = StatusMessageT()
+		status.id = id.toUInt().toLong()
```
ButterscotchV, line 26: I don't quite get this lol
```suggestion
		status.id = id.toLong()
```

server/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -52,13 +58,44 @@ class Tracker @JvmOverloads constructor(
...
+			if (field == value) return
+			field = value
+
+			if (!isInternal) {
+				// Check if a full reset is needed
+				if (needsReset && value != null && lastResetStatus == 0u && status.sendData) {
+					reportRequireReset()
+				} else if (value == null && lastResetStatus != 0u) {
+					vrServer.statusSystem.removeStatus(lastResetStatus)
+					lastResetStatus = 0u
```
ButterscotchV, line 98: This seems like an incredibly large amount of logic to have in a property setter...

server/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -84,6 +121,53 @@ class Tracker @JvmOverloads constructor(
...
+				}
+				trackerNum = tempTrackerNum
+			}
+		}
+		val status = StatusDataUnion().apply {
+			type = StatusData.StatusTrackerError
+			value = statusMsg
+		}
+		lastErrorStatus = vrServer.statusSystem.addStatus(status, true)
+	}
```
ButterscotchV, line 169: This feels like a little awkward of a place to have these?

server/src/main/java/dev/slimevr/status/StatusSystem.kt
ImUrX: ``-52.toLong()`` would just be ``-52``, ``-52.toUInt()`` would convert it into ``4294967244``

server/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
ImUrX: any suggestions?

server/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
ImUrX: the problem is the variable used to hold the latest status reported by the own tracker

server/src/main/java/dev/slimevr/status/StatusSystem.kt
Louka3000: Why would you not want it to go in the negatives? For solarxr?

ImUrX: We are in Kotlin, no purpose of being forced to use `int`s. But I'm still kind of forced to >~>

ButterscotchV (approved):

Eirenliel (approved):

ImUrX:
Ready for merge ^-^

PR #692, opened 2023-05-16
Stop unconditionally hogging serial ports

evg-zhabotinsky:
SlimeVR server was hogging serial ports unconditionally, even when neither provisioning nor serial console were active and serial connection popup was disabled. (Had to manually open serial console and close it to unhog the port.)

Fixed that.

`ProvisioningHandler` registers its serial port discovery handler in constructor and never deregisters it.
The handler was unconditionally opening the ports.
Made it bail out if provisioning isn't running, same way as the provisioning timer handler does.

Verified that provisioning still works as intended, both from the serial connection popup and the setup wizard, and that serial console works normally, regardless if it was opened from popup and from settings.
If popup is disabled or dismissed, and also until user clicks wifi/console button in it, the serial now stays free (checked with vscode serial monitor).


Eirenliel (approved):
Good fix. Will need testing before merging.

ButterscotchV (approved):

Louka3000 (approved):
LGTM, can test myself when I get slimes if needed

evg-zhabotinsky:
Out of curiosity, what kind of testing is expected? As mentioned, I checked everything serial-related I could think of. Also did run wednesday's dance event with that change (dev build), didn't notice any differences.

Eirenliel:
We need to test it inernally that official slimes can provision wifi properly on our PCs.

Louka3000:
This works fine on my devkit slimes!

ImUrX:
should be merged ^^

PR #701, opened 2023-05-26
Require the use of run command for running server

ImUrX:
This forces the use of a command line to run the server, so people who double click it don't accidentally run the server in the background.

### DIFF
--- a/server/src/main/java/dev/slimevr/Main.kt
+++ b/server/src/main/java/dev/slimevr/Main.kt
@@ -7,7 +7,6 @@ import org.apache.commons.cli.CommandLine
 import org.apache.commons.cli.CommandLineParser
 import org.apache.commons.cli.DefaultParser
 import org.apache.commons.cli.HelpFormatter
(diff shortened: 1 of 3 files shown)


server/src/main/java/dev/slimevr/Main.kt
```diff
@@ -49,6 +47,16 @@ fun main(args: Array<String>) {
 		println("SlimeVR Server $VERSION")
 		exitProcess(0)
 	}
+
+	if (cmd.args.isEmpty()) {
+		System.err.println("No command specified, expected 'run'")
+		exitProcess(1)
+	}
+	if (cmd.args[0] != "run") {
```
ButterscotchV, line 55: This could be case-insensitive
```suggestion
	if (!cmd.args[0].equalsIgnoreCase("run")) {
```

ButterscotchV (approved):
This sounds like a great idea, thankyu

Eirenliel (approved):

PR #702, opened 2023-05-27
Fix legtweaks knees condition

Louka3000:
Caused by trackers-rewrite because of bad convertion:
Using SlimeVR for hip tracking without leg trackers will still have the effects of floorclip on the hip tracker's position.

issue (kinda): https://github.com/SlimeVR/SlimeVR-Server/pull/647/files#r1178305303


ButterscotchV (approved):

Eirenliel (approved):

Stermere (commented):
nice catch

PR #716, opened 2023-06-11
Add upper chest tracking point

Louka3000:
Eiren gave me 2 extension spine straps, so I'm gonna use them. 
This adds a basic upper chest tracking point and shouldn't affect normal usage (i.e, upperChest is hidden by default, and tracking is the same with just chest but no upperChest). 
This thus allows a bit more precision for people that have the extra trackers.

Needs https://github.com/SlimeVR/SolarXR-Protocol/pull/113


Eirenliel:
firL

ImUrX (changes requested):
add this on the manual proportions ratio group

Louka3000:
I wrote it down and forgor

ButterscotchV (approved):

server/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.java
```diff
@@ -144,7 +144,7 @@ private void initializeComputedHumanPoseTracker() {
 					VRServer.getNextLocalTrackerId(),
 					"human://CHEST",
 					"Computed chest",
-					TrackerPosition.CHEST,
+					TrackerPosition.UPPER_CHEST,
```
ImUrX, line 147: is this fine?

ImUrX (changes requested):

server/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.java
Louka3000: It really doesn't matter if it's CHEST or UPPER_CHEST as long as it's changed in the rest of the code (doesn't affect config or solarxr I mean). 
So yes it's fine, I did change it in other parts of the code already.

ImUrX (approved):

Eirenliel (approved):

PR #717, opened 2023-06-11
Fix allowDriftCompensation setting to false on mounting or assign

Louka3000:
~~For once a bug that wasn't my fault (_I think_)~~

Update: fml


ButterscotchV (approved):
If null, shouldn't it be `true` as that's the default value? Would work fine either way, idk if it really matters.

ImUrX (approved):

Louka3000:
> If null, shouldn't it be `true` as that's the default value? Would work fine either way, idk if it really matters.

Yeah I thought about it, idk… ig I’ll still do it after all since you also think that

Eirenliel (approved):

PR #719, opened 2023-06-13
Send any computed tracker via VMC for props

Louka3000:
Sends any computed tracker via VMC, like Vive trackers.  
VMC trackers seem to only be used for props.


Louka3000:
This now makes it so the SteamVR trackers use their SteamVR ID as the trackerNum and use the VRServer's unique ID as the tracker ID (as they should).
It also uses their serial for their name, effectively fixing the configs for setup having 2+ trackers of the same kind (e.g. vive_tracker, not 2 different (left and right) controllers)

before: 
image

after: 
image

(ignore role that changed)

server/core/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -198,7 +198,7 @@ protected void trackerAddedReceived(TrackerAdded trackerAdded) {
 			remoteTrackersBySerial.put(tracker.getName(), tracker);
 		}
 		synchronized (remoteTrackersByTrackerId) {
-			remoteTrackersByTrackerId.put(tracker.getId(), tracker);
```
ButterscotchV, line 201: I think this is generally supposed to be persistent, but I don't think we are entirely able to do that right now. Why not use ID though?

server/core/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
```diff
@@ -198,7 +198,7 @@ protected void trackerAddedReceived(TrackerAdded trackerAdded) {
...
+			remoteTrackersByTrackerId.put(tracker.getTrackerNum(), tracker);
```
ButterscotchV, line 201: I think this is generally supposed to be persistent, but I don't think we are entirely able to do that right now. Why not use ID though?

server/core/src/main/java/dev/slimevr/osc/VMCHandler.java
```diff
@@ -453,31 +452,37 @@ && isRightArmUnityBone(bone))
...
-						address = "/VMC/Ext/Tra/Pos";
-					}
-					oscBundle
-						.addPacket(
-							new OSCMessage(
-								address,
-								oscArgs.clone()
-							)
+				for (Tracker tracker : computedTrackers) {
+					if (tracker.getStatus() != TrackerStatus.DISCONNECTED) {
```
ButterscotchV, line 456: Should the status also be checked for `TrackerStatus.ERROR` here?

Louka3000: sure

server/core/src/main/java/dev/slimevr/bridge/ProtobufBridge.java
Louka3000: Because that uses the SteamVR tracker ID, which I changed to be the trackerNum 
image

ButterscotchV (approved):

Eirenliel (approved):

PR #721, opened 2023-06-14
Tracking pause fixes

ButterscotchV:
- Retains the state of tracking pause when a tracker is disconnected
- Retains the pose of the HumanSkeleton when tracking is paused so it doesn't revert to t-pose
- Resets the legtweaks buffer (`legTweaks.resetBuffer();`) when tracking is unpaused


ButterscotchV:
Personally tested in VR and it seems fine 😸

Louka3000 (approved):

Eirenliel (approved):

PR #731, opened 2023-06-15
Finally fix small window

ImUrX:


gui/src-tauri/src/state.rs
```diff
@@ -0,0 +1,109 @@
...
+		Ok(())
+	}
+}
+
+pub trait WindowBuilderExt {
+	fn restore_state(self, state: &WindowState) -> Self;
+}
+
+impl WindowBuilderExt for tauri::WindowBuilder<'_> {
+	fn restore_state(self, state: &WindowState) -> Self {
```
TheButlah, line 86: I couldn't find where this was used. Is it dead code? If so, delete it

gui/src/components/TopBar.tsx
```diff
@@ -194,7 +195,10 @@ export function TopBar({
               </div>
               <div
                 className="flex items-center justify-center hover:bg-background-60 rounded-full w-7 h-7"
-                onClick={() => appWindow.close()}
+                onClick={async () => {
+                  await invoke('update_window_state');
+                  await appWindow.close();
```
TheButlah, line 200: why do you await the close here? Does `.close()` return when the app window gets exited?

gui/src-tauri/src/state.rs
```diff
@@ -0,0 +1,110 @@
...
+		self.inner_size(state.width, state.height)
+			// .maximized(state.maximized)
+	}
+}
+
+trait MonitorExt {
+	fn contains(&self, position: PhysicalPosition<i32>) -> bool;
+}
+
+const ERROR: i32 = 16;
```
TheButlah, line 99: Naming this ERROR is confusing as it is entirely unrelated to error types or error messages. I'm not even sure what the purpose of it is

gui/src-tauri/src/main.rs
```diff
@@ -20,8 +21,19 @@ use crate::util::{
...
+mod state;
 mod util;
 
+#[tauri::command]
+fn update_window_state(
+	window: tauri::Window,
+	state: tauri::State<Mutex<state::WindowState>>,
+) -> Result<(), String> {
+	let mut lock = state.lock().unwrap();
+	lock.update_state(&window).map_err(|err| err.to_string())?;
```
TheButlah, line 33: Once the other errors use `color_eyre::Result`, you can map the `eyre::Report` using `format!("{:?}", err)`

https://docs.rs/eyre/0.6.8/eyre/struct.Report.html#display-representations

TheButlah (changes requested):

gui/src-tauri/Cargo.toml
```diff
@@ -23,27 +23,27 @@ custom-protocol = ["tauri/custom-protocol"]
...
 tempfile = "3"
 which = "4.3"
 glob = "0.3"
-open = "3"
-shadow-rs = { version = "0.21", default-features = false }
+open = "4"
+shadow-rs = { version = "0.22", default-features = false }
 const_format = "0.2.30"
 cfg-if = "1"
+anyhow = "1.0"
```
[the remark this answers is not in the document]
ImUrX: i was going to use eyre, but its less maintained than anyhow

TheButlah:
Also, your approach here of replacing the tauri plugin with your own logic doesn't seem super fun to me. What is here is already very similar to the tauri plugin - can't you just fork the plugin and fix the issue there? Its a very small amount of code there, should be very approachable to find the bug and fix it upstream instead of making our own thing.

gui/src-tauri/Cargo.toml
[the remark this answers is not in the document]
TheButlah: Its definitely not less maintained, I have used it in every project for years

ImUrX:
I dont honestly want to discuss this on here, but it seems we are the only ones with this issue even though I have asked multiple times, and them using a binary format that is based on the rust version and is unreadable for users is not something I think they want to change.

Currently I'm just pushing and debugging with someone's testing on windows because I dont have a machine setup with Windows nearby that I can use for testing this, so I just give them the artifact to test.

gui/src-tauri/src/state.rs
```diff
@@ -0,0 +1,109 @@
...
+
+		if !ignore_maximized && !maximized && self.maximized {
+			window.maximize()?;
+		}
+
+		Ok(())
+	}
+}
+
+pub trait WindowBuilderExt {
```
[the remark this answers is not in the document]
ImUrX: wat

gui/src/components/TopBar.tsx
ImUrX: really weird nitpick, i dont know

TheButlah:
hmm, unfortunate. If you think its best to maintain it ourselves, go for it

gui/src/components/TopBar.tsx
TheButlah: my comment was a question not a nitpick :) If it is waiting for close, then this doensn't make much sense because it just creates a lot of promises that will never complete until the application exits. But maybe I'm misunderstanding something, which is why I asked

gui/src-tauri/src/state.rs
ImUrX: renamed to ``ABSOLUTE_ERROR`` and it's already next to the `impl`

gui/src-tauri/src/state.rs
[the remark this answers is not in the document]
TheButlah: Did you need some additional clarification? Your response isn't very descriptive

gui/src-tauri/src/state.rs
TheButlah: IK, I mean why are you adding this width at all?

ImUrX: for some reason, when maximizing windows on Windows, the `x` and `y` is `-8` on both

gui/src-tauri/src/state.rs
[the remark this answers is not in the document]
ImUrX: like I mean, I don't know how to reply qwq. Using traits like this is a very nice way of maintaining builder methods or complicated code reusable

gui/src-tauri/Cargo.toml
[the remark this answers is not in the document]
ImUrX: seeing both github repositories/issue trackers proves otherwise, but i dunno. I will pass to `eyre` if you think its required

gui/src-tauri/src/state.rs
ImUrX: its for allowing overflow of the window for just 16 pixels

gui/src-tauri/Cargo.toml
```diff
@@ -39,10 +39,10 @@ tempfile = "3"
 which = "4.3"
 glob = "0.3"
 open = "4"
-shadow-rs = { version = "0.22", default-features = false }
+shadow-rs = { version = "0.23", default-features = false }
 const_format = "0.2.30"
 cfg-if = "1"
-anyhow = "1.0"
+color-eyre = { version = "0.6", default-features = false }
```
TheButlah, line 45:
```suggestion
color-eyre = "0.6"
```
I would use the defaults

gui/src-tauri/src/main.rs
```diff
@@ -122,24 +127,28 @@ fn main() {
 	let build_result = tauri::Builder::default()
 		.invoke_handler(tauri::generate_handler![update_window_state])
 		.setup(move |app| {
-			let window_state = if let Some(window_state) = state::WindowState::open_state(
-				app.path_resolver().app_config_dir().unwrap(),
-			) {
+			let window_state = if let Some(window_state) =
+				WindowState::open_state(app.path_resolver().app_config_dir().unwrap())
```
TheButlah, line 131: why not `unwrap_or_default()`?

gui/src-tauri/src/state.rs
```diff
@@ -78,33 +91,20 @@ impl WindowState {
...
-		if !state.is_old() {
-			return self;
-		}
-		self.inner_size(state.width, state.height)
-			// .maximized(state.maximized)
-	}
-}
-
-trait MonitorExt {
+pub trait MonitorExt {
```
TheButlah, line 94: If you are confident this is easier to read for a rust beginner you could keep the Ext trait, but I still think its easier for them to not use this pattern and just have a regular function

ImUrX: thank you ><

Eirenliel (approved):

Issue #725, opened 2023-06-14
Rename "save proportions"


Vyolex:
Save proportions is a bit of an ambuigous term, it is not clear that this will export the proportions to a file. When going through it we had the first impression we should press it after manual/automatic proportions to apply them.

Renaming this button to "Export proportions" or similar might help clarify this.

ImUrX:
I agree with this change

PR #732, opened 2023-06-16
Rename save proportions to export proportions

Vyolex:
Change english text as discussed in #725.

This feels like such a useless PR 😢


Eirenliel (approved):

Eirenliel:
No self-deprecation allowed in PRs :nya_gun: If it made people confused, then it's not useless, and in fact very important.

Issue #723, opened 2023-06-14
Add warning to mounting calibration (in onboarding)


Vyolex:
Mounting calibration is currently too "difficult" and finicky for new users to do reliably. While a good calibration can improve tracking a fair bit improper calibration can also severly diminish the tracking quality. 

For now we want to add following warning in the onboarding and recommend manual mounting as default.
> Mounting calibration can lead to better tracking quality, but improper calibration can make it worse. It requires some time to figure out and check if it's right for you and the way you put on your trackers. We're working on making it better.

PR #734, opened 2023-06-16
Updates the recommended mounting method to manual

Vyolex:
Swaps the order of manual/automatic mounting, along with changing which method is recommended on the mounting selection screen until a better visualisation method is available.

Discussed in #723, we did decide to drop the warning in the GUI.

### DIFF
--- a/gui/public/i18n/en/translation.ftl
+++ b/gui/public/i18n/en/translation.ftl
@@ -649,7 +649,7 @@ onboarding-choose_mounting = What mounting calibration method to use?
 onboarding-choose_mounting-description = Mounting orientation corrects for the placement of trackers on your body.
 onboarding-choose_mounting-auto_mounting = Automatic mounting
 # Italized text
-onboarding-choose_mounting-auto_mounting-subtitle = If you know what you are doing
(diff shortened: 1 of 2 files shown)


ImUrX (approved):

gui/public/i18n/en/translation.ftl
```diff
@@ -649,11 +649,11 @@ onboarding-choose_mounting = What mounting calibration method to use?
 onboarding-choose_mounting-description = Mounting orientation corrects for the placement of trackers on your body.
 onboarding-choose_mounting-auto_mounting = Automatic mounting
 # Italized text
-onboarding-choose_mounting-auto_mounting-subtitle = Recommended
+onboarding-choose_mounting-auto_mounting-subtitle = If you know what you are doing
```
Louka3000, line 652:
```suggestion
onboarding-choose_mounting-auto_mounting-subtitle = Experimental
```
This fits better and makes more sense imo

Louka3000 (changes requested):

Eirenliel (approved):

PR #737, opened 2023-06-20
Acceleration alignment fix

Stermere:
Before the accelerations were only aligned if the trackers were powered on in the same orientation, with this they appear to be aligned with each other on all axis. At a very minimum all trackers now read -9.8 on the Y axis when in free fall which is much better than before, the other axes seem to be aligned but without a constant acceleration available in that direction it is harder to say for sure.


server/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
```diff
@@ -293,7 +293,8 @@ class TrackersUDPServer(private val port: Int, name: String, private val tracker
 			is UDPPacket4Acceleration -> {
 				tracker = connection?.getTracker(packet.sensorId)
 				if (tracker == null) return
-				val acceleration = tracker.getRotation().sandwich(packet.acceleration)
+				val accelRot = tracker.resetsHandler.getMountingAdjustedRotationFrom(tracker.getRawRotation())
+				val acceleration = accelRot.sandwich(packet.acceleration)
```
Louka3000, line 297: You would want to do that directly in TrackerResetsHandler, as trackers with accel can be created from other sources

Louka3000 (changes requested):

server/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
Stermere: good point let me know if the changes I made are what you had in mind

Louka3000: This approach still works obviously, but we risk someone forgetting to call that method when setting accel somewhere else, and it also separates the logic from where I think it should be.  

What I had in mind was doing the same thing as rotation (in Tracker.kt, make accel private and have a custom get method that gets the adjustedAccelerationFromRotation if tracker needsReset).

Stermere: that makes more sense, the naming of needs reset threw me off a little I thought it indicated something else, thanks for elaborating

Louka3000 (approved):

Eirenliel (approved):

server/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
```diff
@@ -96,6 +96,16 @@ class TrackerResetsHandler(val tracker: Tracker) {
...
 	}
 
+	/**
+	 * Takes a rotation and adjusts it to mounting
+	 */
+	fun getMountingAdjustedRotationFrom(rotation: Quaternion): Quaternion {
+		if (mountRotFix == Quaternion.IDENTITY) {
+			return rotation * mountingOrientation
+		}
+		return rotation * mountRotFix
```
ButterscotchV, line 106: Mounting orientation needs to be applied before mounting fix
```suggestion
		if (mountRotFix == Quaternion.IDENTITY) {
			return rotation * mountingOrientation
		}
		return rotation * mountingOrientation * mountRotFix
```

ButterscotchV (changes requested):
I don't believe the math will work correctly with mounting reset applied like this. Also tbh the if there is a bit of a micro optimization in that case, but it's fine either way.

server/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
ImUrX: i would prefer just removing the if

ButterscotchV: Here's if we don't want the if
```suggestion
		return rotation * mountingOrientation * mountRotFix
```

Stermere: I tested this and it seemed to be the same either way. I believe that that was because I have default mounting on all my trackers so this is very likely right, I will test again soon anyways

Stermere: tested it best I can tell it is working

ButterscotchV (approved):

ImUrX (approved):

PR #743, opened 2023-06-26
Improve defaults SkeletonValues & add ratios in gui

Louka3000:
Changes default values a bit and adds them to GUI under dev mode


Louka3000:
Ready for review ig

server/core/src/main/java/dev/slimevr/protocol/rpc/settings/RPCSettingsHandler.java
```diff
@@ -298,44 +298,36 @@ public void onChangeSettingsRequest(GenericConnection conn, RpcMessageHeader mes
...
+				hpm
+					.setValue(
+						SkeletonConfigValues.HIP_LEGS_AVERAGING,
+						ratios.interpHipLegs()
+					);
+				hpm
+					.setValue(
+						SkeletonConfigValues.KNEE_TRACKER_ANKLE_AVERAGING,
+						ratios.interpKneeTrackerAnkle()
+					);
```
ButterscotchV, line 330: wha? why are you removing that if it can be null

Louka3000: It’s not null, it’s just a check if == 0

ButterscotchV: It's a check if the pointer is 0, not the value. The "null" part is in the flatbuffers.

Louka3000: It returns false if the value is 0, I tested it.
I tested changing other settings and it worked fine. If we want to keep the ‘has’ check, we’d have to hardcode ‘0’ to another value (like -1 or something) both in GUI and in server… Not convinced this is needed tho

ImUrX: pls dont remove this, it returns if the pointer doesnt exist in flatbuffer :p
```java
  public boolean hasImputeWaistFromChestHip() { return 0 != __offset(4); }
  public float imputeWaistFromChestHip() { int o = __offset(4); return o != 0 ? bb.getFloat(o + bb_pos) : 0f; }
```

Louka3000: What do you suggest we do then? use -1 to represent 0?

ImUrX: just keep doing the has thing, if we used kotlin flatbuffers it would be simpler >~<

Louka3000: I cannot keep doing the has thing, I **need** to know when the value is 0.

ImUrX: imputeWaistFromChestHip will return 0 if its value is 0...

Louka3000: Which is why I originally removed the has check, why are you telling me that?

Louka3000: Fixed in https://github.com/SlimeVR/SlimeVR-Server/pull/743/commits/6ef30d7f6af4dfa28bd9693eec1398690e6fc977

ImUrX (approved):

ButterscotchV (approved):

Eirenliel (approved):

PR #744, opened 2023-06-27
Fix Dropdown not opening

ImUrX:
Currently the `Dropdown` grabbed the first existing `.dropdown` which might not happen, I fixed that by using `closest()`.

I also fixed scrolling, `scrollIntoView` doesnt work well on webkit, so now we using `scroll` which is a little more manual qwq. But it looks good and also added scroll in the `Dropdown` (that's how i discovered the webkit bug)


TheDevMinerTV (approved):
LGTM

I don't like that there's a custom variant but whatever

loucass003 (approved):

PR #747, opened 2023-06-29
tps upstream

Philip6294:


TheDevMinerTV:
Oop? It seems like you have opened an empty pull request (well, one without any file changes) 😄

I think you meant to change the target to your repository and merge *from* SlimeVR ^^

Philip6294:
@TheDevMinerTV yup that was an accident ^^
(also the second time I managed to do this)

PR #749, opened 2023-07-01
Mocap mode

Stermere:
Adds the ability for some movements (movements where one foot is on the ground at all times) to be represented in 3d without the need for a headset of other 6dof device. There is positional drift, it's generally exaggerated if doing faster movements or when the tracker's connection is not very stable, as long as neither of those things happen it is not a huge issue

~~Needs https://github.com/SlimeVR/SolarXR-Protocol/pull/115 and https://github.com/SlimeVR/SlimeVR-Server/pull/737 merged first~~ Done


server/src/main/java/dev/slimevr/tracking/processor/skeleton/Localizer.kt
```diff
@@ -0,0 +1,425 @@
...
+		// get the buffer that occurred VELOCITY_SAMPLE_RATE ago in time
+		while (buf.timeOfFrame > timeEnd && buf.parent !== null) {
+			buf = buf.parent
+		}
+
+		val comPosEnd: Vector3 = buf.centerOfMass
+		timeEnd = buf.timeOfFrame
+
+		// calculate the velocity
+		comVelocity = comPosStart.minus(comPosEnd).div((timeStart - timeEnd) / LegTweakBuffer.NS_CONVERT)
```
ButterscotchV, line 305: Shouldn't this be reversed? I feel like I'm missing something here lol, it looks like it'd technically work anyways though because of negation
```suggestion
		comVelocity = comPosEnd.minus(comPosStart).div((timeEnd - timeStart) / LegTweakBuffer.NS_CONVERT)
```

ButterscotchV (approved):

server/src/main/java/dev/slimevr/tracking/processor/skeleton/Localizer.kt
Stermere: I got super lucky and it evaluates to the same thing, It should be flipped though for readability

ButterscotchV (approved):

Louka3000:
solarxr submodule has merge conflict

Louka3000 (approved):

ImUrX:
this can be merged

PR #750, opened 2023-07-02
Drift compensation tracker and HMD requirement

Louka3000:
Doesn't do drift compensation if non-imu head tracker is present


Louka3000:
Not enough feedback, so I'm keeping the defaults as before for now (off), so this can be merged.

ButterscotchV (approved):

ImUrX:
fix conflict

Louka3000:
done

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerUtils.kt
```diff
@@ -55,6 +55,24 @@ object TrackerUtils {
...
+	@JvmStatic
+	fun getNonInternalNonImuTrackerForBodyPosition(
+		allTrackers: List<Tracker>,
+		position: TrackerPosition,
+	): Tracker? = allTrackers.firstOrNull {
+		it.trackerPosition === position &&
+			!it.isImu() &&
+			!it.isInternal &&
+			it.status != TrackerStatus.DISCONNECTED &&
+			it.status != TrackerStatus.ERROR
```
ImUrX, line 73: maybe `sendData` or `reset` would help here?

Louka3000: :+1:

ImUrX (approved):

PR #752, opened 2023-07-02
Add rust file logging

ImUrX:
Now we have rust logging into a file, and it includes the logs from webview!


gui/src/hooks/logging.ts
TheDevMinerTV, line 1: IMO this shouldn't be here. `hooks` is meant for React hooks.

TheDevMinerTV (changes requested):

gui/src/hooks/logging.ts
ImUrX: fine now?

gui/src/hooks/config.ts
```diff
@@ -94,7 +95,7 @@ export function useConfigProvider(): ConfigContext {
         setLoading(false);
         return loadedConfig;
       } catch (e) {
-        console.log(e);
+        log(e);
```
TheDevMinerTV, line 98: shouldn't this be `error`?

gui/src/hooks/app.ts
```diff
@@ -104,7 +105,7 @@ export function useProvideAppContext(): AppContext {
         }
       }
     } catch (error) {
-      console.log(error);
+      log(error);
```
TheDevMinerTV, line 108: shouldn't this be `error`?

TheDevMinerTV (changes requested):

TheDevMinerTV (approved):

gui/src-tauri/src/main.rs
```diff
@@ -38,22 +39,53 @@ fn update_window_state(
...
+}
+
+fn main() -> Result<()> {
+	log_panics::init();
+	let hook = panic::take_hook();
 	// Make an error dialog box when panicking
-	panic::set_hook(Box::new(|panic_info| {
-		println!("{}", panic_info);
+	panic::set_hook(Box::new(move |panic_info| {
+		hook(panic_info);
```
TheButlah, line 62: What happens if panic hook aborts or does something wacky? Presumably `show_error` should happen first?

gui/src-tauri/src/main.rs
```diff
@@ -38,22 +39,53 @@ fn update_window_state(
...
+	let tauri_context = tauri::generate_context!();
 
 	// Set up loggers and global handlers
-	{
-		if std::env::var_os("RUST_LOG").is_none() {
-			std::env::set_var("RUST_LOG", "info")
-		}
-		pretty_env_logger::init();
-	}
+	let _logger = {
```
TheButlah, line 70: no need to assign to an unused var.

TheButlah (changes requested):
Also, is there no way for the typescript to keep using console.error? I don't like that this requires us to import our own error function in typescript - not all third party crates will be using that.

gui/src-tauri/src/main.rs
ImUrX: it needs to be alive, its buffering logs and writing them to a file, more info in their [docs.rs](https://docs.rs/flexi_logger/0.25.5/flexi_logger/code_examples/index.html#choose-the-write-mode)

gui/src-tauri/src/main.rs
ImUrX: `show_error` sleeps until the user closes the prompt, what do you think we should do?

ImUrX:
for overriding error handling, the closest thing we have in js is [onerror](https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event) event

gui/src-tauri/src/main.rs
TheButlah: I think we should let the panic handler block, because at least we guarantee that the user sees the popup.

gui/src-tauri/src/main.rs
```diff
@@ -38,22 +39,53 @@ fn update_window_state(
 	Ok(())
 }
 
-fn main() {
+#[tauri::command]
+fn logging(msg: String) {
+	log::info!(target: "webview", "[webview] {}", msg)
```
TheButlah, line 44: I think its an antipattern to include `[webview]` in the log text. Its generally the responsibility of the logger implementation to report the target in the logs.

TheButlah (changes requested):

gui/src-tauri/src/main.rs
ImUrX: It shows the crate name, which is `[slimevr]` :c

TheButlah (approved):

Louka3000 (approved):

loucass003 (approved):

PR #753, opened 2023-07-02
Track connection by SocketAddress instead of IP address.

kitlith:
Reconnections are detected via MAC address instead of by IP address, as well.

This file feels like it needs some cleanup after this change -- those of you that have a better idea of exactly what's going on here, feel free to clean it up. Alternatively, I might have fucked something up.

The goal of this change is to allow slimevr server to talk to trackers through a NAT-like proxy, such as sudppipe (from  <http://aluigi.altervista.org/mytoolz.htm>), or allow multiple programs that work like slimevr wrangler to run on the same machine.

In terms of testing, we need someone to test this with tracker connect/disconnect/reconnect to make sure I didn't break things, and eventually someone who can test proxying multiple trackers to slimevr server. (I might be able to pull off the latter test myself once the house cools down.)

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
@@ -12,7 +12,12 @@ import io.github.axisangles.ktmath.Quaternion.Companion.fromRotationVector
 import io.github.axisangles.ktmath.Vector3
 import org.apache.commons.lang3.ArrayUtils
 import solarxr_protocol.rpc.ResetType
-import java.net.*
+import java.net.DatagramPacket
+import java.net.DatagramSocket
+import java.net.InetSocketAddress
+import java.net.NetworkInterface
+import java.net.SocketAddress
+import java.net.SocketTimeoutException
 import java.nio.ByteBuffer
 import java.nio.ByteOrder
 import java.nio.charset.StandardCharsets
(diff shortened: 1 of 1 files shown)


ImUrX:
i think this can get merged, can you resolve conflicts?

kitlith:
Rebased. Please don't merge until someone has actually tested it, this is *not* something we want to test in production.

kitlith:
lmore377 confirmed that this both seems to work and fixes the issue that I was targeting. Un-marking as ready for review, will rebase again.

Louka3000 (approved):
lgtm

ButterscotchV (approved):
This looks alright, but it's very hard to understand what's going on, it would be of benefit to have some comments explaining some of your decisions here. Though really... Idk it might just be the whole file is way too confusing lol

ImUrX:
i find it quite easy to understand even before I ported to Kotlin :thinking:
but maybe yeah, some comments on another PR documenting the states of the udp server would be good :p

ButterscotchV:
> i find it quite easy to understand even before I ported to Kotlin :thinking:
> but maybe yeah, some comments on another PR documenting the states of the udp server would be good :p

Idk there's just a lot of things being set in different places and exchanged and whatever? Kinda don't understand why there's so much code for it

kitlith:
> > i find it quite easy to understand even before I ported to Kotlin 🤔
> > but maybe yeah, some comments on another PR documenting the states of the udp server would be good :p
> 
> Idk there's just a lot of things being set in different places and exchanged and whatever? Kinda don't understand why there's so much code for it

I agree, see my original comment about needing cleanup. I just haven't had the energy.

Eirenliel (approved):

PR #727, opened 2023-06-15
Separate the server in subprojects

ImUrX:
This PR is a combination of @ButterscotchV's efforts on adding android support + me wanting to have a good folder structure for it.

PR #770, opened 2023-07-08
Save config and logs on appdata folder

ImUrX:
I added a way to get the correct config folder and log folder depending on the OS, now SlimeVR will save them on there (ex: `%APPDATA%\dev.slimevr.SlimeVR\vrconfig.yml`).
As a migration script I made a check if the file doesn't exist on the appdata folder and it does exist next to it (`./vrconfig.yml`), it will move it to the appdata folder.
As an override method, if you have a folder named `config` next to the executable, it will keep up the old behavior :p

Blocked by #727 
Compare without all the commits from rebase: https://github.com/SlimeVR/SlimeVR-Server/compare/ImUrX:subprojects...config-dir-desktop


Louka3000 (approved):
lgtm

server/desktop/.gitignore
```diff
@@ -1 +1,2 @@
-/build
\ No newline at end of file
+/build
+config/
```
ButterscotchV, line 2: This seems fine but is a very generic word and could have consequences from ignoring generically, can we use absolute path?

server/desktop/src/main/java/dev/slimevr/desktop/Main.kt
```diff
@@ -212,3 +221,18 @@ fun provideFeederBridge(
 
 	return feederBridge
 }
+
+const val CONFIG_FILENAME = "vrconfig.yml"
+fun resolveConfig(): String {
+	// If config folder exists, then save config on relative path
+	if (Path("config/").exists()) {
```
ButterscotchV, line 228: ngl having a folder for this feels a little weird, usually a file is used as far as I've seen, but this should work

ImUrX: yeah but for windows you need to create a "type of file" and if you want to remove or change file extension you have to change a windows configuration and like its a pain in the ass qwq

server/desktop/.gitignore
ImUrX: ah sure

server/desktop/src/main/java/dev/slimevr/desktop/Main.kt
ButterscotchV: Understandable, that's alright then

ButterscotchV (approved):

Eirenliel (approved):

Issue #764, opened 2023-07-04
Use path alias instead of relative path in GUI project


marcozzxx810:
Using a relative path seems hard to be understood, For example:

https://github.com/marcozzxx810/SlimeVR-Server/blob/main/gui/src/components/onboarding/pages/ConnectTracker.tsx#L12

Proposed solution

- [ ] use `@` to indicate root path for GUI

Current
```ts
import { useLayout } from '../../../hooks/layout';
import { useOnboarding } from '../../../hooks/onboarding';
```
Later
```ts
import { useLayout } from '@/hooks/layout';
import { useOnboarding } from '@/hooks/onboarding';
```

ImUrX:
you mean `@hooks/` i guess, does the editor do this when auto-importing a new module or does it still use relative pathing? Can we enforce this through eslint?

marcozzxx810:
I mean '@/hooks' so we have '@' pointing to gui folder. It depends on your LSP but editor with intellisense should be able to pick up tsconfig


For eslint, https://www.npmjs.com/package/eslint-import-resolver-typescript  I am not so certain about can we enforce it

ImUrX:
i meant, auto-imports will prefer using the aliases or relative paths?

marcozzxx810:
vscode will prefer the alias

TheDevMinerTV:
> vscode will prefer the alias

It depends on what the user configured, though we can override this by including a `.vscode` folder with the new settings.

ImUrX:
then its fine, sure :+1:

PR #771, opened 2023-07-08
feat: use path-alias

marcozzxx810:
close: #764


gui/package.json
```diff
@@ -43,8 +43,7 @@
     "lint": "eslint --max-warnings=0 \"src/**/*.{js,jsx,ts,tsx,json}\" && prettier --check \"src/**/*.{js,jsx,ts,tsx,css,md,json}\"",
     "lint:fix": "eslint --fix \"src/**/*.{js,jsx,ts,tsx,json}\"",
     "format": "prettier --write \"src/**/*.{js,jsx,ts,tsx,css,md,json}\"",
-    "preview-vite": "vite preview",
-    "javaversion-build": "cd src-tauri/src/ && javac JavaVersion.java && jar cvfe JavaVersion.jar JavaVersion JavaVersion.class"
```
ImUrX, line 47: why did you delete this, its there for a purpose

gui/src/components/tracker/TrackerSettings.tsx
```diff
@@ -10,22 +10,22 @@ import {
...
+import { ArrowLink } from '@/components/commons/ArrowLink';
+import { BodyPartIcon } from '@/components/commons/BodyPartIcon';
+import { Button } from '@/components/commons/Button';
+import { CheckBox } from '@/components/commons/Checkbox';
+import { WarningIcon } from '@/components/commons/icon/WarningIcon';
+import { Input } from '@/components/commons/Input';
+import { Typography } from '@/components/commons/Typography';
+import { MountingSelectionMenu } from '@/components/onboarding/pages/mounting/MountingSelectionMenu';
+import { IMUVisualizerWidget } from '@/components/widgets/IMUVisualizerWidget';
 import { SingleTrackerBodyAssignmentMenu } from './SingleTrackerBodyAssignmentMenu';
```
ImUrX, line 29: so if its next to it then we dont do path aliasing? :sweat:

ImUrX (changes requested):

gui/src/components/tracker/TrackerSettings.tsx
ImUrX: i dont have a formed opinion on this yet, you can close this one

gui/package.json
TheDevMinerTV: This ^

gui/src/components/tracker/TrackerSettings.tsx
TheDevMinerTV: IMO, we should only use `@` if it actually makes the path shorter.

TheDevMinerTV (changes requested):

gui/package.json
marcozzxx810: fixed!

gui/src/components/tracker/TrackerSettings.tsx
marcozzxx810: the current setting is relative import depth = 1, which means we allow relative import if the file share the same place as the import file

ImUrX (approved):

ImUrX:
resolve conflicts pls

Louka3000 (approved):
lgtm

ButterscotchV (approved):

ImUrX:
more conflicts qwq

Louka3000:
Damn... @marcozzxx810 I would merge but can you take care of conflicts?

firefox_dioBdNkefh

marcozzxx810:
Resolved all the conflict!

marcozzxx810:
@Louka3000 anything that I can help for this pr?

Louka3000:
> @Louka3000 anything that I can help for this pr?

If @ImUrX is fine with the changes, I'm fine with them... I don't really do GUI stuff.
If I were to merge the PR tho, @TheDevMinerTV 's request would need to be resolved (https://github.com/SlimeVR/SlimeVR-Server/pull/771#discussion_r1257344870)

TheDevMinerTV (changes requested):
What I request is still not resolved.

marcozzxx810:
> What I request is still not resolved.

The plugin itself doesnt has this rule, also it is hard to determine shorter when we do static analysis 

the plugin works when there are matching path that has alias

I dont think it make sense to only use when the path is shorter, it is aim to remove ambiguity of relative path.

Think about this

Assume there are two base {Rootdir GUI}/component/button/base

{Rootdir GUI}/component/button/SecondButton/base


In a file {rootdir GUI}/component/button/SecondButton/common/index.ts, you will use '../../base', logically you need to think about which base but '@/component/button/base" directly tell u which base you are referring to

The goal is not shorter, is to clear the ambiguity

TheDevMinerTV:
ah fair

TheDevMinerTV (approved):
lgtm

marcozzxx810:
merged the latest main!

PR #772, opened 2023-07-09
Fix chest tracker SteamVR battery

Louka3000:
"Such innocent changes can ruin a release"
-Me a couple days ago

### DIFF
--- a/server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
+++ b/server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
@@ -156,6 +156,7 @@ public abstract class SteamVRBridge extends ProtobufBridge implements Runnable {
 		Tracker primaryTracker = null;
 		Tracker secondaryTracker = null;
 		Tracker tertiaryTracker = null;
+		Tracker fourthTracker = null;
 
 		// Given what the role is of localTracker, the tracker positions that
 		// make up that role are set to primaryTracker, secondaryTracker, and
@@ -166,7 +167,7 @@ public abstract class SteamVRBridge extends ProtobufBridge implements Runnable {
 				TrackerPosition.getByTrackerRole(role)
 			);
 		switch (role) {
-			case WAIST:
+			case WAIST -> {
 				secondaryTracker = TrackerUtils
 					.getNonInternalTrackerForBodyPosition(
 						allTrackers,
@@ -178,27 +179,38 @@ public abstract class SteamVRBridge extends ProtobufBridge implements Runnable {
 					tertiaryTracker = TrackerUtils
 						.getNonInternalTrackerForBodyPosition(
 							allTrackers,
+							TrackerPosition.CHEST
+						);
+					fourthTracker = TrackerUtils
+						.getNonInternalTrackerForBodyPosition(
+							allTrackers,
 							TrackerPosition.UPPER_CHEST
 						);
 				}
-				break;
-			case CHEST:
+			}
+			case CHEST -> {
+				// Get chest because primary is upperChest
+				secondaryTracker = TrackerUtils
+					.getNonInternalTrackerForBodyPosition(
+						allTrackers,
+						TrackerPosition.CHEST
+					);
 				// When the waist SteamVR tracking point is disabled, aggregate
 				// waist and hip battery level with the chest.
 				if (!(config.getBridgeTrackerRole(TrackerRole.WAIST, true))) {
-					secondaryTracker = TrackerUtils
+					tertiaryTracker = TrackerUtils
 						.getNonInternalTrackerForBodyPosition(
 							allTrackers,
 							TrackerPosition.HIP
 						);
-					tertiaryTracker = TrackerUtils
+					fourthTracker = TrackerUtils
 						.getNonInternalTrackerForBodyPosition(
 							allTrackers,
 							TrackerPosition.WAIST
 						);
 				}
-				break;
-			case LEFT_FOOT:
+			}
+			case LEFT_FOOT -> {
 				secondaryTracker = TrackerUtils
 					.getNonInternalTrackerForBodyPosition(
 						allTrackers,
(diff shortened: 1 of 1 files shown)


ImUrX (approved):

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
```diff
@@ -153,117 +154,159 @@ protected void writeBatteryUpdate(Tracker localTracker) {
...
-						.getNonInternalTrackerForBodyPosition(
-							allTrackers,
-							TrackerPosition.UPPER_CHEST
+					batteryTrackers
+						.add(
+							TrackerUtils
+								.getNonInternalTrackerForBodyPosition(
+									allTrackers,
+									TrackerPosition.UPPER_CHEST
+								)
```
NWBx01: This is more of a stylistic comment, but I think it's better to order tracker positions based on their precedence. So, if upper chest is the more common position over chest, then upper chest should come first. Since it's part of a list, it doesn't really matter at any rate.

NWBx01 (approved):

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
ImUrX: They are ordered by their precedence, upper chest should have less priority than chest. But let's see Erimel's response

Louka3000: I think it should go from the nearest for the furthest, so if we ever want to do some weighted average or something based on "which tracker is closest to the tracking point", we can. Just pushed a commit for that

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
```diff
@@ -153,143 +154,177 @@ protected void writeBatteryUpdate(Tracker localTracker) {
...
 						);
 				}
-				break;
-			case CHEST:
+			}
+			case CHEST -> {
+				// Add chest because the firs tracker is upperChest
```
ImUrX, line 204:
```suggestion
				// Add chest because the first tracker is upperChest
```

ImUrX (changes requested):

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
NWBx01: > They are ordered by their precedence, upper chest should have less priority than chest.

Not with the chest, though. This was under waist. If upper chest is the primary tracker position for the chest role, then chest is the secondary tracker position for the chest role.

ImUrX (approved):

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
```diff
@@ -153,143 +154,177 @@ protected void writeBatteryUpdate(Tracker localTracker) {
...
+		// lower than the lowest level and has a battery voltage (owoTrack
+		// devices do not).
+		for (Tracker batteryTracker : batteryTrackers) {
+			if (batteryTracker != null && batteryTracker.getBatteryLevel() != null) {
+				lowestLevel = batteryTracker.getBatteryLevel();
+
+				if (batteryTracker.getBatteryVoltage() != null) {
+					trackerVoltage = batteryTracker.getBatteryVoltage();
+				}
 			}
```
ButterscotchV, line 328: `lowestLevel` is no longer checked here, so instead of using the lowest battery level, this PR makes it use the last tracker's battery level. Should be fixed by just adding `batteryTracker.getBatteryLevel() < lowestLevel` to the if condition. I'd make a suggestion but formatting would be a little messed up.

ButterscotchV (changes requested):

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
Louka3000: I cannot believe no one caught this

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
```diff
@@ -153,143 +154,181 @@ protected void writeBatteryUpdate(Tracker localTracker) {
...
+			if (
+				batteryTracker != null
+					&& batteryTracker.getBatteryLevel() != null
+					&& batteryTracker.getBatteryLevel() < lowestLevel
+			) {
+				lowestLevel = batteryTracker.getBatteryLevel();
+
+				if (batteryTracker.getBatteryVoltage() != null) {
+					trackerVoltage = batteryTracker.getBatteryVoltage();
+				}
```
ButterscotchV, line 331: This is mainly just discussion, but this code currently doesn't have any predictable behaviour when it comes to which tracker `trackerVoltage` comes from, so this suggestion just uses the lowest battery tracker's value. There's also the alternative of using the highest tracker voltage for charging detection, so we show if *any* tracker is charging. Thoughs on this?
```suggestion
				if (batteryTracker.getBatteryVoltage() != null) {
					trackerVoltage = batteryTracker.getBatteryVoltage();
				} else {
					trackerVoltage = 0;
				}
```
```suggestion
				if (
					batteryTracker.getBatteryVoltage() != null
						&& batteryTracker.getBatteryVoltage() > trackerVoltage
				) {
					trackerVoltage = batteryTracker.getBatteryVoltage();
				}
```

ButterscotchV (approved):
Approving because it at a minimum matches original behvaiour, but check my suggestion about `trackerVoltage` pls.

server/src/main/java/dev/slimevr/platform/SteamVRBridge.java
ImUrX: i prefer the first, we need to show only the information on lowest battery

ButterscotchV: Actually my second suggestion doesn't match my intended behaviour either, it'd have to be an entirely separate if statement if we want that, so pls don't do as they one suggests.

ImUrX (approved):

PR #773, opened 2023-07-11
Bundle & Receive OSCTrackers

Louka3000:
- Port VRCOSCHandler to kotlin
- Bundle sent OSC Trackers messages to VRChat, reducing network traffic.
- Improve how the VRChat head data is handled
- Receive OSC Trackers data, and adjust it exactly like VRChat does
    The received trackers are treated like Vive trackers in the FK. That is, unless they're controllers or HMD, only their rotation is taken into account.


server/core/src/main/java/dev/slimevr/filtering/QuaternionMovingAverage.kt
```diff
@@ -62,13 +65,13 @@ class QuaternionMovingAverage(
 
 				// Slerps the target rotation to that predicted rotation by
 				// a certain factor.
-				filteredQuaternion = filteredQuaternion.interpR(quatBuf, predictFactor * fpsTimer.timePerFrame)
+				filteredQuaternion = filteredQuaternion.interpR(quatBuf, predictFactor * (fpsTimer?.timePerFrame ?: 1f))
```
ButterscotchV, line 68: I really don't see a case where fpsTimer should be null... idk I guess this is fine, but it's an edge case that maybe should be detected and reported instead of potentially introducing unexplained weird behaviour?

server/core/src/main/java/dev/slimevr/osc/VRCOSCHandler.kt
```diff
@@ -0,0 +1,485 @@
...
+		// Update current time
+		val currentTime = System.currentTimeMillis().toFloat()
+
+		// Gets timer from vrServer
+		if (fpsTimer == null) {
+			fpsTimer = VRServer.instance.fpsTimer
+		}
+		// Update received trackers' offset rotation slerp
+		if (receivingRotationOffset != receivingRotationOffsetGoal) {
+			receivingRotationOffset = receivingRotationOffset.interpR(receivingRotationOffsetGoal, OFFSET_SLERP_FACTOR * (fpsTimer?.timePerFrame ?: 1f))
```
ButterscotchV, line 324: Ideally, we shouldn't be using a constant interval for our slerp time, this will cause inconsistent movement speed. It really depends on the rate here whether that wouldn't matter or what, but I think we should instead use the time between the last received rotation and the current one, so it would take an equivalent time to interpolate as the time between the messages. If it's always consistent, it should stay smooth, but that's really only an ideal... The *"best"* solution would be introducing a buffer (latency) to prevent jitter, but that may not be ideal for our low-latency goal... I'd like some input on this.

server/core/src/main/java/dev/slimevr/osc/VRCOSCHandler.kt
```diff
@@ -0,0 +1,485 @@
...
+								e
+						)
+				}
+			} catch (e: OSCSerializeException) {
+				if (currentTime - timeAtLastError > 100) {
+					timeAtLastError = System.currentTimeMillis()
+					LogManager
+						.warning(
+							"[VRCOSCHandler] Error sending OSC message to VRChat: " +
+								e
```
ButterscotchV, line 408: I see the code was already like this before, but can't `e` be passed as an argument? It should be using string interpolation (i.e. `"error: $e"`) otherwise anyways.

server/core/src/main/java/dev/slimevr/filtering/QuaternionMovingAverage.kt
Louka3000: It’s because this code can potentially run before vrserver lateinit

server/core/src/main/java/dev/slimevr/osc/VRCOSCHandler.kt
Louka3000: Using this kind of interpolation is intentional, I talked about it with Kung as I initially used my Quaternion filtering class (that does consistent smoothing/no acceleration) 
The design goal is that the rotation should not change if the user is moving their head a bit, but change more if they happen to do a 180. This is only used if constant head rotation is received, for debug or constant alignment (in which case the speed shouldn’t matter)

server/core/src/main/java/dev/slimevr/filtering/QuaternionMovingAverage.kt
ButterscotchV: Hmm alright.

server/core/src/main/java/dev/slimevr/osc/VRCOSCHandler.kt
ButterscotchV: Hmm okay. I don't quite get your explanation, but if there's solid reasoning behind this decision then that's fine.

ButterscotchV (approved):

server/core/src/main/java/dev/slimevr/osc/VRCOSCHandler.kt
Louka3000: Done

ImUrX (approved):
those 26 `!!` >~>

Louka3000:
Kotlin auto-conversion be like

PR #774, opened 2023-07-11
Add accessibility settings

ImUrX:
Moved the interface settings into its own menu and separated it into Notifications and Appearance.
In appearance I added a font size range and a font selector with included fonts ^-^

Added the following fonts:
* OpenDyslexic
* Lexend
* Noto Sans (CJK)
* Ubuntu
* OS's sans-serif font


ImUrX:
rest of commits are me trying to reduce filesize the maximum amount i can

Cargo.toml
```diff
@@ -11,3 +11,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 rust-version = "1.65" # This version stabilized GATs and let-else
 repository = "https://github.com/SlimeVR/SlimeVR-Server"
+
+[profile.release]
+lto = true
+strip = "debuginfo" # Only affects Unix binaries with DWARF
```
TheButlah, line 17: Does this actually matter? Release builds already have `debug = false`: https://doc.rust-lang.org/cargo/reference/profiles.html#release - I'm not sure that stripping debug symbols when they don't exist already, matters

As for LTO: Fat lto significantly reduce binary size, but like 10xes the link time in projects. As long as everyone thinks this tradeoff is worth it, I'm good with it. You could use think LTO instead if you wish, it isn't very slow to run. But it doesn't decrease binary size nearly as much.

For some concrete numbers: I looked at one of the other CI runs, rust portion of build times was 2m 54s
This PR's rust build time is 4m 39s. So its almost a 2 minute increase to CI times.

ImUrX: Im not stripping symbols, im stripping [debuginfo](https://doc.rust-lang.org/rustc/codegen-options/index.html#strip), I could also strip everything else, it would further decrease the Linux binary size.

TheButlah: Yes... but how does that change anything I said

ImUrX: before debuginfo:
```bash
objdump -g slimevr | wc --bytes
24739162
```
after debuginfo:
```bash
$ objdump -g slimevr | wc --bytes
5878290
```

TheButlah: these numbers are with release mode?

ImUrX: Yeah, I did it with the CI of [`164e393`](https://github.com/SlimeVR/SlimeVR-Server/pull/774/commits/164e393e64d58fa476537f2837bf8d2a148a6fb4) and [`da67e60`](https://github.com/SlimeVR/SlimeVR-Server/pull/774/commits/da67e600be3d07b2bccb96eedf5c2f820ce9ebc0)

Let me try in my side with each `strip` option

ImUrX: no `strip`:
```bash
$ ls -lh target/release/slimevr
-rwxr-xr-x 2 uri users 29M Jul 17 18:49 target/release/slimevr
```

`debuginfo`:
```bash
$ ls -lh target/release/slimevr
-rwxr-xr-x 2 uri users 27M Jul 17 18:53 target/release/slimevr
```

`true`/`symbols`:
```bash
$ ls -lh target/release/slimevr
-rwxr-xr-x 2 uri users 25M Jul 17 18:55 target/release/slimevr
```

TheButlah: huh, interesting. I wonder how that works under the hood, maybe I misunderstand what the `debug` vs `strip` is doing.

PR #775, opened 2023-07-14
Make displayName consistent via hash

Louka3000:
Old way of doing it used the VRServer tracker ID, which means each tracker ID depended on when it was connected, and changed upon server restart.
This uses the cropped hash256 of the mac address
Note: the displayName is currently never saved to config, and the name is used for config, so this is solely a UX change.

image


ButterscotchV (approved):
Kinda a personal preference sorta thing. I think it's neat, might be helpful in a rare case.

ImUrX (approved):

PR #776, opened 2023-07-14
Improvements on the Autobone GUI

ImUrX:
It improves the autobone GUI by forcing you to have an HMD and also letting you setup the user's height, now it can also show failure and the ETA is more precise.
Many improvements were made on the autobone code itself by simplyifing on just reusing `SkeletonConfigOffsets`. Also some bugs were found on how offsets were compensated.


ButterscotchV (approved):
I approve of the GUI code, I wrote the server code so my review of that doesn't mean much, but I did double check it.

Eirenliel (approved):

PR #790, opened 2023-07-20
Improve debugging experience

ImUrX:
Make last log be rotated so we can have the previous run log be saved.
Made dev environment use portable config mode!


Louka3000 (approved):

Eirenliel:
What? FileHandler already keeps 2 logs, if you need more you can just increase the number.

Eirenliel (approved):

PR #794, opened 2023-07-23
Implement packet bundling

0forks:
Companion PR: https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/263

ESP8266 does not support frame aggregation natively in its networking implementation, this PR adds it on application level to decrease airtime by 50-70%.
- New packet 22 (0x16): feature flags - packet-sized bitset (uint8[length]);
- New packet 100 (0x64): bundle - contains multiple type- (uint32) and length- (uint16) prefixed packets.


Eirenliel:
I merged firmware without properly digging into packing of the flags. Packing sounds cool, but it's hard for me to understand easier if that's trully expandable if we add for example some flags that go over current maximum bits and it's received by an older version sf/fw. Should we maybe just keep it 16bit without packing?

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
```diff
@@ -391,6 +393,17 @@ class TrackersUDPServer(private val port: Int, name: String, private val tracker
 					"[TrackerServer] User action from ${connection.descriptiveName } received. $name reset performed."
 				)
 			}
+
+			is UDPPacket22FeatureFlags -> {
+				if (connection == null) return
+				// Respond with server flags
+				bb.limit(bb.capacity())
+				bb.rewind()
+				parser.write(bb, connection, packet)
```
Eirenliel, line 402: Does it just send the packet back without actually setting anything that server supports?

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPPacket.kt
```diff
@@ -320,6 +320,19 @@ data class UDPPacket21UserAction(var type: Int = 0) : UDPPacket(21) {
 	}
 }
 
+class UDPPacket22FeatureFlags(
```
Eirenliel, line 323: Is there a reason why it's a separate packet instead of putting it into handshake or sensor info?

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/FeatureFlags.kt
```diff
@@ -0,0 +1,74 @@
+package dev.slimevr.tracking.trackers.udp
+
+import java.nio.ByteBuffer
+
+/**
+ * Bit packed flags, enum values start with 0 and indicate which bit it is.
```
Eirenliel, line 6: Wait it's a bit set... Okay, I understand now. I'm not sure if it's better than being an int list or not...

Eirenliel (changes requested):

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPPacket.kt
0forks: Data exchange is bidirectional here, adding extra data to server->firmware handshake breaks owotrack, though I didn't dig into the handshake format too much.

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
0forks: It's abstracted into `writeData` in packet classes:
https://github.com/SlimeVR/SlimeVR-Server/blob/f84efc413dc7cb0e4b810609a17aaab55491c92f/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPProtocolParser.kt#L34-L38

0forks:
> for example some flags that go over current maximum bits and it's received by an older version sf/fw

The receiving side has a fixed size buffer and it's checked with a `min(received.length, flags.length)`, so the extra bytes will be thrown out.

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/FeatureFlags.kt
```diff
@@ -0,0 +1,74 @@
...
+		@JvmStatic
+		fun from(received: ByteBuffer, length: Int): FirmwareFeatures {
+			val res = FirmwareFeatures()
+			res.available = true
+			received.get(res.flags, 0, Math.min(res.flags.size, length))
+			return res
+		}
+	}
+
+	private var available = false
```
ImUrX, line 40: You dont need to make a getter, Kotlin already makes one
```suggestion
	var available = false
		private set
```

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/FeatureFlags.kt
```diff
@@ -0,0 +1,74 @@
...
+
+	companion object {
+		private var flagsEnabled: List<ServerFeatureFlags> = listOf(
+			PROTOCOL_BUNDLE_SUPPORT
+
+			// Add enabled flags here
+		)
+
+		private val flagsLength = BITS_TOTAL.ordinal / 8 + 1
+		private val flags = run {
```
ImUrX, line 60: You can make this public, why have it as `getPacked` if you already have a field for it?

All these fields are `val` (immutable) so they cant be changed, not even the `List` because its not a `MutableList`

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPDevice.kt
```diff
@@ -53,6 +53,9 @@ class UDPDevice(
 	var timedOut = false
 	override val trackers = ConcurrentHashMap<Int, Tracker>()
 
+	@JvmField
```
ImUrX, line 56: I honestly prefer not having these if we are adding a new field, it was mainly for when we were migrating old stuff to Kotlin
```suggestion
```

ImUrX (changes requested):

ImUrX (approved):

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
Eirenliel: But the packed is the same that was received from the tracker.

Eirenliel (approved):

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
0forks: It writes a fixed value, different from what server had received.

```kotlin
: UDPPacket(22) {
		firmwareFeatures = FirmwareFeatures.from(buf, buf.remaining())
	}

	override fun writeData(buf: ByteBuffer) {
		buf.put(ServerFeatureFlags.packed)
	}
```

Do we want server flags to be toggleable from the UI? If so, we can make `ServerFeatureFlags` writable, but the firmware only accepts 1 flags packet per connection, will need to remove that check.

PR #798, opened 2023-07-27
Quick fix for BVH issues

ButterscotchV:
This will fix the problem with BVH recordings temporarily, but could have issues in the long term due to only accounting for a parent with the same name and only on one level. There will be a more proper fix later down the line, but for now this should at least get it working with Blender.


Louka3000 (approved):
beautiful hack

Eirenliel (approved):

PR #803, opened 2023-08-02
things I got pinged about to fix

ImUrX:
* @Louka3000 asked me to first mention hmd height and then real height because thats how the calculation is done.
* @sctanf asked me to add zero padding to voltage on developer mode
* @ZRock35 showed me a padding bug because of flexbox not wrapping on serial console


loucass003 (approved):
LGTM

Eirenliel (approved):

PR #805, opened 2023-08-03
File Save RPC

ImUrX:


marcozzxx810:
Do we have a issue for this PR?

(not related to this PR, but talking about this issue) 


I think it will be great if we can save all the gui config.

Also thinking about pulling the config from a URL.

ImUrX:
> Do we have a issue for this PR?

nope, we dont. 


> I think it will be great if we can save all the gui config.

what for? I think it's fine for us to keep using `localStorage`

PR #806, opened 2023-08-04
Resets settings

Louka3000:
Adds an option to enable mounting reset for feet, for those that would rather have something than nothing :3 
Added 3 more options for the arms behaviour upon reset
1. Back (default, what we already have)
2. Forward (every part of the arm forward, useful for tubing)
3. T-pose up (full reset in i-pose, mounting reset in t-pose)
4. T-pose down (full reset in t-pose, mounting reset in i-pose)

~~needs https://github.com/SlimeVR/SolarXR-Protocol/pull/121~~


server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
```diff
@@ -217,36 +248,54 @@ class TrackerResetsHandler(val tracker: Tracker) {
...
-			(if (reverseYaw) yawAngle else yawAngle - Math.PI.toFloat()),
-			0f
-		).toQuaternion()
-
-		// Get the difference from the last mounting to the current mounting and apply
-		// the difference to the yaw fix quaternion to correct for the rotation change
-		yawFix *= (buffer * mountRotFix.inv()).inv()
-		mountRotFix = buffer
+	fun resetMounting(reference: Quaternion) {
+		if (resetMountingFeet || !isFootTracker()) {
```
ButterscotchV, line 252: Pls just make this return instead of nesting the entire method

ButterscotchV (approved):

Eirenliel (approved):

PR #809, opened 2023-08-08
Kotlin leg tweaks

Stermere:
Converts the legTweaks files to Kotlin and improves the readability a touch. ~~There should be no difference in functionality~~ This fixes a minor bug that has been present since the math lib switch, so there should be a slight improvement!


server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.kt
```diff
@@ -0,0 +1,1167 @@
...
+	private var bufferInvalid = true
+
+	constructor(skeleton: HumanSkeleton, config: LegTweaksConfig) : this(skeleton) {
+		this.config = config
+		updateConfig()
+	}
+
+	fun getLeftFootPosition(): Vector3 {
+		return leftFootPosition
+	}
```
Louka3000, line 132: I would just make all of these public variables and remove the get/set

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaksBuffer.kt
```diff
@@ -0,0 +1,821 @@
...
+
+	fun setCorrectedRotations(leftFootRotation: Quaternion, rightFootRotation: Quaternion) {
+		this.leftFootRotationCorrected = leftFootRotation
+		this.rightFootRotationCorrected = rightFootRotation
+	}
+
+	// getters and setters
+	fun getLeftFootPosition(): Vector3 {
+		return leftFootPosition
+	}
```
Louka3000, line 242: Same here, just make them public imo

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaksBuffer.kt
```diff
@@ -0,0 +1,821 @@
...
+	private fun detectOutsideForces(f1: Vector3, f2: Vector3): Boolean {
+		val force: Vector3 = GRAVITY.plus(f1).plus(f2)
+		val error: Vector3 = centerOfMassAcceleration.minus(force)
+		return error.lenSq() > FORCE_ERROR_TOLERANCE_SQR
+	}
+
+	// simple error function for the force vector gradient descent
+	private fun getForceVectorError(testForce: Vector3, otherForce: Vector3): Vector3 {
+		return centerOfMassAcceleration
+			.minus(testForce.plus(otherForce).plus(GRAVITY))
```
Louka3000, line 819:
```suggestion
		return centerOfMassAcceleration - testForce + otherForce + GRAVITY
```
Just as an example, you can do this in Kotlin, which makes it easier to read IMO. You can look through your code and use +, -, * and / wherever

Louka3000 (changes requested):

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.kt
Stermere: didn't know Kotlin had that feature! thanks

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/LegTweaks.kt
```diff
@@ -0,0 +1,1079 @@
...
+
+	fun setSkatingReductionEnabled(skatingCorrectionEnabled: Boolean) {
+		this.skatingCorrectionEnabled = skatingCorrectionEnabled
+
+		// reset the buffer
+		bufferHead = LegTweaksBuffer()
+		bufferInvalid = true
+	}
+
+	fun setToeSnapEnabled(value: Boolean) {
```
Louka3000, line 151: You should also be able to remove these getters and setters up to line 186

Louka3000 (approved):

ButterscotchV (approved):
Code looks fine, but I wouldn't trust my review as a definite. It was very hard to review this so I definitely could've missed something... Just make sure this is well tested please. If not before the PR is merged, then *absolutely* before a release is made.

Stermere:
I've tested all the features that I would expect to break but have not yet done all the integration tests. Once I have and everything works as it should I'll update this comment.

ImUrX:
how is this PR doing?

Stermere:
> how is this PR doing?

It's ready, I re-tested it after rebasing it to the bones rewrite about 2 weeks ago.

ImUrX (approved):

PR #818, opened 2023-08-12
Fix OSCTrackers yawAlign

Louka3000:
now for main instead of bones so this can get merged faster

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.java
+++ b/server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.java
@@ -87,8 +87,8 @@ public class HumanSkeleton {
 		0
 	).toQuaternion();
 	// #region Tracker Input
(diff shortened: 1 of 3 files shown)


server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.java
```diff
@@ -87,8 +87,8 @@ public class HumanSkeleton {
 		0
 	).toQuaternion();
 	// #region Tracker Input
-	protected Tracker headTracker;
-	protected Tracker neckTracker;
+	public Tracker headTracker;
+	public Tracker neckTracker;
```
Louka3000, line 91: Note: these will be public with the kotlin conversion anyways

ImUrX (approved):
looking sexy

Louka3000:
> looking sexy

no u

ButterscotchV (approved):

PR #831, opened 2023-08-23
tauri should work now

ImUrX:


TheButlah (approved):
Approving - but the fact that this is necessary reinforces my opinion that maybe upgrading to tauri 2 when its still in alpha is jumping the gun

Louka3000 (approved):

Eirenliel (approved):

PR #832, opened 2023-08-25
Make UDPPackets Floats safe

Louka3000:
NaN propagate and break everything. Needs testing to see if this actually solves the problem (otherwise than that, runs as expected)


server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPPacket.kt
```diff
@@ -153,7 +149,7 @@ data class UDPPacket4Acceleration(var acceleration: Vector3 = Vector3.NULL) :
 	UDPPacket(4), SensorSpecificPacket {
 	override var sensorId = 0
 	override fun readData(buf: ByteBuffer) {
-		acceleration = Vector3(buf.float, buf.float, buf.float)
+		acceleration = Vector3(UDPUtils.getSafeBufferFloat(buf), UDPUtils.getSafeBufferFloat(buf), UDPUtils.getSafeBufferFloat(buf))
```
ButterscotchV, line 152: This is fine, but in the future we might as well make a method for Vectors too. If you want you can make that now or just leave it for later, either is fine really.

ButterscotchV (approved):
The behaviour between Quaternions and Vectors isn't consistent, but that doesn't really matter. Should work fine for keeping numbers safe.

ImUrX (approved):

PR #836, opened 2023-09-01
Fix config shenanigans

ImUrX:
Fixed how config looks, and fixed the actual bug which was the layout effect for onboarding check triggering whenever config was changed in general instead of just when the onboarding key happening.


loucass003 (approved):
LGTM

Louka3000 (approved):

PR #842, opened 2023-09-08
Android ByteBuffer and Color fixes

ButterscotchV:
This didn't compile/run fine on Android. Simple one-line fix, this should be equivalent functionality. Also as a small note, the `limit(int)` method returns `Buffer` instead of `ByteBuffer`, so it's just on a separate line as it affects the object it's called on.

This doesn't compile/run because the method used is only available in SDK 34 (Android 14).
https://developer.android.com/reference/java/nio/ByteBuffer#slice(int,%20int)

Also worth mentioning that this could've been avoided by compiling fully with Android, though I like our current structure... If we could have some way to compile `:server:core` with Android *and* with regular Java depending on which is requested, that would be nice.


ImUrX:
> If we could have some way to compile :server:core with Android and with regular Java depending on which is requested, that would be nice.

i think i can do that, i have had something in my list to do it with kotlin multiplatform

ButterscotchV:
Also fixed OSC missing `java.awt.Color` now by a simple hack lmfao, Android Studio complains about it, but it compiles just fine!

Louka3000:
Can’t OSCStatic and its usages be removed now?

ButterscotchV:
> Can’t OSCStatic and its usages be removed now?

Yeah probably, I'm not sure if keeping this class in would cause any problems or not? I can do that in this PR if you would like.

Louka3000:
>  I can do that in this PR if you would like.

yes please

Louka3000 (approved):

ImUrX (approved):

PR #849, opened 2023-09-20
Fix RPCUtil not checking for the interface before looking for it

ImUrX:


Louka3000 (approved):
looks fine to me but didn't test

ButterscotchV (approved):
Looks fine and I did test it, the GUI finally actually shows my IP now, magnificent

PR #851, opened 2023-09-25
Track connection by socket address if HWID is unavailable

ButterscotchV:
It seems that sometimes owoTrack or the sort doesn't send HWID properly. That's not really good, but we should be able to handle it anyways by falling back to tracking the connection by the IP address and port. This isn't necessarily a bug, but it was changed in the last release and has broken previously "functioning" connection tracking.

I've also tried to clean up `TrackersUDPServer.kt` a bit and explain some of the things with comments. I removed the second MAC address check because the first one should find all of them and handle it just fine.


Louka3000 (approved):
lgtm

ImUrX (approved):

PR #856, opened 2023-09-26
Add a hash check for the jar when running

ImUrX:
This will warn of the user when the server jar is modified and ask if they want to continue running


ImUrX:
will close this but I got some ideas for later because of it

Issue #24, opened 2021-08-12
Prompt when closing if user still has tracker connected. 


Torikari:
In case of misclick or user still have tracker forgot to be turned off.

Eirenliel:
![](https://github.trello.services/images/mini-trello-icon.png) [Prompt when closing if user has trackers connected](https://trello.com/c/pMiIJ64J/38-prompt-when-closing-if-user-has-trackers-connected)

Eric13246:
I would like to work on this one as a newcomer, can someone assign this to me please?

Louka3000:
Don’t need it to be assigned for you to work on it if you want to but here you go ^^ @Eric13246

ImUrX:
Done by #862

PR #862, opened 2023-09-30
Prompt before closing SlimeVR when trackers are still connected

Eric13246:
Completed task #24

Users will be prompted to confirm if they want to close SlimeVR if trackers are still connected. This warning is also configurable in the app's settings.


loucass003 (changes requested):
Please change the naming of CloseApp() to closeApp 
And move the function inside the component as a closure

Alse use braces on if statements

loucass003 (approved):

loucass003:
> That should do it. By the way, are there any unit tests or integration tests that are required when adding features?

No we do not have unit test.
This pr is first gonna be sent as a release candidate, and we have a full process of test done manually to ensure no critical feature is broken

ImUrX (approved):

PR #870, opened 2023-10-15
Add new message from firmware for finding server

ButterscotchV:
Since https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/256, the message in the firmware for "Looking for the server" has been changed to "Searching for the server". This PR just adds the new message so WiFi provisioning will work with the new firmware.

We really gotta make a better way of doing this lol. 0.10.2 when??


TheDevMinerTV (approved):
LGTM

Louka3000 (approved):
That is indeed a questionable way to do it, but lgtm

PR #900, opened 2023-11-17
Allow multiple servers in the same network

### LINKED ISSUE #74: Make it possible for multiple users on the same network to use SlimeVR
All it needs is an input field, would be easy to build into the doc pages while setting up firmware to change the port field there

ImUrX:
Depends on https://github.com/SlimeVR/SolarXR-Protocol/pull/127

This adds the following features:
* A list of known devices that only works with mac addresses
* A new model version for config and migrate by adding all known trackers that are mac addresses to a set called `knownDevices`
* A way to forget connected devices (FIXME: for some reason it ends up disconnecting after 30 seconds have passed instead of immediately)
* A prompt whenever a new unknown device is found, where you are asked if you want to connect to it or ignore it for the rest of this session.
* Make trackers auto-connect without a prompt when on the connect trackers page

Fixes #74


salmoooon:
Didn't know this wasn't supported and ended up spending too much time retrying, setting up a VLAN, and then buying a router as a last-ditch effort.

Bump on this and hope this could be merged soon!

Eirenliel (approved):

PR #906, opened 2023-11-28
Add new IMUs, show IMU calibration during onboarding for LSM6DSV

kounocom:
Depends on https://github.com/SlimeVR/SolarXR-Protocol/pull/129

Adds new IMUs so they show up in tracker info, and displays IMU calibration tutorial during onboarding if an LSM6DSV tracker is connected


ImUrX:
LSM is not supported by the current firmware from what I checked?

PR #907, opened 2023-11-28
Add reboot retries when trying to connect to WiFi

ImUrX:
I had my slime fail to connect to Wi-Fi and just restarting it makes things work it seems


ButterscotchV (approved):
Feels really weird using a `byte` for that but it makes sense, just not used to it

Louka3000 (approved):

PR #909, opened 2023-12-13
allow to set the socket dir via SLIMEVR_SOCKET_DIR

Links2004:
this is needed if the vrserver is running in a chroot env and we need to mount the sockets to the HOME dir.

see: https://github.com/SlimeVR/SlimeVR-OpenVR-Driver/issues/43


ImUrX (approved):

ButterscotchV (approved):
I cannot comment on the usefulness of this PR, but the code should work.

PR #913, opened 2023-12-21
Add initial HID trackers support

sctanf:
making this now because it is pain keeping the branch up to date

Add support for USB device acting as a receiver for HID tracker data

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/trackers/hid/TrackersHID.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/trackers/hid/TrackersHID.kt
@@ -1,0 +1,247 @@
+package dev.slimevr.tracking.trackers.hid
+
+import com.jme3.math.FastMath
+import dev.slimevr.VRServer
+import dev.slimevr.tracking.trackers.Device
+import dev.slimevr.tracking.trackers.Tracker
+import dev.slimevr.tracking.trackers.TrackerStatus
+import dev.slimevr.tracking.trackers.udp.IMUType
+import io.eiren.util.logging.LogManager
+import io.github.axisangles.ktmath.Quaternion
+import io.github.axisangles.ktmath.Quaternion.Companion.fromRotationVector
+import io.github.axisangles.ktmath.Vector3
+import org.hid4java.HidDevice
+import org.hid4java.HidManager
+import org.hid4java.HidServices
+import org.hid4java.HidServicesListener
+import org.hid4java.HidServicesSpecification
+import org.hid4java.event.HidServicesEvent
+import java.util.function.Consumer
+import kotlin.experimental.and
+
+/**
+ * Receives trackers data by UDP using extended owoTrack protocol.
+ */
+class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>) : Thread(name), HidServicesListener {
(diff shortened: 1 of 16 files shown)


server/core/src/main/java/dev/slimevr/tracking/trackers/hid/TrackersHID.kt
ImUrX, line 1: this should have some kind of abstract class and then an implementation inside the desktop subproject, because `hid4java` is desktop exclusive it seems.

you can see examples in the serial and SteamVR implementation

server/core/src/main/java/dev/slimevr/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -0,0 +1,249 @@
...
+				null,
+				trackerNum = trackerId,
+				hasRotation = true,
+				hasAcceleration = true,
+				userEditable = true,
+				imuType = sensorType,
+				allowFiltering = true,
+				needsReset = true,
+				needsMounting = true,
+				usesTimeout = false
```
ImUrX, line 110: using timeout is based on you executing the `heartbeat()` function of the tracker each time you receive a tracker packet.

The idea is that when it's taking too long it first gets a `TrackerStatus.TIMED_OUT` and after a few more it gets `TrackerStatus.DISCONNECTED`

The special thing about the `TIMED_OUT` status is to give less priority in tracker assignment and also to let the user know it's getting timed out.

It should be used in my opinion but the HID device should communicate some kind of heartbeat packet. (I know BLE is more like TCP, but maybe there is a way to have access to the heartbeat packets?)

ImUrX (changes requested):

server/core/src/main/java/dev/slimevr/tracking/trackers/hid/TrackersHID.kt
sctanf: The timeout does get reset every time the tracker is updated, and it will work, but the tracker will not send updates if it doesn't need to. In this case, if timeout is used, whenever the tracker is updated the status also should be forced to `OK`.

The tracker also sleeps so quickly that heartbeat packets don't make sense. Plus, if the receiver is removed, all the linked trackers already get marked as disconnected, so I chose to not use timeout. Otherwise there might be no visible difference between tracker not sending data for any reason versus the receiver being disconnected.

ImUrX: How do you know the device gets shutdown or lost connection then?

sctanf: In context of trackers themselves, there is no way, aside from using a timeout
Should timeout be used anyways, even if `DISCONNECTED` could have multiple causes?

ImUrX: Ok so we go back again, how do you timeout a BLE tracker? What do you use as a heartbeat?

Eirenliel (approved):

ImUrX (approved):
oh you made it desktop-exclusive, i guess its good enough for now

PR #920, opened 2024-01-06
Positional tracker support

Stermere:
Adds an inverse kinematics solver to the skeleton allowing for positional and rotational trackers to be used together. An example would be using 3 positional trackers for the waist and feet and 2 rotational trackers on your leg to add knee-tracking. That is just one example though, any combination of trackers is theoretically supported.


server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
ImUrX, line 1: why does constraint need to be abstract? why is there so many vars instead of vals? For example does `allowModifications` need to be modifiable?

I would very much like to reduce the OOP-style of this

ImUrX (changes requested):

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
Stermere: Constraint is abstract because different joints have different constraints, originally I wanted to make a hinge joint for knees and elbows but ended up not getting to that, there are still different constraint types though for example the CompleteConstraint is for bones that the solver should ignore, I don't know how I would do this without an abstract class but I am open to suggestions! With the current way things are set up the Var's cannot be changed to Val's without significant change, probably pulling the Constraint out of the Bone class, and moving it into the IKChain's. Let me know if that change sounds reasonable, and if you still don't like the abstract class what do you suggest I do to remove it?

ImUrX (approved):

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/IKChain.kt
```diff
@@ -0,0 +1,143 @@
...
+		return sum / weightSum
+	}
+
+	/**
+	 * Resets the chain to its default state
+	 */
+	fun resetChain() {
+		distToTargetSqr = Float.POSITIVE_INFINITY
+		centroidWeight = 1f
+		val lockedReduction = if (locked)IKSolver.LOCKED_REDUCTION else 1f
```
ImUrX, line 86:
```suggestion
		val lockedReduction = if (locked) IKSolver.LOCKED_REDUCTION else 1f
```

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/IKChain.kt
```diff
@@ -0,0 +1,143 @@
...
+	fun resetTrackerOffsets() {
+		computedTailPosition?.reset(bones.last().getTailPosition())
+		computedBasePosition?.reset(bones.first().getPosition())
+	}
+
+	/**
+	 * Allow constrained bones to deviate more
+	 */
+	fun decreaseConstraints() {
+		val lockedReduction = if (locked)IKSolver.LOCKED_REDUCTION else 1f
```
ImUrX, line 108:
```suggestion
		val lockedReduction = if (locked) IKSolver.LOCKED_REDUCTION else 1f
```

ImUrX (changes requested):

Eirenliel (approved):

Eirenliel:
Would've been cool to use the constraints to correct for drift 🥺 like we know if the bone is out of constraints, then something drifted, and we can at least limit to how bad it looks, and also maybe warn the user. It's a powerful thing that we wanted long ago, and now it feels reachable

ImUrX (approved):

Stermere:
> Would've been cool to use the constraints to correct for drift 🥺 like we know if the bone is out of constraints, then something drifted, and we can at least limit to how bad it looks, and also maybe warn the user. It's a powerful thing that we wanted long ago, and now it feels reachable

That is something I am definitely interested in making happen in another PR! You are right though it is certainly reachable with the Constraints as a start

Eirenliel (approved):

PR #922, opened 2024-01-15
Add pause tracking toggle to user action packet

ButterscotchV:
I missed this before, pause tracking already been added to basically everywhere else, so I figure this is fitting


ImUrX (approved):

Louka3000 (approved):
Lgtm

PR #923, opened 2024-01-23
Use tauri store for GUI config

ImUrX:
Uses tauri store if tauri is being used, if not, localStorage is still used.
There is multiple reasons for this like being able to backup your GUI config, and not losing localStorage if for some reason Tauri changes the domain url.

There is a migration script also


gui/src/hooks/config.ts
```diff
@@ -86,18 +100,16 @@ export function useConfigProvider(): ConfigContext {
...
-            dir: BaseDirectory.AppConfig,
-          }).catch(() => null);
+          const oldConfig = localStorage.getItem('config.json');
 
-          if (oldConfig) localStorage.setItem('config.json', oldConfig);
+          if (oldConfig) await store.set('config.json', oldConfig);
 
-          localStorage.setItem('configMigrated', 'true');
+          localStorage.setItem('configMigratedToTauri', 'true');
         }
```
ButterscotchV, line 110: Shouldn't you be setting `configMigratedToTauri` in the `store` config, not `localStorage`?

ImUrX: i guess it really doesnt matter haha

ImUrX: i will use the `store` then

Louka3000 (approved):

ButterscotchV (approved):

PR #929, opened 2024-01-30
Add build dependencies to nix flake

AnnoyingRain5:
These dependencies are required for building on nixos, they should be in the flake.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -94,6 +94,9 @@
               librsvg
               freetype
               expat
+              husky
+              cargo
+              rustc
             ])
             ++ lib.optionals pkgs.stdenv.isDarwin [
               pkgs.darwin.apple_sdk.frameworks.Security


ImUrX:
Rust is already included via fenix, is husky not included?

AnnoyingRain5:
When I ran the script on my nixOS system, I can confirm that rustc was not included by that package. I had to manually add these packages to the flake before it would compile.

ImUrX:
That's so weird, why does it download them for me then

AnnoyingRain5:
Have you tried setting your system to sandbox mode?

https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#tested-using-sandboxing

also, what system are you using nix on?

ImUrX:
NixOS, will try it when I have time

ImUrX:
Sandboxing is already enabled, nothing changed for me.

ImUrX:
I guess I'm gonna go back to the classic `nix develop` instead of the fancy devenv, I don't really wanna debug it.

AnnoyingRain5:
If it helps, my dotfiles are at https://github.com/AnnoyingRain5/dotfiles

ImUrX:
can you check if #1023 fixes the problem?

AnnoyingRain5:
Yep! It looks like it's fixed pending #1023

PR #941, opened 2024-02-16
try removing kotlin warnings

ImUrX:
Saw that there were a lot of kotlin warnings, so fixed some

### DIFF
--- a/server/core/src/main/java/dev/slimevr/autobone/AutoBone.kt
+++ b/server/core/src/main/java/dev/slimevr/autobone/AutoBone.kt
@@ -20,7 +20,6 @@ import org.apache.commons.lang3.tuple.Pair
 import java.io.File
 import java.util.*
 import java.util.function.Consumer
-import java.util.function.Function
 
 class AutoBone(server: VRServer) {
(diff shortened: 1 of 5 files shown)


server/core/src/main/java/dev/slimevr/autobone/AutoBone.kt
ButterscotchV, line 1: We cannot assume that `humanPoseManager` is not null, as it may be initialized before AutoBone in `VRServer`. It could *technically* be considered safe since I moved the initialization lower in `VRServer`, so it could be changed... Idk, I guess I won't block it over this, but technically it is not assured to be safe, I'd left it in before despite the warnings because of this reason.

ButterscotchV (approved):

server/core/src/main/java/dev/slimevr/autobone/AutoBone.kt
ImUrX: thats an interesting thing, what I'm curious about is that it's not a lateinit and kotlin is sure that it's not null, is something wrong in the parser amybe?

PR #944, opened 2024-02-17
Fix state bug on tray exit

ImUrX:
The first time you configure if you want to use the tray, the topbar might still think config is null for `useTray`. This just fixes it

### DIFF
--- a/gui/src/components/TopBar.tsx
+++ b/gui/src/components/TopBar.tsx
@@ -102,7 +102,7 @@ export function TopBar({
     return () => {
       unlisten.then((fn) => fn());
     };
-  }, []);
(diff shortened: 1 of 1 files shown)


ButterscotchV (approved):

gui/src/components/TopBar.tsx
```diff
@@ -265,6 +265,7 @@ export function TopBar({
         accept={async (useTray) => {
           await setConfig({ useTray });
           setShowTrayOrExitModal(false);
+          // forceUpdate()
```
Louka3000, line 268:
```suggestion
```
leftover from debugging?

Louka3000 (changes requested):

Louka3000 (approved):

PR #948, opened 2024-02-26
add a button to unassign vrm model

lamystere:
Currently there is no way to remove a vrm model once one is loaded.

### DIFF
--- a/gui/src/components/settings/pages/VMCSettings.tsx
+++ b/gui/src/components/settings/pages/VMCSettings.tsx
@@ -59,9 +59,10 @@ export function VMCSettings() {
     setTimeout(() => setFlashLoaded(!bool), 1000);
   };
 
-  const { reset, control, watch, handleSubmit, resetField } = useForm<VMCSettingsForm>({
-    defaultValues: defaultValues,
-  });
+  const { reset, control, watch, handleSubmit, resetField } =
+    useForm<VMCSettingsForm>({
+      defaultValues: defaultValues,
+    });
 
   const onSubmit = async (values: VMCSettingsForm) => {
     const settings = new ChangeSettingsRequestT();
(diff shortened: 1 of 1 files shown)


Louka3000:
Why do you want a way to unassign VRM models? There should not be any advantage to not having one loaded.

lamystere:
> Why do you want a way to unassign VRM models? There should not be any advantage to not having one loaded.

I was not able to switch to a different VRM model once one was loaded.  Or at least when I tried to the new model name was not displayed in place of the old.  Also, its minor, but the model json can take up a disproportionate amount of space in the settings YML.  Editing the YML directly was the only was I could be sure a new VRM model was being loaded.

Louka3000:
if your problem is that you somehow can’t assign a new model normally, that should be fixed directly. Not by having this workaround of removing a model before reassigning a new one.

and there’s no default VRM model assigned because VRM models don’t all have the same bone lengths :p

lamystere:
> if your problem is that you somehow can’t assign a new model normally, that should be fixed directly. Not by having this workaround of removing a model before reassigning a new one.

My point wasn't VRM issues, that was just an example story to highlight that it can be confusing, even for technical users, because the GUI is not informative.  A button provides additional insight.  The gui currently does not signal the success or failure of loading a new model.  This can't always be inferred by the title changing.  The CSS flash of the title provides a false sense of success since it happens regardless.

Louka3000:
So the issue is that there's no validation of the VRM model to decide if the title should flash or not?
Why not just do that, and show an error message if the validation fails?

lamystere:
> So the issue is that there's no validation of the VRM model to decide if the title should flash or not? Why not just do that, and show an error message if the validation fails?

That makes sense also.  It just seems strange to me as a workflow to be able to add something but not remove it.

ButterscotchV:
Resolved in #1246

PR #959, opened 2024-03-13
Make ktmath object-less

ImUrX:
This should improve performance in SlimeVR by making all math classes not be boxed into the heap.
* Implements an object wrapper for Quaternion so it's simpler to save in the Jackson config
* Converts classes that use Quaternion into Kotlin so they can compile
* Should check if config works correctly
* `HumanPoseManager` got a big rewrite by using `lateinit`, it was simplified by a lot, but also requires testing just in case.


Louka3000 (approved):
lgtm

server/core/src/main/java/com/jme3/math/FastMath.kt
ButterscotchV, line 1: There's way too many instances for me to note each one, please use the appropriately typed math functions where appropriate instead of converting everything to doubles.

server/core/src/main/java/io/github/axisangles/ktmath/Matrix3.kt
```diff
@@ -41,6 +45,16 @@ data class Matrix3(
...
 
+	operator fun component1(): Float = xx
+	operator fun component2(): Float = yx
+	operator fun component3(): Float = zx
+	operator fun component4(): Float = xy
+	operator fun component5(): Float = yy
+	operator fun component6(): Float = zy
+	operator fun component7(): Float = xz
+	operator fun component8(): Float = yz
+	operator fun component9(): Float = zz
```
ButterscotchV, line 56: What are these for?

server/core/src/main/java/io/github/axisangles/ktmath/Quaternion.kt
```diff
@@ -171,6 +174,11 @@ data class Quaternion(val w: Float, val x: Float, val y: Float, val z: Float) {
 	 **/
 	operator fun div(that: Quaternion): Quaternion = this * that.inv()
 
+	operator fun component1(): Float = w
+	operator fun component2(): Float = x
+	operator fun component3(): Float = y
+	operator fun component4(): Float = z
```
ButterscotchV, line 180: Same as for Matrix, what is this for?

server/core/src/main/java/io/github/axisangles/ktmath/Vector3.kt
```diff
@@ -15,6 +18,10 @@ data class Vector3(val x: Float, val y: Float, val z: Float) {
 		val NEG_Z = Vector3(0f, 0f, -1f)
 	}
 
+	operator fun component1() = x
+	operator fun component2() = y
+	operator fun component3() = z
```
ButterscotchV, line 23: Same as Matrix and Quaternion, what is this for?

ButterscotchV (changes requested):
I haven't really reviewed the Kotlin conversions since there is a lot to review, but please fix these issues. For the math functions, please ensure this is done for all converted classes.

server/core/src/main/java/io/github/axisangles/ktmath/Matrix3.kt
ImUrX: its for destructuring, its used in some places
```kt
val (_,x,y) = Quaternion(1,0,0,0)
```

ImUrX: It wasn't here before because data classes implement component destructuring by default

server/core/src/main/java/dev/slimevr/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,25 @@
...
+	abstract fun messagesCount(): Int
+
+	fun registerPacketListener(packetType: Byte, consumer: BiConsumer<GenericConnection, H>?) {
+		if (handlers[packetType.toInt()] != null) {
+			handlers[packetType.toInt()] =
+				consumer?.let { handlers[packetType.toInt()]!!.andThen(it) }
+		} else {
+			handlers[packetType.toInt()] = consumer
+		}
+	}
```
ButterscotchV, line 24: If consumer is null then it'll stop the existing handler from functioning. Why not just make it not nullable?
```suggestion
	fun registerPacketListener(packetType: Byte, consumer: BiConsumer<GenericConnection, H>) {
		val packetInt = packetType.toInt()
		if (handlers[packetInt] != null) {
			handlers[packetInt] = handlers[packetInt]!!.andThen(consumer)
		} else {
			handlers[packetInt] = consumer
		}
	}
```

ButterscotchV (changes requested):

server/core/src/main/java/dev/slimevr/config/VRConfig.kt
```diff
@@ -0,0 +1,133 @@
...
+
+	val filters: FiltersConfig = FiltersConfig()
+
+	val driftCompensation: DriftCompensationConfig = DriftCompensationConfig()
+
+	val oscRouter: OSCConfig = OSCConfig()
+
+	val vrcOSC: VRCOSCConfig = VRCOSCConfig()
+
+	val vMC: VMCConfig = VMCConfig()
```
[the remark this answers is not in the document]
ImUrX: no thats why its named like that, it can't be named `VMC` as that would be a constant naming and if we did `vmc` it would be `getVmc`

server/core/src/main/java/dev/slimevr/config/TrackerConfig.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.JsonNodeFactory
+import dev.slimevr.tracking.trackers.Tracker
+import io.github.axisangles.ktmath.ObjectQuaternion
+
+class TrackerConfig {
+	var customName: String? = null
+	var designation: String? = null
+	var isHide: Boolean = false
```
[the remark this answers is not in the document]
ImUrX: yes it will, but i can fix that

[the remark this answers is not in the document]
ImUrX: fixed

server/core/src/main/java/dev/slimevr/config/VRConfig.kt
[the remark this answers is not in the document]
ImUrX: now its better

server/core/src/main/java/dev/slimevr/config/TrackerConfig.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.JsonNodeFactory
+import dev.slimevr.tracking.trackers.Tracker
+import io.github.axisangles.ktmath.ObjectQuaternion
+
+class TrackerConfig {
+	var customName: String? = null
+	var designation: String? = null
+
+	@get:JvmName("isHide")
```
ButterscotchV, line 12: Does this not still affect serialization? I'm not sure about the logic with it, just make sure it still works I guess?

ImUrX: the original serialization was based on setHide and isHide, so it should work

ButterscotchV (approved):

PR #962, opened 2024-03-18
Add missing head tracker status

ImUrX:
* adds `isHmd` flag in `Tracker`
* make an error when there is available `isHmd` trackers and `headTracker` in main skeleton is null
* require reset on all assigned trackers when the `HumanSkeleton#headTracker` has changed


ImUrX:
Haven't tested it yet!

ImUrX:
i tested it and it seems to work, status system is a little finicky in the GUI side, sometimes packets are lost, it should be solved someday tbh

server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
```diff
@@ -694,5 +700,68 @@ class HumanPoseManager(val server: VRServer?) {
...
+		if (server == null ||
+			skeleton.headTracker == null ||
+			skeleton.headTracker?.isHmd != true
+		) {
+			return
+		}
+		server.allTrackers
+			.filter { !it.isInternal && it.trackerPosition != null }
+			.forEach {
+				// This can't be solved with a yaw reset, it requires a full reset
```
Erimelowo, line 719: This can be solved with a yaw reset

server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
```diff
@@ -694,5 +700,68 @@ class HumanPoseManager(val server: VRServer?) {
...
 	}
 
-	fun togglePauseTracking(sourceName: String?): Boolean = skeleton.togglePauseTracking(sourceName) // #endregion
+	fun togglePauseTracking(sourceName: String?): Boolean = skeleton.togglePauseTracking(sourceName)
+
+	// This should be executed when the head tracker is changed
+	fun checkTrackersRequiringReset() {
+		// Checks if this is main human pose manager (having server) or
+		// skeleton doesn't have a head tracker or not an HMD one
+		// TODO: Should this be ran anyways if headTracker is null or is not an HMD?
```
Erimelowo, line 709: It needs to be ran if headTracker is null (specifically when the previous headTracker was computed). Doesn't need to be ran if headTracker is not computed as IMU head trackers will be adjusted to Identity.

server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
```diff
@@ -694,5 +700,68 @@ class HumanPoseManager(val server: VRServer?) {
...
+		if (server == null ||
+			skeleton.headTracker == null ||
+			skeleton.headTracker?.isHmd != true
```
Erimelowo, line 712: Should probably check for isComputed instead
```suggestion
			skeleton.headTracker?.isComputed != true
```

Erimelowo (changes requested):

Erimelowo (approved):

ButterscotchV (approved):

PR #964, opened 2024-03-23
Add Yaw Reset Smoothing

nekomona:
Adding a configurable smoothing on yaw resets to allow a more natural and less noticeable yaw resets

Depending on: SlimeVR/SolarXR-Protocol#134


gui/public/i18n/en/translation.ftl
```diff
@@ -303,6 +303,7 @@ settings-general-tracker_mechanics-filtering-type-smoothing-description = Smooth
 settings-general-tracker_mechanics-filtering-type-prediction = Prediction
 settings-general-tracker_mechanics-filtering-type-prediction-description = Reduces latency and makes movements more snappy, but may increase jitter.
 settings-general-tracker_mechanics-filtering-amount = Amount
+settings-general-tracker_mechanics-yaw-reset-smooth-time = Yaw reset smooth time (0.0 for disable smoothing)
```
ButterscotchV, line 306:
```suggestion
settings-general-tracker_mechanics-yaw-reset-smooth-time = Yaw reset smooth time (0.0 disables smoothing)
```

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
```diff
@@ -222,12 +235,20 @@ class TrackerResetsHandler(val tracker: Tracker) {
...
+		yawFixOld = yawFix
 		yawFix = fixYaw(tracker.getRawRotation() * mountingOrientation, reference)
+		yawFixNew = yawFix
 
 		makeIdentityAdjustmentQuatsYaw()
 
 		calculateDrift(rot)
 
+		// Start at yaw before reset if smoothing enabled
+		if (yawResetSmoothTime != 0.0f) {
```
ButterscotchV, line 247:
```suggestion
		if (yawResetSmoothTime > 0.0f) {
```

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
```diff
@@ -432,6 +453,28 @@ class TrackerResetsHandler(val tracker: Tracker) {
...
+				deltaTime = fpsTimer.timePerFrame
+			}
+			yawResetSmoothTimeRemain = max(0.0f, yawResetSmoothTimeRemain - deltaTime)
+			if (yawResetSmoothTimeRemain == 0.0f) {
+				yawFix = yawFixNew
+			} else {
+				yawFix = yawFixOld.interpR(
+					yawFixNew, 
+					animateEase( (yawResetSmoothTime - yawResetSmoothTimeRemain) / yawResetSmoothTime)
+				)
```
ButterscotchV, line 473: This might make it a bit less readable, but it would be simpler to just ease "backwards". Maybe a comment could also be added to explain this?
```suggestion
				yawFix = yawFixNew.interpR(
					yawFixOld,
					animateEase(yawResetSmoothTimeRemain / yawResetSmoothTime)
				) 
```

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
```diff
@@ -432,6 +453,28 @@ class TrackerResetsHandler(val tracker: Tracker) {
...
+	 */
+	@Synchronized
+	fun update() {
+		if (yawResetSmoothTimeRemain != 0.0f) {
+			var deltaTime = 0.001f
+			if (::fpsTimer.isInitialized) {
+				deltaTime = fpsTimer.timePerFrame
+			}
+			yawResetSmoothTimeRemain = max(0.0f, yawResetSmoothTimeRemain - deltaTime)
+			if (yawResetSmoothTimeRemain == 0.0f) {
```
ButterscotchV, line 467: Can simplify this code and avoid some precision-based bugs
```suggestion
		if (yawResetSmoothTimeRemain > 0.0f) {
			var deltaTime = 0.001f
			if (::fpsTimer.isInitialized) {
				deltaTime = fpsTimer.timePerFrame
			}
			yawResetSmoothTimeRemain = yawResetSmoothTimeRemain - deltaTime
			if (yawResetSmoothTimeRemain <= 0.0f) {
```

ButterscotchV (changes requested):
I haven't looked in-depth, but I think there may be some usage of `yawFix` in other resets where it should be using the new one and not the interpolated one. I can look more in-depth for a later review.

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
nekomona: I'm more with having an accurate 0.0f as some sort of flag and assigning 0.0f would always be precise. Anyway, the sign could still be a boolean-ish flag so probably it's fine.

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt
nekomona: Probably yes, I'll add some comment here when solving conflicts

nekomona:
> I haven't looked in-depth, but I think there may be some usage of `yawFix` in other resets where it should be using the new one and not the interpolated one. I can look more in-depth for a later review.

There're some indirect calls to `yawFix`  in `resetFull` and `resetMounting`, I'll add some guards to clear smoothing before them

nekomona:
I've changed yaw smoothing to an additional step rather than tweaking `yawFix` to avoid interfering with other resets. Merging conflict is also resolved.

ButterscotchV (approved):
Would still like this tested before merge, but if I don't do it myself then we can just test in RCs

gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -636,6 +640,19 @@ export function GeneralSettings() {
...
               />
             </div>
+            <div className="flex gap-5 pt-5 md:flex-row flex-col">
+              <NumberSelector
+                control={control}
+                name="resetsSettings.yawResetSmoothTime"
+                label={l10n.getString(
+                  'settings-general-tracker_mechanics-yaw-reset-smooth-time'
+                )}
+                valueLabelFormat={(value) => `${Math.round(value * 20) / 20} s`}
```
ImUrX, line 650: Please use `Intl.NumberFormat` for formatting numbers:
```ts
const secondsFormat = new Intl.NumberFormat(currentLocales, {
  style: 'unit',
  maximumFractionDigits: 1,
  unit: 'second'
});
```

ImUrX (changes requested):

gui/src/components/settings/pages/GeneralSettings.tsx
nekomona: The formatter is changed along with the ones for reset delay timings -w-/

ImUrX (approved):

gui/public/i18n/en/translation.ftl
```diff
@@ -311,6 +311,7 @@ settings-general-tracker_mechanics-filtering-type-smoothing-description = Smooth
 settings-general-tracker_mechanics-filtering-type-prediction = Prediction
 settings-general-tracker_mechanics-filtering-type-prediction-description = Reduces latency and makes movements more snappy, but may increase jitter.
 settings-general-tracker_mechanics-filtering-amount = Amount
+settings-general-tracker_mechanics-yaw-reset-smooth-time = Yaw reset smooth time (0.0 disables smoothing)
```
Louka3000, line 314:
```suggestion
settings-general-tracker_mechanics-yaw-reset-smooth-time = Yaw reset smooth time (0s disables smoothing)
```

Louka3000 (changes requested):

gui/public/i18n/en/translation.ftl
Louka3000: It appears as 0s in GUI, not 0.0

nekomona: Ouch.

Louka3000 (approved):

PR #967, opened 2024-03-28
Tpose down fix

Louka3000:
TPose (down) requires the user to Full Reset in a tpose and Mounting Reset in a i-pose. It never really worked, even tho TPose (up) worked fine due to my sloppy implementation that wasn't tested thoroughly.

One problem was not testing twist, as I tested with SteamVR trackers.

The other problem is that all of our reset code is made with the assumption that trackers are pointing down below the horizon line (because of yaw or idk), so I probably got lucky on testing and it worked.

This fixes both of these issues.

I spent way too much time on this, so this is as far as I go. It works. I'm open to code improvements ^^


ButterscotchV (approved):
Looks fine, can't really confirm the math so I would like this to be tested properly before release.

ImUrX (approved):

PR #972, opened 2024-03-31
Add network protocol version

ImUrX:
makes firmware version be actual firmware version and adds a network protocol version for the firmwareBuild number


Eirenliel (approved):

ImUrX:
AH WAIT SOLARXR!

PR #982, opened 2024-04-06
Add support for left/right steamvr toggles

### LINKED ISSUE #786: Allow toggling left/right limb trackers
Separate left and right feet when advanced settings/dev mode is enabled in GUI. Shouldn’t require any server change afaik.

ImUrX:
Removes the original ones, it's too much work doing both of them in React, we need to refactor settings for it.

Also checks for feeder trackers so it doesn't replace them. (maybe could be removed on #920 merge)

Fixes #786 

depends on https://github.com/SlimeVR/SolarXR-Protocol/pull/138


Louka3000 (approved):

ButterscotchV (approved):
Make sure the hand warning popup still works, I didn't review that thoroughly

ImUrX:
it works but its lazily put together, settings is currently a mess. (when you enable one, the warning popups and turns both on, you can later disable the one you dont want tho)

PR #984, opened 2024-04-11
Delete old resources from java

ImUrX:
This deletes the license which was changed to a dual one (if we want to add both licenses to the zip, we could), and deletes run.bat which no longer works


Erimelowo:
> if we want to add both licenses to the zip, we could

I would like that

PR #991, opened 2024-04-14
macOS app naming improvements

i0ntempest:
Sets "SlimeVR" as app name in multiple locations. In the future this could be replaced by `displayName` if [tauri ever implement it](https://github.com/tauri-apps/tauri/issues/8109).

### DIFF
--- a/.github/workflows/gradle.yaml
+++ b/.github/workflows/gradle.yaml
@@ -263,16 +263,21 @@ jobs:
         run: |
           cd target/universal-apple-darwin/release/bundle/macos/slimevr.app/Contents/MacOS
           cp $( git rev-parse --show-toplevel )/server/desktop/build/libs/slimevr.jar ./
-          cd ../../../../dmg/
-          ./bundle_dmg.sh --volname slimevr --icon slimevr 180 170 --app-drop-link 480 170 \
-          --window-size 660 400 --hide-extension ../macos/slimevr.app \
-          --volicon ../macos/slimevr.app/Contents/Resources/icon.icns --skip-jenkins \
-          --eula ../../../../LICENSE-MIT slimevr.dmg ../macos/slimevr.app
+          cd ../../../
+          /usr/libexec/PlistBuddy -c "Set :CFBundleDisplayName SlimeVR" slimevr.app/Contents/Info.plist
+          /usr/libexec/PlistBuddy -c "Set :CFBundleName SlimeVR" slimevr.app/Contents/Info.plist
+          codesign --sign - --deep --force slimevr.app
+          mv slimevr.app SlimeVR.app
+          cd ../dmg/
+          ./bundle_dmg.sh --volname SlimeVR --icon slimevr 180 170 --app-drop-link 480 170 \
+          --window-size 660 400 --hide-extension ../macos/SlimeVR.app \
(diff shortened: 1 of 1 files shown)


ImUrX (changes requested):
The SlimeVR Server is not the GUI, and the app is specifically the GUI

Please change it to SlimeVR

i0ntempest:
> The SlimeVR Server is not the GUI, and the app is specifically the GUI

Okay, but the windows installer installs a “SlimeVR Server” shortcut into start menu that launches the GUI, so I wonder why is it not appropriate here? It’s the name for the package that contains both GUI and server.

ImUrX:
And we also have the linux desktop file which is just `SlimeVR`, we have the binary called `slimevr.exe` in Windows.

I don't really know the reason for why it's called `SlimeVR Server` in the start menu but I suspect it's been like that since the start and having to change it seems like too much problem.

ImUrX:
why are you signing the app

i0ntempest:
> why are you signing the app

Do you not want that? I was just about to say I added that because without a signature users are probably going to have problems launching the app because macOS Gatekeeper is only getting stricter.

ImUrX:
I mean, we don't have any signature to sign it with XD. Or does every macOS come with a user signature?

i0ntempest:
> I mean, we don't have any signature to sign it with XD. Or does every macOS come with a user signature?

I signed with `-` identity which is local adhoc signature. It's not a developer identity but should be better than flat out no signature.

Also idealy version of the app bundle should match github releases, right now I believe it's inherting the GUI version. Do you want that to be overwritten in Info.plist as well?

ImUrX:
then its fine to sign it.

the version stuff, it would be nice to fix the actual issue, the `package.json` of `gui` is not the same as the git tag and i don't really know how to fix it

i0ntempest:
> the version stuff, it would be nice to fix the actual issue, the `package.json` of `gui` is not the same as the git tag and i don't really know how to fix it

It's possible with `npm version`, but it's usually used to create/manage git tags, not the other way around (following existing git tags). I'm experimenting with it now, there should be a way around.

i0ntempest:
I kinda figured it out. Workflows triggered by releases will have their version in `gui/packages.json` corrected but that will also make the build dirty. To truely get around this I guess you really need to use `npm version` the way it meant to be used.

.github/workflows/gradle.yaml
```diff
@@ -256,23 +256,37 @@ jobs:
...
           rustup target add aarch64-apple-darwin
+          tag=${GITHUB_REF#refs/*/}
+          if [[ "$tag" == v* ]]; then
+            echo "Overwriting version: $tag"
+            cd gui
+            pnpm version "$tag" --force --allow-same-version --workspaces-update false
+            git add package.json
+            cd ..
+            pnpm version "$tag" --force --allow-same-version --workspaces-update false
+          fi
```
ImUrX, line 267:
```suggestion
```
i have another way to fix this, but it's gonna be in another PR.
Mind removing this? :3

ImUrX (changes requested):
let's leave this for another PR

ImUrX (approved):

PR #993, opened 2024-04-15
Check if saved window size is at least the minimum allowed

ImUrX:


Louka3000 (approved):
lgtm

gui/src-tauri/src/state.rs
```diff
@@ -73,7 +75,11 @@ impl WindowState {
 			window.unmaximize()?;
 		}
 
-		window.set_size(LogicalSize::new(self.width, self.height))?;
+		if self.width > util::MIN_WINDOW_SIZE_WIDTH
+			&& self.height > util::MIN_WINDOW_SIZE_HEIGHT
+		{
+			window.set_size(LogicalSize::new(self.width, self.height))?;
+		}
```
TheDevMinerTV, line 82: Why not just use `std::cmp::max()`?

ImUrX: the idea is to not set size if somehow it saved an incorrect one

TheDevMinerTV: nvm, you're right

Eirenliel (approved):

PR #998, opened 2024-04-21
Support Datagrams up to 1472 bytes

9il:
Gesture mobile app can send large datagrams because it combines information from multiple sensors.


TheDevMinerTV (approved):
LGTM, especially with bundling now being in the firmware and some people running more than four IMUs on a single ESP32.

Eirenliel (approved):

PR #999, opened 2024-04-21
Add more compact bundle protocol and IMU frame (rot+acc)

9il:


Eirenliel (approved):

ImUrX:
Bundle protocol exists as a bypass for old hardware that doesn't implement frame aggregation in their networking implementation (ESP8266).
I can see that the optimization of this implementation is only having a subset of packets being supported (as maximum package type is 255u and length is also 255u). But there is no firmware development which verifies that this actually is useful by some benchmarks and a mobile device wouldn't really be taking advantage of it because it already does via frame aggregation

9il:
@ImUrX What is the frame aggregation you are talking about?

ImUrX:
It's used in Wi-Fi protocols https://en.wikipedia.org/wiki/Frame_aggregation
Bundling was implemented in SlimeVR as ESP8266 doesn't implement frame aggregation.

9il:
OK, I see. However, the main point is to reduce the amount of data transferred over the air. This PR reduces it more than three times.

That isn't beneficial for slimes because they have relatively large batteries and don't need to aggregate data from dozens of sensors.

On the other hand, the volume of data transmitted over the air is a critical factor for mobile phones, primarily due to power consumption concerns. For instance, iOS has a policy to terminate any background app that consumes some energy. Therefore, if an iPhone is required to act as a proxy, the data transfer needs to be carefully managed.

The same thing happens for some Android vendors as well.

Mocopi example can't be used here because they have only six sensors and a lower frame rate.

ImUrX:
Please split your PR into two, having the IMU packet in another PR and the simplified bundle protocol in another then.

9il:
@ImUrX The new packed is here #1001 now

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/FeatureFlags.kt
```diff
@@ -43,6 +43,7 @@ class FirmwareFeatures {
 enum class ServerFeatureFlags {
 	/** Server can parse bundle packets: `PACKET_BUNDLE` = 100 (0x64). */
 	PROTOCOL_BUNDLE_SUPPORT,
+	PROTOCOL_BUNDLE_V2_SUPPORT,
```
0forks, line 46: Overall LGTM, but could use some docs, something like this would be more descriptive I think:
```suggestion
	/** Server can parse bundle packets with compact headers and packed IMU rotation/acceleration frames:
		- `PACKET_BUNDLE_COMPACT` = 101 (0x65),
		- `PACKET_ROTATION_AND_ACCELERATION` = 23 (0x17). */
	PROTOCOL_BUNDLE_COMPACT_SUPPORT,
```

0forks (changes requested):

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/FeatureFlags.kt
9il: @0forks done, thank you

0forks (approved):

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPProtocolParser.kt
```diff
@@ -40,6 +40,25 @@ class UDPProtocolParser {
 				buf.position(bundlePacketStart + bundlePacketLen)
 			}
 			return bundlePackets.toTypedArray()
+		} else if (packetId == PACKET_BUNDLE_V2) {
```
0forks, line 43: These too please, for consistency with the flag docs.
```suggestion
		} else if (packetId == PACKET_BUNDLE_COMPACT) {
```

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPProtocolParser.kt
```diff
@@ -129,6 +148,7 @@ class UDPProtocolParser {
 		const val PACKET_FEATURE_FLAGS = 22
 		const val PACKET_ROTATION_AND_ACCELERATION = 23
 		const val PACKET_BUNDLE = 100
+		const val PACKET_BUNDLE_V2 = 101
```
0forks, line 151:
```suggestion
		const val PACKET_BUNDLE_COMPACT = 101
```

0forks (approved):

9il:
OK, CI passes, please merge it :)

ImUrX:
Sorry but we are releasing 0.12 without this PR, we have a release candidate system and it's been almost a month since we released the first one.

We don't really want to do another release candidate because of this.

ImUrX:
Can you update the PR?

9il:
Do you mean an update from the main? Done.

PR #1000, opened 2024-04-21
Fix the HID decoding, replace Q14 with Q15, optimize

9il:
The commit [Unify quaternion order with UDP protocol](https://github.com/SlimeVR/SlimeVR-Server/pull/1000/commits/caffc40ed4b2754e9c666c9ead50a616e09385ee) is optional. However, I added it because the `xyzw` order is already used in UDP. Furthermore, that will unify the data encoding with #1001.

UPDATE: corresponding firmware PR https://github.com/SlimeVR/SlimeNRF-Tracker/pull/3

### DIFF
--- a/server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
+++ b/server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
@@ -162,6 +162,8 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 	private fun dataRead() {
 		synchronized(devicesByHID) {
 			var devicesPresent = false
+			val q = intArrayOf(0, 0, 0, 0)
+			val a = intArrayOf(0, 0, 0)
 			for ((hidDevice, deviceList) in devicesByHID) {
 				val dataReceived: Array<Byte> = try {
 					hidDevice.read(80, 1) // Read up to 80 bytes, timeout 1ms
@@ -182,20 +184,17 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 						val idCombination = dataReceived[i + 1].toInt()
 						val rssi = -dataReceived[i + 2].toInt()
 						val battery = dataReceived[i + 3].toInt()
-						val battery_mV = dataReceived[i + 5].toInt() and 255 shl 8 or (dataReceived[i + 4].toInt() and 255)
-						val q = floatArrayOf(0f, 0f, 0f, 0f)
-						val a = floatArrayOf(0f, 0f, 0f)
-						for (j in 0..3) { // quat received as fixed 14
-							var buf =
-								dataReceived[i + 6 + j * 2 + 1].toInt() and 255 shl 8 or (dataReceived[i + 6 + j * 2].toInt() and 255)
-							buf -= 32768 // uint to int
-							q[j] = buf / (1 shl 14).toFloat() // fixed 14 to float
+						// ushort big endian
+						val battery_mV = dataReceived[i + 5].toUByte().toInt() shl 8 or dataReceived[i + 4].toUByte().toInt()
+						// Q15: 1 is represented as 0x7FFF, -1 as 0x8000
+						// The sender can use integer saturation to avoid overflow
+						for (j in 0..3) { // quat received as fixed Q15
+							// Q15 as short big endian
+							q[j] = dataReceived[i + 6 + j * 2 + 1].toInt() shl 8 or dataReceived[i + 6 + j * 2].toUByte().toInt()
 						}
 						for (j in 0..2) { // accel received as fixed 7, in m/s
-							var buf =
-								dataReceived[i + 14 + j * 2 + 1].toInt() and 255 shl 8 or (dataReceived[i + 14 + j * 2].toInt() and 255)
-							buf -= 32768 // uint to int
-							a[j] = buf / (1 shl 7).toFloat() // fixed 7 to float
+							// Q15 as short big endian
+							a[j] = dataReceived[i + 14 + j * 2 + 1].toInt() shl 8 or dataReceived[i + 14 + j * 2].toUByte().toInt()
 						}
 						val trackerId = idCombination and 0b1111
 						val deviceId = (idCombination shr 4) and 0b1111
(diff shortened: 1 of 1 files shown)


server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -208,11 +207,17 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 						// tracker.batteryVoltage = if (battery and 128 == 128) 5f else 0f // Charge status
 						tracker.batteryVoltage = battery_mV.toFloat() * 0.001f
 						tracker.batteryLevel = (battery and 127).toFloat()
-						var rot = Quaternion(q[0], q[1], q[2], q[3])
-						rot = AXES_OFFSET.times(rot)
+						// The data comes in the same order as in the UDP protocol
+						// x y z w -> w x y z
+						var rot = Quaternion(q[3].toFloat(), q[0].toFloat(), q[1].toFloat(), q[2].toFloat())
```
sctanf, line 212: Existing HID devices sent rotation as WXYZ already, is it important to match UDP protocol here?

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -182,20 +184,17 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
...
-						val q = floatArrayOf(0f, 0f, 0f, 0f)
-						val a = floatArrayOf(0f, 0f, 0f)
-						for (j in 0..3) { // quat received as fixed 14
-							var buf =
-								dataReceived[i + 6 + j * 2 + 1].toInt() and 255 shl 8 or (dataReceived[i + 6 + j * 2].toInt() and 255)
-							buf -= 32768 // uint to int
-							q[j] = buf / (1 shl 14).toFloat() // fixed 14 to float
+						// ushort big endian
+						val battery_mV = dataReceived[i + 5].toUByte().toInt() shl 8 or dataReceived[i + 4].toUByte().toInt()
+						// Q15: 1 is represented as 0x7FFF, -1 as 0x8000
```
sctanf, line 189: Originally used Q14 to not worry about representing 1 but it is probably not important. You cannot really represent exactly 1 with Q15 format, but the nearest value is good enough

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
9il: It is not critical. However,
 - We change Q14 to Q15 anyway, so the firmware is subject to change.
 - XYZW has a minor performance advantage for SIMD optimizations compared to WXYZ
 - And most of the major APIs consider XYZW as standard:
   -  [.NET](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.quaternion.-ctor?view=net-8.0#system-numerics-quaternion-ctor(system-single-system-single-system-single-system-single))
   - [Unreal Engine](https://docs.unrealengine.com/4.27/en-US/API/Runtime/Core/Math/FQuat/)
   - [Apple](https://developer.apple.com/documentation/accelerate/working_with_quaternions)

Also, the unification with UDP will help to simplify firmware logic with #1001.

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
9il: The quaternion `(0, 0, 1, 0)` will be encoded to `(0, 0, 7FFF, 0)`. If one decode it and normalize it will be `(0, 0, 1, 0)` up to a single ULP.

Q15 is twice (one bit) more precise for all values except for four values like `(0, 0, 1, 0)`.

Plus, Q15 is a standard for `[-1, 1]` values encoding in 16-bits.

And again, unification with #1001.

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
9il: Please note that documentation sometimes list properties/members in WXYW 
image
because of alphabetical sorting.

What we need to look to is constructors:
image

9il:
I have added accompanying MR for firmware https://github.com/SlimeVR/SlimeNRF-Tracker/pull/3

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
ImUrX: > XYZW has a minor performance advantage for SIMD optimizations compared to WXYZ
any reference pls?

> And most of the major APIs consider XYZW as standard: 
yeah in APIs they use XYZW but in actual math its WXYZ

ImUrX: > Also, the unification with UDP will help to simplify firmware logic with https://github.com/SlimeVR/SlimeVR-Server/pull/1001.

this doesnt matter because UDP logic and HID logic are separated, are we expecting in the future to somehow unify parsing logic?

9il: > any reference pls?

It is simple. Some operations need to work independently with XYZ and W.
For W, it doesn't matter where it is because it is a scalar. For XYZ, it matters in the `float4`lready; SIMD register a then it can de facto be used in `float3` SIMD library functions (if they are clever, they will use a single `float4` for most architectures instead of `float2` + `float`) because if it s. But, if the encoding is WXYZ, then the code needs to add a few assembler instructions to bring XYZ to the front of `float4`. That especially matters in heavily optimized operations with a lot of math because the number of SIMD registers may be relatively small for them. Compilers may fail to optimize the SIMD rotation of heavily optimized code, and that sometimes makes the code a few times slower, mostly because the amount of move operations exceeds the amount of floating-point operations.

> yeah in APIs they use XYZW but in actual math its WXYZ

Some math use WXYZ because of legacy code that was born without SIMD in mind.

> this doesnt matter because UDP logic and HID logic are separated, are we expecting in the future to somehow unify parsing logic?

The logic is unified in Gestures software. Although the fact that the sensors aren't slimes, I think it would be beneficial for both sides to make them work with NRF dongle and SlimeVR-Server.

9il: I don't have a link to some proof of my words. However, the following link help to ensure that I know what I am talking about https://github.com/libmir/mir-glas

9il: > Some operations

For example, a vector rotation.

ImUrX (approved):

sctanf (approved):

Louka3000 (approved):

ImUrX:
pls update

9il:
> pls update

done

PR #1001, opened 2024-04-21
Add compact IMU frame (rot+acc)

9il:


ImUrX (approved):

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPProtocolParser.kt
```diff
@@ -127,6 +127,7 @@ class UDPProtocolParser {
 		const val PACKET_TEMPERATURE = 20
 		const val PACKET_USER_ACTION = 21
 		const val PACKET_FEATURE_FLAGS = 22
+		const val PACKET_IMU_FRAME = 23
```
TheDevMinerTV, line 130: I'd change this to `PACKET_ROTATION_AND_ACCELERATION` for consistencies sake.

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPPacket.kt
```diff
@@ -329,6 +329,25 @@ class UDPPacket22FeatureFlags(
 	}
 }
 
+data class UDPPacket23IMUFrame(
+	override var rotation: Quaternion = Quaternion.IDENTITY,
+	var acceleration: Vector3 = Vector3.NULL,
+) : UDPPacket(23),
+	RotationPacket {
+	override var sensorId: Int = 0
+	override fun readData(buf: ByteBuffer) {
```
TheDevMinerTV, line 338:
```suggestion
	override fun readData(buf: ByteBuffer) {
		// s16 s16 s16 s16 s16 s16 s16
		// qX  qY  qZ  qW  aX  aY  aZ
```

TheDevMinerTV (approved):
The code looks fine, but having an actual implementation in the official firmware would be amazing.

TheDevMinerTV (changes requested):
wrong button

9il:
@TheDevMinerTV I have renamed the packet as you suggested.

| The code looks fine, but having an actual implementation in the official firmware would be amazing.

Unfortunately, our code isn't compatible with the SlimeVR firmware due to its reliance on a special slime extension of Mir Ion, a Dlang serialization library.

However, it will be encoding will be very similar to HID after #1000 . BTW, I think the quaternion order in the HID protocol is different. Will update #1000 then.

TheDevMinerTV (approved):
LGTM

Eirenliel (approved):

PR #1004, opened 2024-04-22
Create tarball of GUI bundle on release

ImUrX:
This works to simplify packaging in other places as the HTML/CSS/JS code is portable and works everywhere, it doesn't need to be edited.


TheDevMinerTV (approved):
LGTM

Erimelowo (approved):

PR #1009, opened 2024-04-25
GUI - Tracker assignment options

ZycaR:
- Added multiple tracker assignment options according common use cases.
- In case there is specific number of trackers connected, 
the options with 'less' trackers are disabled and first available option is selected.

Available Options (feel free to suggest better options or wordings):
**x5 Lower-Body Set:** chest, upper & lower legs
**x6 Core Set:**  chest, **hip**, upper & lower legs
**x8 Enhanced Core Set:** chest, hip, upper & lower legs,  **foots**
**x10 Full Body Tracking (FBT):** chest, hip, **upper arms**, upper & lower legs,  foots

image

ability to select number of trackers (old version):
tracker-options


ZRock35:
Some observations:
The Lower Body 5 set has waist, when it should be assigned as chest for best tracking. Virtual waist kicks in, and sitting and laying down are far better in comparison to having it set as waist.
6 is the most commonly recommended set, and is not present in this, and waist is not the recommended assignment, hip is.
8 "enhanced set of core trackers" is not quite the right wording, since you skipped putting the "core" set it now doesn't have context.
10 is not "universal" nor would I say it's suitable for most users, it's kind of an advanced set, for dancers most often.
13 how it is pictured there doesn't make sense for VRChat, as VRChat doesn't use lower arm trackers. 13 tracker setup in VRChat would be adding the waist tracker, then either shoulders OR upper chest + neck tracker for additional torso stability.
20 is not good for VRChat at all and will cause major problems such as controllers not working. It is strictly for mocap/vtubing.
There are a lot more details, but figured I'd at least mention those things.

In my honest opinion, although I think this is neat and it's been suggested in the discord before, it would be better to have a slider that goes from 1 up to 20, that fills in at the proper points, as you slide it.

sctanf:
assuming at this point the user has connected all their possible trackers, you could use the number of trackers and extensions connected as a hint to what tracking assignment is recommended

ZycaR:
@ZRock35 Thank you for your input, especially regarding my English, which ( as you probably noticed) is not my strong suit.

I had an idea to introduce these selectable options because first-timers are often confused about where to mount their trackers.
In this PR, I'll aim to update the options according to the official sets: x5, **x6**, x8, x10, **x16**, and all x20 trackers.

To have a slider instead of options is a great idea. One significant advantage is the better granularity of options; a slider saves screen space; and last but not least, we can cut some corners on translations.
I'll do my best with research (on Discord) to find recommended configurations for the slider and prepare something if it's still a requested feature.

ZycaR:
Could you please take a look at the options section for the number of trackers?
I'm particularly interested in your feedback on the options themselves and the wording used for each one.

gui/src/components/onboarding/pages/trackers-assign/TrackerAssignment.tsx
```diff
@@ -46,27 +48,54 @@ interface FlatDeviceTrackerDummy {
...
   const { config, setConfig } = useConfig();
-  const { useAssignedTrackers, trackers } = useTrackers();
+  const { useAssignedTrackers, useConnectedIMUTrackers, trackers } =
+    useTrackers();
   const { applyProgress, state } = useOnboarding();
   const { sendRPCPacket, useRPCPacket } = useWebsocketAPI();
   const defaultValues = {
-    advanced: config?.advancedAssign ?? false,
+    assignMode: config?.assignMode ?? AssignMode.Core,
     mirrorView: config?.mirrorView ?? true,
```
ImUrX, line 73:
```suggestion
    assignMode: config?.assignMode ?? defaultConfig.assignMode,
    mirrorView: config?.mirrorView ?? defaultConfig.mirrorView,
```

gui/src/components/onboarding/pages/trackers-assign/TrackerAssignment.tsx
```diff
@@ -46,27 +48,54 @@ interface FlatDeviceTrackerDummy {
...
+
+    const selectedAssignMode =
+      (Object.entries(ASSIGN_MODE_OPTIONS).find(
+        ([_, count]) => count >= connectedIMUTrackers
+      )?.[0] as AssignMode) ?? AssignMode.All;
+
+    if (assignMode !== selectedAssignMode) {
+      setValue('assignMode', selectedAssignMode);
+    }
+  }, [connectedIMUTrackers]);
```
ImUrX, line 98: shouldn't the effect have `assignMode` as a dependency if it checks it?
```suggestion
  }, [connectedIMUTrackers, assignMode]);
```

gui/src/components/onboarding/pages/trackers-assign/TrackerAssignment.tsx
```diff
@@ -290,13 +319,44 @@ export function TrackersAssignPage() {
...
+                      disabled={
+                        connectedIMUTrackers > trackersCount &&
+                        mode !== AssignMode.All
+                      }
+                      className="hidden"
+                    >
+                      <div className="flex flex-row md:gap-4 sm:gap-2 mobile:gap-2">
+                        <div style={{ width: '2.5rem', textAlign: 'right' }}>
+                          <Typography variant="mobile-title">
+                            {`x${trackersCount}`}
```
ImUrX, line 339: use a translation string for this and use plural rules for it like
```ftl
onboarding-assign_trackers-option-amount = { $trackersCount ->
    [one] x{ $trackersCount }
    *[other] x{ $trackersCount }
}
```

gui/public/i18n/en/translation.ftl
```diff
@@ -677,6 +677,20 @@ onboarding-assign_trackers-assigned = { $assigned } of { $trackers ->
...
+    [full-body] Full Body Tracking (FBT)
+    *[all] Complete Set
+}
+onboarding-assign_trackers-option_descriptions = { $mode ->
+    [lower-body] Set of trackers for lower body tracking
+    [core] Most commonly recommended set of trackers
+    [enhanced-core] Enhanced set of core trackers
+    [full-body] Advanced set of trackers, for dancers
+    *[all] Complete set of all available trackers
+}
```
ImUrX, line 693: probably rename these to be like `option-label` and `option-description`, idk why they are in plural

ImUrX (changes requested):

gui/src/components/onboarding/BodyAssignment.tsx
```diff
@@ -94,6 +143,10 @@ export function BodyAssignment({
   const left = +!mirror;
   const right = +mirror;
 
+  const hasBodyPart = (part: BodyPart) =>
+    assignMode === AssignMode.All ||
+    ASSIGNMENT_MODES[assignMode].indexOf(part) > -1;
```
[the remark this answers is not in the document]
ZycaR: this is a good point, thanks!

gui/public/i18n/en/translation.ftl
ZycaR: It is my fault. I didn't realize that it's not enum, but name for translation record/entry.
fixed.

ImUrX (approved):

ImUrX:
yeah dont worry about that, just that most maintainers are away :p

ZRock35:
A complete set assigned, of all 20 Slimes, will conflict with an active SteamVR headset, and should probably be blocked so it can't accidentally be done - it might work, but have a popup that gives a warning about that or make it have an option to use 19.

Erimelowo:
The server will prioritize head trackers with positional data (e.g. HMDs) so it’s not a problem.

ZRock35:
Relevant PR about the HMD+SlimeVR head tracker bug: https://github.com/SlimeVR/SlimeVR-Server/issues/1040

Erimelowo (changes requested):
This happens to hide the warnings. I'd move the warning above the buttons and add a scrollbar on the page.

slimevr_GqftJKvSPp

ZycaR:
> This happens to hide the warnings. I'd move the warning above the buttons and add a scrollbar on the page.
> 
> slimevr_GqftJKvSPp

The latest commit should reflect your request. I was able to relocate the warning message and add a scrollbar without visual changes to the other onboarding pages.
Warnirng Scrollbar

Erimelowo (approved):

ImUrX:
can u please fix conflicts?

ImUrX (approved):

PR #1013, opened 2024-04-29
don't toggle trackers if tracking is paused

Louka3000:


server/core/src/main/java/dev/slimevr/VRServer.kt
```diff
@@ -308,12 +308,26 @@ class VRServer @JvmOverloads constructor(
...
+	fun getPauseTracking(): Boolean = humanPoseManager.getPauseTracking()
+
 	fun setPauseTracking(pauseTracking: Boolean, sourceName: String?) {
-		queueTask { humanPoseManager.setPauseTracking(pauseTracking, sourceName) }
+		queueTask {
+			humanPoseManager.setPauseTracking(pauseTracking, sourceName)
+			// Toggle trackers as they don't toggle when tracking is paused
+			if (this.getVRBridge(ISteamVRBridge::class.java)?.updateShareSettingsAutomatically() == true) {
+				RPCSettingsHandler.sendSteamVRUpdatedSettings(protocolAPI, protocolAPI.rpcHandler)
+			}
```
ImUrX, line 319: should this be done before or after the set pause tracking

ImUrX (approved):

ButterscotchV (approved):

server/core/src/main/java/dev/slimevr/VRServer.kt
Louka3000: After

PR #1014, opened 2024-04-29
Fix acceleration axis and leg tweaks

### LINKED ISSUE #1006: Acceleration decoding doesn't respect AXES_OFFSET
SlimeVR do convert rotation from tracker to openGL AXIS, but forget to do so for acceleration vector.

9il:
Fixes #1006

This PR fixes the axis for leg tweaks! However, an additional update of thresholds may be required.

@ImUrX, please test it in standard mode with leg tweaks like skating correction.

I don't have slimes for testing.

image

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
@@ -6,6 +6,7 @@ import dev.slimevr.tracking.trackers.TrackerPosition.Companion.getByDesignation
 import dev.slimevr.tracking.trackers.udp.IMUType
 import dev.slimevr.tracking.trackers.udp.MagnetometerStatus
 import dev.slimevr.tracking.trackers.udp.TrackerDataType
+import io.eiren.math.FloatMath.INV_SQRT_TWO
 import io.eiren.util.BufferedTimer
 import io.github.axisangles.ktmath.Quaternion
 import io.github.axisangles.ktmath.Vector3
@@ -435,5 +436,14 @@ class Tracker @JvmOverloads constructor(
 	 */
 	fun resetFilteringQuats(reference: Quaternion) {
 		filteringHandler.resetMovingAverage(getAdjustedRotation(), reference)
+	}
+
+	companion object {
+		/**
+		 * Changes from IMU axis to OpenGL/SteamVR axis
+		 */
+		fun axisOffset(v: Vector3): Vector3 = Vector3(v.x, v.z, -v.y)
+		private val rotationOffset = Quaternion(INV_SQRT_TWO, -INV_SQRT_TWO, 0f, 0f)
+		fun axisOffset(q: Quaternion): Quaternion = rotationOffset * q
 	}
 }
--- a/server/core/src/main/java/io/eiren/math/FloatMath.kt
+++ b/server/core/src/main/java/io/eiren/math/FloatMath.kt
@@ -17,7 +17,7 @@ object FloatMath {
 	const val ZERO_TOLERANCE_D: Double = 0.0001
 
 	val SQRT_TWO: Float = kotlin.math.sqrt(2.0).toFloat()
-	val INV_SQRT_TWO: Float = 1f / SQRT_TWO
+	val INV_SQRT_TWO: Float = (1.0 / kotlin.math.sqrt(2.0)).toFloat()
 	val SQRT_THREE: Float = kotlin.math.sqrt(3.0).toFloat()
 	val INV_SQRT_THREE: Float = 1f / SQRT_THREE
 	const val TWO_FPI: Float = PI * 2
--- a/server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
+++ b/server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
@@ -1,9 +1,9 @@
 package dev.slimevr.desktop.tracking.trackers.hid
 
-import com.jme3.math.FastMath
 import dev.slimevr.VRServer
 import dev.slimevr.tracking.trackers.Device
 import dev.slimevr.tracking.trackers.Tracker
+import dev.slimevr.tracking.trackers.Tracker.Companion.axisOffset
 import dev.slimevr.tracking.trackers.TrackerStatus
 import dev.slimevr.tracking.trackers.udp.BoardType
 import dev.slimevr.tracking.trackers.udp.IMUType
@@ -11,7 +11,6 @@ import dev.slimevr.tracking.trackers.udp.MCUType
 import dev.slimevr.tracking.trackers.udp.MagnetometerStatus
 import io.eiren.util.logging.LogManager
 import io.github.axisangles.ktmath.Quaternion
-import io.github.axisangles.ktmath.Quaternion.Companion.fromRotationVector
 import io.github.axisangles.ktmath.Vector3
 import org.hid4java.HidDevice
 import org.hid4java.HidException
@@ -400,7 +399,7 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 							// x y z w -> w x y z
 							var rot = Quaternion(q[3].toFloat(), q[0].toFloat(), q[1].toFloat(), q[2].toFloat())
 							val scaleRot = 1 / (1 shl 15).toFloat() // compile time evaluation
-							rot = AXES_OFFSET.times(scaleRot).times(rot) // no division
+							rot = axisOffset(rot * scaleRot) // no division
 							tracker.setRotation(rot)
 						}
 						if (packetType == 2) {
@@ -419,13 +418,14 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 							val s = sin(a)
 							val k = s * invSqrtD
 							var rot = Quaternion(cos(a), k * v[0], k * v[1], k * v[2])
-							rot = AXES_OFFSET.times(rot) // no division
+							rot = axisOffset(rot) // no division
 							tracker.setRotation(rot)
 						}
 						if (packetType == 1 || packetType == 2) {
 							// TODO: Acceleration "was" wrong
 							val scaleAccel = 1 / (1 shl 7).toFloat() // compile time evaluation
 							var acceleration = Vector3(a[0].toFloat(), a[1].toFloat(), a[2].toFloat()).times(scaleAccel) // no division
+							acceleration = axisOffset(acceleration)
 							tracker.setAcceleration(acceleration)
 							tracker.dataTick() // only data tick if there is rotation data
 						}
@@ -492,10 +492,6 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 	}
 
 	companion object {
-		/**
-		 * Change between IMU axes and OpenGL/SteamVR axes
-		 */
-		private val AXES_OFFSET = fromRotationVector(-FastMath.HALF_PI, 0f, 0f)
 		private const val resetSourceName = "TrackerServer"
 	}
 }
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ local.properties
 
 # Ignore temporary config
 vrconfig.yml.tmp
+*.DS_Store
--- a/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
@@ -1,16 +1,14 @@
 package dev.slimevr.tracking.trackers.udp
 
-import com.jme3.math.FastMath
 import dev.slimevr.NetworkProtocol
 import dev.slimevr.VRServer
 import dev.slimevr.config.config
 import dev.slimevr.protocol.rpc.MAG_TIMEOUT
 import dev.slimevr.tracking.trackers.*
+import dev.slimevr.tracking.trackers.Tracker.Companion.axisOffset
 import io.eiren.util.Util
 import io.eiren.util.collections.FastList
 import io.eiren.util.logging.LogManager
-import io.github.axisangles.ktmath.Quaternion.Companion.fromRotationVector
-import io.github.axisangles.ktmath.Vector3
 import kotlinx.coroutines.*
 import solarxr_protocol.rpc.ResetType
 import java.net.DatagramPacket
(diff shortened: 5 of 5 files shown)


server/core/src/main/java/io/eiren/math/FloatMath.kt
```diff
@@ -17,7 +17,7 @@ object FloatMath {
 	const val ZERO_TOLERANCE_D: Double = 0.0001
 
 	val SQRT_TWO: Float = kotlin.math.sqrt(2.0).toFloat()
-	val INV_SQRT_TWO: Float = 1f / SQRT_TWO
+	val INV_SQRT_TWO: Float = (1.0 / kotlin.math.sqrt(2.0)).toFloat()
```
Stermere, line 20: Why change this?

Stermere (commented):
This won't affect skating-correction since the magnitude of acceleration is not being changed. The logic looks good as far as I can tell

server/core/src/main/java/io/eiren/math/FloatMath.kt
9il: `1f / SQRT_TWO` guarantees the error is less or equal to 1.5 ULP.

`(1.0 / kotlin.math.sqrt(2.0)).toFloat()` guarantees the error is less or equal to 0.5 ULP.

9il:
> This won't affect skating-correction since the magnitude of acceleration is not being changed.

Not really. Some code uses only the two components, like `Vector3(footDif.x, 0f, footDif.z) `. It uses `x` and `z` components, which correspond to trackers to original (`x` `y`) components in this PR and trackers (`y` `z`) in the current `main` branch. `z` in the tracker coordinates corresponds to the gravity direction. That is what the fix is about.

```kotlin
	private fun correctUnlockedFootTracker(footPosition: Vector3, previousFootPosition: Vector3, previousFootPositionCorrected: Vector3, footVelocity: Vector3, framesUnlocked: Int): Vector3 {
		var newFootPosition = footPosition
		var footDif = footPosition - previousFootPositionCorrected
		footDif = Vector3(footDif.x, 0f, footDif.z)
```

server/core/src/main/java/io/eiren/math/FloatMath.kt
Stermere: Cool I thought it would be something like that

Stermere:
> > This won't affect skating-correction since the magnitude of acceleration is not being changed.
> 
> Not really. Some code uses only the two components, like `Vector3(footDif.x, 0f, footDif.z) `. It uses `x` and `z` components, which correspond to trackers to original (`x` `y`) components in this PR and trackers (`y` `z`) in the current `main` branch. `z` in the tracker coordinates corresponds to the gravity direction. That is what the fix is about.
> 
> ```kotlin
> 	private fun correctUnlockedFootTracker(footPosition: Vector3, previousFootPosition: Vector3, previousFootPositionCorrected: Vector3, footVelocity: Vector3, framesUnlocked: Int): Vector3 {
> 		var newFootPosition = footPosition
> 		var footDif = footPosition - previousFootPositionCorrected
> 		footDif = Vector3(footDif.x, 0f, footDif.z)
> ```

That is a positional correction, the acceleration is not used like that, I just checked. Mocap mode does use the acceleration direction though so this will help that!

Stermere (approved):

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -236,13 +235,15 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 						// x y z w -> w x y z
 						var rot = Quaternion(q[3].toFloat(), q[0].toFloat(), q[1].toFloat(), q[2].toFloat())
 						val scaleRot = 1 / (1 shl 15).toFloat() // compile time evaluation
-						rot = AXES_OFFSET.times(scaleRot).times(rot) // no division
+						rot = rot.times(scaleRot) // no division
```
ImUrX, line 238: use operators?
```suggestion
						rot *= scaleRot // no division
```

ImUrX (changes requested):

ImUrX (approved):

9il:
@ImUrX is anything else required for this PR to be merged?

ImUrX:
You need to wait for more reviewers

9il:
I though that two reviwers are enough. Should we ping someone to make a review?

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -236,13 +235,15 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 						// x y z w -> w x y z
 						var rot = Quaternion(q[3].toFloat(), q[0].toFloat(), q[1].toFloat(), q[2].toFloat())
 						val scaleRot = 1 / (1 shl 15).toFloat() // compile time evaluation
-						rot = AXES_OFFSET.times(scaleRot).times(rot) // no division
+						rot *= scaleRot // no division
+						rot = axisOffset(rot)
```
ButterscotchV, line 239: Why not just do it in one line
```suggestion
						rot = axisOffset(rot * scaleRot) // no division
```

ButterscotchV (approved):

Erimelowo (changes requested):
code itself looks good but testing with official SlimeVR trackers, looking at the values on the GUI (not the preview), it still seems like the original way was the right way (`Vector3(v.y, v.x, v.z)`). Maybe it'd be worth looking in the firmware if the acceleration is transformed. 

However, [the accel arrows preview](https://github.com/SlimeVR/SlimeVR-Server/blob/e2a511b552d52ebc77da0cb029440600edd5ddfb/gui/src/components/widgets/IMUVisualizerWidget.tsx#L57) and [the local -> global conversion for accel](https://github.com/SlimeVR/SlimeVR-Server/blob/e2a511b552d52ebc77da0cb029440600edd5ddfb/server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerResetsHandler.kt#L149) are both broken in some way

9il:
@Stermere id don't know if the fusion does the order right, but, at least, they in the right order during [packet formation](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/3ae17abdf478cc88b4d1598edc14af0c61f056e7/src/network/connection.cpp#L239C1-L244C1).

If you like I could take the original `AXES_OFFSET` and apply it both to rotation and acceleration. So, it will be defined by mostly by software, not a math.

On the other hand, if the bug is in the fustion, than the firmware has to be fixed instead of tuning the server for the firmware. At least, we could define a new version of protocol. The reason is because slime wants to be a kind of opensource standard for VR.

The current acceleratoin offsets is defenetely wrong. Some code may work better now because it was tuned for the current wrong offsets. That could work for now but it can be so for a robust code evolution.

Eirenliel:
We should check firmware for all main IMUs and make sure that it sends acceleration in the same axises, I guess.

9il:
both bugs may be on the server side as well.

Shade-emry:
@9il I am interested in testing this, I had a college do some research into it. Where is this github at that contains this fix? Or has this already been fixed and the issue is not closed?

sctanf:
Hi @Shade-emry, there are currently other issues as mentioned which should be resolved first.

9il:
Hi folks. I can update this PR if you confirm it that nothing else is blocking it. We don' use slime firmware, and can't fix it.

Eirenliel:
This fix loos correct. We should confirm that it doesn't break legtweaks, @Stermere @ButterscotchV and we can merge it after 0.14.0 release

9il:
rebased

9il:
@Eirenliel @sctanf Do I need to made another rebase? when it will be merged?

sctanf:
Front may not be a good descriptor
image

9il:
@sctanf Did you post a hardware tracker axis or a slime VR axis?
Because this picture corresponds to hardware tracker axis.
If slime VR axis are the same as hardware axis, then NO conversion should be performed at all.

9il:
The PR made acceleration conversion based on existing quaternion conversions.
If you treat SlimeVR as hardware axis on the picture, that isn't correct.
We should easier to get rid of conversion and use hardware axis in the Server, or merge this PR that fixes acceleration axis.

sctanf:
> @sctanf Did you post a hardware tracker axis or a slime VR axis? Because this picture corresponds to hardware tracker axis. If slime VR axis are the same as hardware axis, then NO conversion should be performed at all.

This should be what a device reports, with axes local to the device

9il:
> This should be what a device reports, with axes local to the device

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -436,4 +437,13 @@ class Tracker @JvmOverloads constructor(
 	fun resetFilteringQuats(reference: Quaternion) {
 		filteringHandler.resetMovingAverage(getAdjustedRotation(), reference)
 	}
+
+	companion object {
+		/**
+		 * Changes from IMU axis to OpenGL/SteamVR axis
+		 */
+		fun axisOffset(v: Vector3): Vector3 = Vector3(v.x, v.z, -v.y)
+		private val rotationOffset = Quaternion(INV_SQRT_TWO, -INV_SQRT_TWO, 0f, 0f)
```
sctanf, line 446: The axes may need to be rotated along the device before remap
```suggestion
		fun axisRemap(v: Vector3): Vector3 = Vector3(v.x, v.z, -v.y)
		private val rotationOffset = Quaternion(INV_SQRT_TWO, -INV_SQRT_TWO, 0f, 0f)
		fun axisOffset(v: Vector3): Vector3 = axisRemap(rotationOffset.sandwich(v))
```

9il: `axisOffset` is an optimized version for `rotationOffset.sandwich(v)`
It either one or another, they are the same.

BTW, I would recommend to 
- store only normalized rotations
- use a single rotation function, which needs ~15 multiplications, or at least use two quaternion multiplications. Current `sandwich` code uses quaternion inversion, which us slow. With current state of Koltin you will not feel the difference. However, in the future, with Project Valhalla, Quaternion can be optimized a lot, so the division will be bottleneck. I am thinking about SlimeVR server running on lightweight XR headsets directly, not PC.

Eirenliel:
The accel axises are definitely wrong in this PR, front and top are swapped when testing it with current slimevr fw for both BNO and ICM.

9il:
> The accel axises are definitely wrong in this PR, front and top are swapped when testing it with current slimevr fw for both BNO and ICM.

i would expect they are swapped also somewhere else, if the server shows the expected axis structure before this PR

sctanf:
rotation is corrected for the change in system, but the axes between the two systems are already aligned, the accelerometer data most likely do not need to be remapped

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -436,4 +437,13 @@ class Tracker @JvmOverloads constructor(
...
+		fun axisOffset(q: Quaternion): Quaternion = rotationOffset * q
```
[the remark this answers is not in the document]
9il: hmm, i though i did an equivalent transformation, thank you for the fix

Eirenliel (approved):

Eirenliel:
Seems like stuff is still broken... Debugging it

Eirenliel:
table1

i.. se...

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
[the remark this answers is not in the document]
Eirenliel: You're correct, your transformation is equivalent. Now i'm trying to understand why either of them isn't enough...

Eirenliel:
@9il @sctanf i started working on it in a separate branch, testing with a gravity vector. FOR SOME REASON, it works when used like this, but I don't understand why we need to apply the axis offset twice...

https://github.com/SlimeVR/SlimeVR-Server/tree/accel-mapping-fixes

9il:
@Eirenliel I would like to have but it is hard to understand what is going on.

Does firmware uses global coordinate or local coordinate for acceleration?
In global coordinate Z axis always follows the gravity acceleration (looks to the sky).

What transformation represent the rotation quaternion?
How you transform from global to local or vise-verse (it is the same up to conjugation)?

Does Slime expect acceleration vector with substracted gravity or with gravity present?

Suggestion to get rid of chaos:
1. Use hardware axis structure in the Server
2. Use global coordinate for acceleration (Z -> Sky) (it isn't clear what it is now)
3. Get rid of incoming transformations
4. Adjust Server algorithms such us leg tweaks to work with this input data

Les transformations along the way.

9il:
P.S. thank you for digging in to this rabbit hole

Eirenliel:
@9il do you do any rotations or transformations in firmware?

9il:
> @9il do you do any rotations or transformations in firmware?

@Eirenliel Our firmware uses the same axis as defined in ICM specification (tracker axis). In this case I would say "No". On the other hand, sensor fusion algorithms like VQF has a few axis they operate during fusion, in this terms, "Yes".

Eirenliel:
Thanks. And you confirmed that it works with just your changes? I think I know where the issue is in our code...

9il:
@Eirenliel

I can't confirm our firmware works with SlimeVR, but, it works in our test software. We will test slimeVR later this summer.

One thing I would add is that "front", "right", and "up" aren't very descriptive because they are relative to what one had in their had when they think about a specific tracker.

Please take a look at
image
credit: https://techarthub.com/a-practical-guide-to-unreal-engines-coordinate-system/

Transformation between Left and Right isn't possible with Quaternion. I think this may be a part of the issue. Right-left conversion should be defined with axis permutation and rotation in the same time.

What we need to define for firmware and software API is:

1. Z up or Y up?. The gravity is positive in default tracker position. My personal preference is Z up like in ICM / tracker / blender.
2. Right hand or Left hand. my personal preference is Right up like in ICM / tracker / blender).
3. Is gravity `g` value subtracted from the up axis? e.g. should tracker transfer adjusted acceleration or pure acceleration (that includes g acceleration up).
4. Is rotation quaternion is transformation from local to global axis or vise-versa. They are identical up to conjugation.
5. Should the acceleration be in local or global coordinates. As far as I know fusion frameworks, it should be in global coordinates (gravity vector always up).

image

PR #1016, opened 2024-04-29
GUI - Fix condition for promise detection in a11y

ZycaR:
Small fix to properly detect if condition's result is promise<boolean> or just boolean.
This will fix tons of errors from gui (and dev. console).

image


ImUrX (approved):

ImUrX:
oopsie

Louka3000 (approved):

PR #1029, opened 2024-05-12
Add a way to save serial logs as a file

ImUrX:
This adds a button that let's you save to a file the serial logs, it also does `GET INFO` if it wasn't triggered already.

### DIFF
--- a/gui/public/i18n/en/translation.ftl
+++ b/gui/public/i18n/en/translation.ftl
@@ -469,6 +469,8 @@ settings-serial-get_infos = Get Infos
 settings-serial-serial_select = Select a serial port
 settings-serial-auto_dropdown_item = Auto
 settings-serial-get_wifi_scan = Get WiFi Scan
+settings-serial-file_type = Serial log
+settings-serial-save_logs = Save Logs
 
 ## OSC router settings
 settings-osc-router = OSC router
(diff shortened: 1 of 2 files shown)


gui/public/i18n/en/translation.ftl
```diff
@@ -469,6 +469,8 @@ settings-serial-get_infos = Get Infos
 settings-serial-serial_select = Select a serial port
 settings-serial-auto_dropdown_item = Auto
 settings-serial-get_wifi_scan = Get WiFi Scan
+settings-serial-file_type = Serial log
```
Erimelowo, line 472: Is this really correct? Why not "Text file" or "Plain text"?

Erimelowo (changes requested):
Works fine

gui/public/i18n/en/translation.ftl
ImUrX: i mean sure? idk lol

Erimelowo (approved):

gui/public/i18n/en/translation.ftl
```diff
@@ -477,6 +477,8 @@ settings-serial-get_infos = Get Infos
 settings-serial-serial_select = Select a serial port
 settings-serial-auto_dropdown_item = Auto
 settings-serial-get_wifi_scan = Get WiFi Scan
+settings-serial-file_type = Plain text
+settings-serial-save_logs = Save Logs
```
ButterscotchV, line 481: Maybe this could be a bit clearer about what the button will do?
```suggestion
settings-serial-save_logs = Save To File
```

ButterscotchV (approved):
Looks alright as it is, but maybe the wording can be a little better

PR #1030, opened 2024-05-12
Improvement on firewall batch scripts

### LINKED ISSUE #1026: SlimeVR installer firewall rules
Got connection issues when using slimes for the first time.

There is a firewall script that pus in rules for all the ports, but even with that enabled windows still asks for firewall permission on the OpenJDK binary.

IMG_20240508_153913

When not allowing this the slimes will not connect properly.

The firewall script probably needs to be updated for a smoother user experience.

VocalFan:
Let's go in a list here...

1. Fixes #1026 
2. Makes it where the scripts request admin privileges as they can't modify the firewall without them.
3. Makes sure all the firewall rules are properly enabled with `enable=yes`
4. Fixed the goofball typo of `Discovery defauly port` to `Discovery default port`
5. Prevents duplicate rules by checking if the rule already exists
6. Echoes/Tells the user what the name of the rule being added is.

Before:

image

After:

image

(UAC prompt quickly appears. It's so fast I had to use `timeout 5` to have a delay so I could screenshot)

(User presses No, the script closes.)

(User presses Yes, script continues as below)

image


VocalFan:
@electromuis

For you, my king-

electromuis:
Just did a quick test and indeed seems to fix the issue. Although the %CD% value does have to be correct (Right click as admin will run it from System32 and thus not specify the correct java binary). But this would have to be setup correctly in the installer I'd assume.

VocalFan:
> Just did a quick test and indeed seems to fix the issue. Although the %CD% value does have to be correct (Right click as admin will run it from System32 and thus not specify the correct java binary). But this would have to be setup correctly in the installer I'd assume.

Sadly just a skill issue on Window's part.

VocalFan:
@electromuis Gonna try changing to Powershell when I get home.

VocalFan:
@electromuis Try again! I think I fixed the issue with right clicking and using Run As Administrator... Also first, use uninstall_firewall a few times as I also added duplicate detection to the install script... So you might currently have duplicate rules.

unlogisch04:
Just short, I think you need to allow javaw.exe and not java.exe do you have a way to confirm?

VocalFan:
> Just short, I think you need to allow javaw.exe and not java.exe do you have a way to confirm?

...Look above, @electromuis had his issue fixed with java.exe being allowed (also some internet searches agree with this)

ButterscotchV:
> Just short, I think you need to allow javaw.exe and not java.exe do you have a way to confirm?

You could do both, I believe we're only using java.exe though

ImUrX:
do we need to have the UAC prompt script?

VocalFan:
> do we need to have the UAC prompt script?

As discussed on the top post, firewall modification fails without admin perms. And besides, if the script is already given admin, the prompt is skipped

ButterscotchV:
> do we need to have the UAC prompt script?

iirc it's the best way to do it, batch scripts stink. I have worries about it having different behaviour in different versions of Windows but it's probably alright?

VocalFan:
It's been basically a tried and true method for most versions of windows... Atleast any that support UAC :P

ButterscotchV:
> It's been basically a tried and true method for most versions of windows... Atleast any that support UAC :P

Well if it works then sweet at least Windows can do something good lmao, we already had a blunder with the installer crashing because we had the audacity to format JSON on Windows 7.

This PR looks good to be, but not reviewing because I can't really thoroughly check it rn. May do so later.

server/core/resources/firewall.bat
```diff
@@ -1,17 +1,49 @@
...
+)
+
 echo Installing firewall rules...
 
 rem Discovery default port
-netsh advfirewall firewall add rule name="SlimeVR UDP 35903 incoming" dir=in action=allow protocol=UDP localport=35903
-netsh advfirewall firewall add rule name="SlimeVR UDP 35903 outgoing" dir=out action=allow protocol=UDP localport=35903
-
+call :AddRule "SlimeVR UDP 35903 incoming" "dir=in action=allow protocol=UDP localport=35903 enable=yes"
+call :AddRule "SlimeVR UDP 35903 outgoing" "dir=out action=allow protocol=UDP localport=35903 enable=yes"
```
unlogisch04, line 19:
```suggestion
```
I only see this port number used here in the code:
https://github.com/SlimeVR/SlimeVR-Server/blob/e2a511b552d52ebc77da0cb029440600edd5ddfb/server/desktop/src/main/java/dev/slimevr/desktop/Main.kt#L99

In a close statement, so don't think it is used anywhere

unlogisch04 (commented):
Well i tested some Stuff on Windows 11 Version 22631.3593
I did not test the script itself. To this date.

For me Windows 11 only adds rules on Incomming to the firewall when i deny or allow the Program via popup. So not sure if this is different in the Version @electromuis uses.

server/core/resources/firewall.bat
VocalFan: To be fair, I kept all the rules already existing in the old script.

server/core/resources/firewall.bat
```diff
@@ -1,17 +1,49 @@
...
-
+call :AddRule "SlimeVR UDP 6969 incoming" "dir=in action=allow protocol=UDP localport=6969 enable=yes"
+call :AddRule "SlimeVR UDP 6969 outgoing" "dir=out action=allow protocol=UDP localport=6969 enable=yes"
 rem WebSocket server default port
-netsh advfirewall firewall add rule name="SlimeVR TCP 21110 incoming" dir=in action=allow protocol=TCP localport=21110
-netsh advfirewall firewall add rule name="SlimeVR TCP 21110 outgoing" dir=out action=allow protocol=TCP localport=21110
+call :AddRule "SlimeVR TCP 21110 incoming" "dir=in action=allow protocol=TCP localport=21110 enable=yes"
+call :AddRule "SlimeVR TCP 21110 outgoing" "dir=out action=allow protocol=TCP localport=21110 enable=yes"
+rem OpenJDK Platform Binary access
+call :AddRule "SlimeVR OpenJDK Platform incoming" "dir=in action=allow program=%~dp0jre\bin\java.exe enable=yes"
```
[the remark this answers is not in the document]
VocalFan: Stated in other message, I just kept all the rules in the old script

VocalFan:
So uh, idk if I should do those changes or keep the rules for 'better safe than sorry'

server/core/resources/firewall.bat
ImUrX: this can be removed, @Eirenliel do you know where this comes from?

ButterscotchV: I say we get rid of those rules, there's no need anymore

ButterscotchV (approved):

loucass003 (approved):

server/core/resources/firewall.bat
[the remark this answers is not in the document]
VocalFan: Keeping In, as Out is needed, potentially In is needed too.

server/core/resources/firewall.bat
```diff
@@ -1,17 +1,47 @@
...
+    :: Temp script to request admin... Works and doesn't leave a mess.
+    echo Set UAC = CreateObject^("Shell.Application"^) > "%temp%\getadmin.vbs"
+    echo UAC.ShellExecute "%~s0", "", "%~dp0", "runas", 1 >> "%temp%\getadmin.vbs"
+    "%temp%\getadmin.vbs"
+    exit /B
+)
+
 echo Installing firewall rules...
 
 rem Discovery default port
```
ButterscotchV, line 17:
```suggestion
```

ImUrX (approved):

PR #1033, opened 2024-05-14
Added wifi dropdown to the wizzard

electromuis:
This feature changes the wifi form in the setup wizzard.

It will find networks from:
 - Connected slime tracker wifi scanning
 - Windows saved wifi networks
 - Windows wifi scanning

Stubs have been put in place for linux and mac. These could possibly filled in with: https://github.com/booyaa/wifiscanner/tree/main

Although scanning via a slime tracker should always work.

This is my first contribution, please let me know if changes are required.


Eirenliel (approved):

Erimelowo (changes requested):
ZRock and I couldn't get it to work (it just falls back to the text field input, which I assume means no wifi network was found somehow). On Windows 11

electromuis:
> ZRock and I couldn't get it to work (it just falls back to the text field input, which I assume means no wifi network was found somehow). On Windows 11

Your computer did have wifi networks available? I'll check it on another computer.

Also would a notice if it didn't find anything be a good addition or just bloat for the end user?

Erimelowo:
I didn’t have wifi turned on, I’ll test again with it turned on tomorrow (Idk about ZRock). Didn’t think about the fact that Ethernet wouldn’t give any SSID, my bad ahah. I thought I did test with a SlimeVR tracker turned on and connected via Wifi tho, how does finding networks via “Connected slime tracker wifi scanning” work/how can I test that?

As for a notice, I think it might just confuse users into thinking something is wrong when it’s not.

Erimelowo:
https://cdn.discordapp.com/attachments/823658497454374962/1248420842917072957/slimevr_W4v5cdBMYd.mp4?ex=66639a03&is=66624883&hm=84c14bfe79ebc35e1822806223b76a6660722a59265b97e937b8a03b3a4aeac6&

idk why it autofilled but didn't show a dropdown on the first try, but I recorded that video to show that going from Other to a Wifi network in the dropdown will leave the Submit button disabled.

Not sure how many wifi networks it's supposed to have in the dropdown, but it only showed the one I was connected to (which shouldn't be a problem, but I thought it'd also have the scanned available wifi networks)

Another issue: it only shows the dropdown every other time: https://cdn.discordapp.com/attachments/823658497454374962/1248422134905311353/slimevr_emybDRyOrh.mp4?ex=66639b37&is=666249b7&hm=ee1909308e4b0df2895f96f7fd65a80778c048985b510a11947a05f1b2b6b93a&

electromuis:
Fixed some bug where it wouldn't load or just partially load

ImUrX:
btw i just checked and `netsh` output is language-dependent, its not always english :p

ImUrX:
Alternative for windows would be to do a [WlanScan](https://learn.microsoft.com/en-us/windows/win32/api/wlanapi/nf-wlanapi-wlanscan) and then a [WlanGetAvailableNetworkList](https://learn.microsoft.com/en-us/windows/win32/api/wlanapi/nf-wlanapi-wlangetavailablenetworklist) which sounds very fun

PR #1036, opened 2024-05-19
De-duplicate repeated HID packets

nekomona:
Current implementation of slimenrf-receiver fills empty space with repeated duplication of the last packet to fulfill HID packet length requirement. The duplicated packets would interfere with TPS count and filtering etc.. This PR filters out duplicated packets on the server-side.


sctanf (approved):

ImUrX (changes requested):
is it possible to not use a reference instead of copying part of an array?

ImUrX (approved):

Eirenliel (approved):

Eirenliel:
If you want to access arrays by references, you need to use ByteBuffers.

PR #1045, opened 2024-06-01
Basic motion compensation

Erimelowo:
This currently requires the tracker to be aligned to the Identity rotation IRL, which isn't very user-friendly as users can only know where it is by trial and error. Input welcome on alternate ways to handle this.


Eirenliel:
What is this? For people on a boat?

PR #1046, opened 2024-06-02
Fix mocap mode breaking on startup

Stermere:
extendedPelvisYawRoll in HumanSkeleton could sometimes return an invalid rotation quaternion on startup. With mocap mode disabled this value gets overwritten on the next iteration. However, when mocap mode is enabled the invalid rotation is used on calculations involved in setting the next head position causing NaNs to propagate to the head position.

This PR replaces the invalid rotation with the identity quaternion preventing this from occurring.


Erimelowo (approved):
lgtm, altho this could be done directly in the `extendedPelvisYawRoll` method

ImUrX (approved):

ButterscotchV (approved):

Eirenliel (approved):

Issue #961, opened 2024-03-15
Support for bend/potentiometer sensors


Eirenliel:
I think we will be able to support so many new hardware if the server can support a bend / potentiometer tracker. It's a tracker placed on a joint, instead of a bone, that has only one input - bend angle. For example, this can be used to track bend of the ankle, or in the future - fingers for some applications.

PR #1095, opened 2024-06-29
Flex resistance and angle support

Erimelowo:
*This is a WIP prototype and everything may change since this is a draft!*

Implements flex sensor support for UDP trackers.
This will be able to be used for any joint with a single (more or less...) degree of freedom such as the knee, fingers or even shoulders. Main use is for fingers.
GizmoQC's working repo for flex gloves firmware: https://github.com/Guizmo12/gizmoglovesmocap

Needs https://github.com/SlimeVR/SolarXR-Protocol/pull/149  
Based off https://github.com/SlimeVR/SlimeVR-Server/pull/1091
Closes https://github.com/SlimeVR/SlimeVR-Server/issues/961


server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerDataSupport.kt
```diff
@@ -0,0 +1,17 @@
+package dev.slimevr.tracking.trackers
+
+// Note: there is a -1 offset when receiving via UDP, thus ROTATION must be sent as 0 from firmware.
```
Eirenliel, line 3: Why is there offset? 0 is the default state and rotation is the default state. Make it ROTATION(0). If you need OTHER for some reason, make it 255.

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerDataSupport.kt
```diff
@@ -0,0 +1,17 @@
...
+enum class TrackerDataSupport(val id: Int) {
```
Eirenliel, line 4:
```suggestion
enum class TrackerDataType(val id: Int) {
```

Eirenliel (changes requested):

Eirenliel (changes requested):
Update tracker position protocol

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerDataSupport.kt
Erimelowo: It was because I wanted the default 0 to be Other, I don't remember why

Issue #1079, opened 2024-06-18
Add a background or a border to the 3d preview so it does not merge with the page background


loucass003:
![image](https://github.com/SlimeVR/SlimeVR-Server/assets/3948904/c9a02a4e-0970-4b4c-99b6-fbe0c9693b91)

ZycaR:
@loucass003, which background is better: darker or lighter?
![obrazek](https://github.com/SlimeVR/SlimeVR-Server/assets/11602729/4d61bad6-cad3-4a53-93ff-cf6ac0fcb583)
![obrazek](https://github.com/SlimeVR/SlimeVR-Server/assets/11602729/c7435cab-8b37-4650-b0a5-c287b4b89ea2)

Additionally, we have a skeleton image on Home page (and other places). Should we apply a background for the 3D preview everywhere, or just on the Manual Body Proposition page?
![obrazek](https://github.com/SlimeVR/SlimeVR-Server/assets/11602729/1e218078-8924-48d9-a0a9-3ee0d650748d)

loucass003:
imo lighter fits the UI more

PR #1107, opened 2024-07-11
Issue#1079 - Add a background to the 3d preview skeleton

ZycaR:
Fixes https://github.com/SlimeVR/SlimeVR-Server/issues/1079 - Add a background or a border to the 3d preview so it does not merge with the page background.

**UI Changes:** Added a background to the 3D preview to ensure it stands out from the page background.
The 3D preview now does not blend into the page background and should work for all visual **themes** (see image).

**Implementation Note:** The background for the 3D preview has been applied universally across all pages, as per the feedback received.
skell-anim

### DIFF
--- a/gui/src/components/widgets/SkeletonVisualizerWidget.tsx
+++ b/gui/src/components/widgets/SkeletonVisualizerWidget.tsx
@@ -110,9 +110,9 @@ export function ToggleableSkeletonVisualizerWidget(
         </Button>
       )}
       {enabled && (
-        <>
+        <div className="flex flex-col gap-2">
           <Button
-            className="w-full"
+            className="w-full "
             variant="secondary"
             onClick={() => {
               setEnabled(false);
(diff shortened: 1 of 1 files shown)


loucass003 (approved):

gui/src/components/widgets/SkeletonVisualizerWidget.tsx
```diff
@@ -110,7 +110,7 @@ export function ToggleableSkeletonVisualizerWidget(
         </Button>
       )}
       {enabled && (
-        <>
+        <div className="flex flex-col gap-2">
```
ImUrX, line 113: why?

ImUrX (approved):

gui/src/components/widgets/SkeletonVisualizerWidget.tsx
ZycaR: small gap between button and background of 3D preview. Without gap it renders together and it does not fit actual layout:
image
with gap:
image

PR #1116, opened 2024-07-21
Use default security attributes for named pipes

ButterscotchV:
There's a *lot* of weird confusing stuff behind this, but basically this allows the named pipes to connect if SlimeVR is run as an admin. I don't know any use-case where people would *want* to run SlimeVR as an admin, but for some reason a good number of people do. This would just make it work regardless. Alternatively we could just warn the user if it's run as an admin and tell them not to.

I have *no* idea how this security attribute stuff works, but I know at a minimum that this configuration *works*. Please suggest anything that can be changed about it as I'm certain this can be done better.

Relevant resources:
- https://stackoverflow.com/questions/57748611/how-to-make-named-pipe-accessible-for-non-elevated-processes
- https://learn.microsoft.com/en-us/windows/win32/api/wtypesbase/ns-wtypesbase-security_attributes
- https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-security_descriptor
- https://github.com/java-native-access/jna/blob/master/contrib/platform/test/com/sun/jna/platform/win32/Advapi32Test.java


ImUrX (commented):
we should just error when its being ran as admin tbh

ImUrX (approved):

Erimelowo (approved):
Tested and it works fine on Windows 11

PR #1121, opened 2024-07-24
Update flake nixpkgs revision

gale-username:
Newer nixpkgs revisions need GST_PLUGIN_SYSTEM_PATH_1_0 set. Keeping stuff working with the latest nixpkgs helps with packaging and getting mesa versions to match.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -313,11 +313,11 @@
     },
     "nixpkgs_2": {
       "locked": {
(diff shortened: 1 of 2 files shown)


ImUrX:
oh so this was the reason it didnt find with newer versions? where did u find this btw?

ImUrX (approved):

PR #1080, opened 2024-06-18
Remove useLayout

loucass003:
This will remove the use of useLayout that was handling the size of most layouts with javascript. This was slow and also didnt work all the time when resizing the window. I removed that function and redid all the layouts with css only.
I also slided in minor css tweaks for padding and alignments, mostly for mobile

Fixes #1067

PR #1129, opened 2024-08-02
Fix many GUI bugs found

### LINKED ISSUE #1109: Tracker tap assignment stops working after first use
The 2-tap for tracker assignment in the Tracker Assignment page stops working after only using it just once, it starts working again if you leave and enter the page which seems to be an state bug somewhere on the SolarXR packet or in the server-side

### LINKED ISSUE #1120: Hand warning modal triggers if hand trackers were already on
The hand warning modal triggers everytime even if the hand trackers were already on. It should only trigger when toggling on the hand trackers, not when it is toggled on.

https://cdn.discordapp.com/attachments/932234552431247431/1265186615278571531/slimevr_3hzHFWMyTo.mp4?ex=66a1411b&is=669fef9b&hm=8165600d1e7b6663e24ca04e82e9ff6e97d0b069a3d303e902ba31dd92dd40ee&

ImUrX:
- truncate too big tracker names
- fix elements using hidden not working (all #1080's fault)
- improve rest calibration tutorial by checking accel data of all trackers
- removed automatic mounting modal warning and made it recommended again
- fix rust warning about unused variable
- fix tracker assign using sm instead of xs media queries
- fixes #1109 
- fixes #1120


gui/src/maths/vector3.ts
```diff
@@ -3,6 +3,12 @@ import { Vector3 } from 'three';
 
 export type Vector3Object = { x: number; y: number; z: number };
 
+export function averageVector(vecs: Vector3[]) {
+  if (vecs.length === 0) return new Vector3();
+  const sum = vecs.reduce((prev, curr) => prev.add(curr), new Vector3());
+  return sum.divideScalar(vecs.length);
```
ButterscotchV, line 9: This seems like a solid chance of having a length of 0, which would not be good to divide by

ButterscotchV (approved):

PR #1132, opened 2024-08-07
Fix skeleton issues

ButterscotchV:
Fixes yaw reset condition, head length condition, and replaces String.format with inline string format from Kotlin.
Needs to be tested before being merged.

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
@@ -889,7 +889,7 @@ class HumanSkeleton(
 		var transOffset = offset
 
 		// If no head position, headShift and neckLength = 0
-		if (boneType == BoneType.HEAD || boneType == BoneType.NECK && (headTracker == null || !(headTracker!!.hasPosition && headTracker!!.hasRotation))) {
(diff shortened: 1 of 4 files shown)


Erimelowo (approved):
lgtm

ButterscotchV:
Tested myself and found the neck still wasn't being updated, so I made it update all the nodes since this code isn't run often and doing so will just prevent any future bugs like that

Other than that, it seems to work for me, though I haven't tested thoroughly

ImUrX (approved):

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -298,11 +298,8 @@ class HumanSkeleton(
 		// Rebuilds the arm skeleton nodes attachments
 		assembleSkeletonArms(true)
 
-		// Refresh headShift
-		humanPoseManager.computeNodeOffset(BoneType.HEAD)
-
-		// Refresh node offsets for arms
-		computeDependentArmOffsets()
```
Erimelowo, line 305: is that method no longer being used? It should be removed if we don't want to use it anymore then

ButterscotchV: It's still used when the arms are switched, I made sure, I figured I'd just leave it because there's no reason to switch that

PR #1133, opened 2024-08-08
Fix Android build & OSC support

ButterscotchV:
JavaOSC 0.9 breaks Android support in this commit https://github.com/hoijui/JavaOSC/commit/2cbbc433a354fc0ba326b44826e1ae2b0a7f8350 by using `ClassLoader.getDefinedPackage()`, so we need to revert back to version 0.8.

Also it seems some dependency is using a different version of log4j so we need to add their classes to proguard rules to comply for minify on Android.


ImUrX (approved):
sad moment

PR #1057, opened 2024-06-13
Computed head reset

Erimelowo:
Status: waiting on preliminary feedback before doing an open beta

This PR adds a toggle to reset the HMD's pitch upon full reset.
It also allows other computed (non-hmd) trackers, such as vive trackers, to be Full Reset and Mounting Reset according to the identity quaternion when wearing them on the head (unrelated to the aforementioned toggle). TODO (?): use mounting reset to adjust instead of the identity quaternion.

Needs: https://github.com/SlimeVR/SolarXR-Protocol/pull/142, https://github.com/SlimeVR/SlimeVR-Server/pull/962
Closes: https://github.com/SlimeVR/SlimeVR-Server/issues/628

Tracker unit tests currently fail idk why

PR #1150, opened 2024-08-28
Fix tracker reset logic

ButterscotchV:
The tracker reset logic was changed in #1057, this broke AutoBone by causing the recorded trackers to be reset. This PR should fix the logic, though I'm not exactly sure what was intended by the logic. I've also added comments because it's not super readable.

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
@@ -323,7 +323,8 @@ class Tracker @JvmOverloads constructor(
 			_rotation
 		}
 
-		if (needsReset || (isComputed && !isInternal)) {
+		// Reset if needed and is not computed and internal
+		if (needsReset && !(isComputed && isInternal)) {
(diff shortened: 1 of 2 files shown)


ImUrX (approved):

Erimelowo (changes requested):
This’ll change behaviour. You should just set the Autobone trackers as internal if you don’t want them reset as that’s what they are

ButterscotchV:
> This’ll change behaviour. You should just set the Autobone trackers as internal if you don’t want them reset as that’s what they are

If I set it to internal, then HumanSkeleton won't accept it. It's not internal anyways, that's not what that flag was supposed to mean. And why should the default behaviour be ignoring the `needsReset` flag?

Erimelowo:
IIRC it’s because the HMD has the needReset flag to false. 
Would need to set that flag to true in the SteamVRDriver class and then there might be a condition to change in the resetHandler or HumanSkeleton.

Erimelowo (approved):
Good for old behaviour and Autobone seems to work (both tested)

PR #1151, opened 2024-09-02
Fix audio/video issues not playing on Linux

ImUrX:
Audio and video doesnt work on Linux because of dumb bug https://github.com/tauri-apps/tauri/issues/3725, and i fixed it in a way lol


ImUrX:
ignore stupid mistake of leaking auth key, i revoked it lol

kruemmelbande:
Can confirm, fixes things :3

ButterscotchV (approved):

PR #1152, opened 2024-09-03
Hide navbar onboarding

### LINKED ISSUE #894: GUI - first page of Setup Wizard shouldn't block Left side menu
Setup Wizard occupies the entire window and if you click on it by accident (and accidental clicks are not uncommon when you using controllers) forces you to go to the top right corner and aim for the close button which isn't that big either. So at least the first page should have acess to the left side menu allowing to quickly leave the wizard.

It would also be nice to be able to completely hide the wizard button somewhere in Settings - Interface - Appearance.

wizard

JovannMC:
Partially fixes #894 by allowing users to hide the onboarding button to prevent accidental clicks

image


gui/public/i18n/en/translation.ftl
```diff
@@ -424,6 +424,9 @@ settings-general-interface-dev_mode = Developer Mode
 settings-general-interface-dev_mode-description = This mode can be useful if you need in-depth data or to interact with connected trackers on a more advanced level.
 settings-general-interface-dev_mode-label = Developer Mode
 settings-general-interface-theme = Color theme
+settings-general-interface-show-navbar-onboarding = Show setup wizard on navigation bar
+settings-general-interface-show-navbar-onboarding-description = This changes if the "Setup Wizard" button shows on the navigation bar.
```
ImUrX, line 428: instead of "Setup Wizard" maybe you want to reference the translation key itself?
you can do that by doing `{ navbar-onboarding }`

ImUrX (changes requested):

gui/public/i18n/en/translation.ftl
```diff
@@ -424,6 +424,9 @@ settings-general-interface-dev_mode = Developer Mode
 settings-general-interface-dev_mode-description = This mode can be useful if you need in-depth data or to interact with connected trackers on a more advanced level.
 settings-general-interface-dev_mode-label = Developer Mode
 settings-general-interface-theme = Color theme
+settings-general-interface-show-navbar-onboarding = Show "{ navbar-onboarding }" on navigation bar
+settings-general-interface-show-navbar-onboarding-description = This changes if the "{ navbar-onboarding }" button shows on the navigation bar.
+settings-general-interface-show-navbar-onboarding-label = Show { navbar-onboarding }
```
ImUrX, line 429:
```suggestion
settings-general-interface-show-navbar-onboarding-label = Show "{ navbar-onboarding }"
```

ImUrX (approved):

gui/public/i18n/en/translation.ftl
ImUrX: a little more correct

ButterscotchV:
Maybe there could also be a button in the settings to open the setup wizard so even with it hidden you could go back?

JovannMC:
> Maybe there could also be a button in the settings to open the setup wizard so even with it hidden you could go back?

I don't really see the point of having the button just in case the user wants to use it. At that point they can just temporarily show the setup wizard again because they are already in settings.

ButterscotchV (approved):

PR #1160, opened 2024-09-16
Add drift compensation warning

JovannMC:
Adds a warning when enabling drift compensation - it should only be used on "bad" IMUs and cause more drift on better IMUs (like BNO)

image

### DIFF
--- a/gui/src/components/settings/pages/GeneralSettings.tsx
+++ b/gui/src/components/settings/pages/GeneralSettings.tsx
@@ -436,7 +436,7 @@ export function GeneralSettings() {
   //   }
   // }, [state]);
 
-  const [skipDriftCompWarning, setSkipDriftCompWarning] = useState(false);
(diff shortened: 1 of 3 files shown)


ImUrX (approved):

gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -435,6 +436,8 @@ export function GeneralSettings() {
   //   }
   // }, [state]);
 
+  const [skipDriftCompWarning, setSkipDriftCompWarning] = useState(false);
```
ButterscotchV, line 439: The name of this doesn't really fit the behaviour, this could follow the implementation for the hand warning
> If null, we still haven't shown the hands warning if false then initially the hands warning was disabled

Or just rename to `showDriftCompWarning`, I think it'd be worthwhile to just show it all the time anyways

ButterscotchV (changes requested):

ButterscotchV (approved):

PR #1168, opened 2024-09-18
check flatpak runtime dir first

0chroma:
This change would allow SlimeVR to set up sockets on the host while running in a flatpak. In order to do this, we set the socket dir to `$XDG_RUNTIME_DIR/app/$FLATPAK_ID` if the `FLATPAK_ID` environment variable is set.

Flatpak's sandbox doesn't allow writes to the runtime root, for more details see: https://docs.flatpak.org/en/latest/sandbox-permissions.html

Related changes: https://github.com/SlimeVR/SlimeVR-OpenVR-Driver/pull/48 and https://github.com/SlimeVR/SlimeVR-Feeder-App/pull/22


ButterscotchV (approved):

Eirenliel (approved):

ImUrX:
currently discussing where stuff should be on https://github.com/SlimeVR/SlimeVR-OpenVR-Driver/pull/48

PR #1169, opened 2024-09-19
Restrict mounting reset on server by tracker reset status

ButterscotchV:
You can still do mounting reset even when the button on the GUI is disabled by using tap reset or bindings, this PR just restricts the reset on the backend entirely, this isn't a perfect implementation but I've tested and it works good enough.


ImUrX (approved):
very cool, there is no feedback on GUI but its not that bad tbh

loucass003 (approved):

PR #1170, opened 2024-09-19
Tracker status box shadow color fix

Polymoria:
I fixed a color on status box, now its grabs from theme instead of being static. Also i proposed to change a color of shaking slimes, i think its more cool, but still in line with guidlines. 

image

### DIFF
--- a/gui/src/components/tracker/TrackerCard.tsx
+++ b/gui/src/components/tracker/TrackerCard.tsx
@@ -78,7 +78,7 @@ function TrackerSmol({
       <div className="flex flex-col justify-center items-center fill-background-10">
         <BodyPartIcon bodyPart={tracker.info?.bodyPart}></BodyPartIcon>
       </div>
-      <div className="flex flex-col flex-grow ">
+      <div className="flex flex-col flex-grow justify-center">
         <Typography bold truncate>
           {trackerName}
         </Typography>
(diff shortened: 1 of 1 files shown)


gui/src/components/tracker/TrackerCard.tsx
```diff
@@ -150,7 +150,7 @@ export function TrackerCard({
           ? {
               boxShadow: `0px 0px ${Math.floor(velocity * 8)}px ${Math.floor(
                 velocity * 8
-              )}px #BB8AE5`,
+              )}px var(--tw-shadow-color)`,
```
ImUrX, line 153:
```suggestion
              )}px var(--accent-background-30)`,
```

gui/src/components/tracker/TrackerCard.tsx
```diff
@@ -139,7 +139,7 @@ export function TrackerCard({
     <div
       onClick={onClick}
       className={classNames(
-        'rounded-lg overflow-hidden',
+        'rounded-lg overflow-hidden shadow-accent-background-30',
```
ImUrX, line 142:
```suggestion
        'rounded-lg overflow-hidden',
```

ImUrX (changes requested):
I don't like the new color that much, I think you can just use the `accent-background` variable directly on the style :3

Polymoria:
> I don't like the new color that much

This color is still recogniseble, but it goes more inline with a dark theme and other color on the page and not creating color dominant with a little oversaturated vibe, but still work for color-blind ppl as i check by making a screenshot for this site: 

image

ImUrX (approved):

ButterscotchV (approved):
Approved without testing, please validate in release candidate

PR #1173, opened 2024-09-21
Update all gradle dependencies we have

ImUrX:
- Updated Kotlin to v2, which includes compiler improvements and fixes value classes bugs.
- Updated Android stuff as we can't upload to Google Play otherwise
- Updated Gradle to latest as Android needed it

Requires Android testing to see if update didn't have any issues


ButterscotchV (approved):

loucass003 (approved):
Tested with android, looks good

PR #1175, opened 2024-09-21
fix(gui): tab title

TheDevMinerTV:


ImUrX (approved):

ButterscotchV (approved):
Could also just be "SlimeVR", fine either way since most don't see this

PR #1184, opened 2024-09-28
Add extrapolated prediction to drift compensation

nekomona:
Previous drift compensation only interpolates between current yaw and the inverse of recorded drifts. This limits the compensation effectiveness as the drift would spin the tracker continuously, moving beyond the recorded drift angle.

This PR would keep rotating the compensation angle at the speed averaged from recorded drifts. It would give better compensation result at the presence of continuous drifting for a longer period of time.

Depending on https://github.com/SlimeVR/SolarXR-Protocol/pull/148


ButterscotchV (approved):
Looks good, previous implementation was kinda weird so this is much cleaner

PR #1186, opened 2024-09-29
add a way to skip rest calibration tutorial

ImUrX:
Adds a 10s timer that makes appear the skip button, and also doesnt show calibration tutorial for mocopi or haritora


loucass003 (changes requested):
Aside for the one comment, this looks good

gui/src/hooks/imu-logic.ts
```diff
@@ -1,11 +1,21 @@
...
 ): boolean {
   const imuExists = useMemo(
-    () => connectedTrackers.some((tracker) => tracker.tracker.info?.isImu),
+    () =>
+      connectedTrackers.some(
+        (tracker) =>
+          tracker.tracker.info?.isImu &&
+          !(
+            tracker.device?.hardwareInfo?.boardType &&
+            IGNORED_BOARDS.has(tracker.device?.hardwareInfo?.boardType as string)
```
ButterscotchV, line 16: I would suggest using `McuType` but it doesn't look like we have one for Haritora, so whatever

ButterscotchV (commented):
Looks good besides the comment from earlier

gui/src/hooks/imu-logic.ts
ImUrX: but its wrong to call haritora an mcutype, its a DA14

ButterscotchV (approved):

loucass003 (approved):

PR #1191, opened 2024-10-07
Add acceleration & position to tracker page

### LINKED ISSUE #1187: Show accel and position outside of dev mode
This should be super easy and all the data is already available, we just need some way of actually seeing a tracker's acceleration and position outside of dev mode. I'd suggest alonside the rotation on the tracker page.

JovannMC:
Adds acceleration & position to the tracker page, renaming "Rotation" to "Tracking data". Fixes #1187.

Acceleration/position will be hidden if not available (position hidden in screenshot).

image


ButterscotchV (approved):
🐱

loucass003 (approved):

PR #1192, opened 2024-10-08
enforce lf with editorconfig

ImUrX:


ImUrX:
we can add git also, i just lazy for adding a `.gitattributes`

PR #1196, opened 2024-10-12
Add settings profiles

### LINKED ISSUE #38: Body proportions profiles
Need to add a system to save and load configuration profiles, so different people could use the same PC to play, and to do easy experiments.

JovannMC:
Draft PR to add settings profiles so users can create different profiles for different settings, including body proportions.

Current plan is to have a `profiles` folder in SlimeVR's data folder which will have folders with the names of the profiles, each of them containing their own `vrsettings.yml` and `gui-settings.dat` (e.g. `dev.slimevr.SlimeVR/profiles/Bingus`).

Fixes #38.

### DIFF
--- a/gui/public/i18n/en/translation.ftl
+++ b/gui/public/i18n/en/translation.ftl
@@ -280,6 +280,7 @@ settings-sidebar-utils = Utilities
 settings-sidebar-serial = Serial console
 settings-sidebar-appearance = Appearance
 settings-sidebar-notifications = Notifications
+settings-sidebar-profiles = Profiles
 settings-sidebar-advanced = Advanced
 
 ## SteamVR settings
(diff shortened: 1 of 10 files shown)


Eirenliel:
I feel like body proportions should be a profile, but settings - idk maybe? Kinda weird to group them all together, because one person might need multiple body proportions profiles for like different sets/avatars/people or to experiment, but the same settings.

JovannMC:
Mm that's fair. Since we were going to do body proportions, I thought I might as well implement settings too.
We could make them separate if we still want settings profiles, but if it's not really necessary I can implement just body proportions.

Spazznyan:
I like the idea of it being all the settings. Can have like a streaming profile and vrchat profile without having to go through and toggle all the VMC/OSC stuffs. 

Question is, should a new profile be a copy of the current one or completely fresh with default values?

JovannMC:
I was thinking there could be a prompt to ask what the user wants, the defaults or copy their current profile settings

loucass003:
Lets wake up this PR. 

There are the requirements i think need to be meet for this PR:

- Implement user profiles and environment profiles
    - User profiles describe body proportions, mounting and assignment settings. Settings that should always be different from an user to another
    - environment  profiles describe steamvr / osc / vmc settings…, settings that related to the hardware or software running on the computer and how the trackers are used
    - Implementation details:
        - This is probably gonna need a re organisation of the config files, one per profiles and split between user and environment profiles
        - This will also change the solarxr protocol, the problem is that the settings are quite messy at the moment. idk how clean we can make this atm
        - we will also need a migration procedure for this
    - The user should be able to quickly switch between profiles at any moments and the ui should update accordingly (more tricky than it looks as almost everything use settings) / We could also reset the ui when switching profiles as a first implementation (safest)
    - The user should be able to create as many profiles as they like and manage them (copy / delete / rename)
    - There should always be at least one profile of each type (user, environment) available
    - any changes in settings always modify the current selected profiles
    
    i will ofc help with the implemetation
    
    @Eirenliel @JovannMC Please tell me what you think.

PR #1200, opened 2024-10-14
Hid update protocol

sctanf:
switch to 16 byte packets
fill in more data for hid devices and hid trackers
change unique identifier for hid trackers
add support for exponential map quantized quaternions
http://marc-b-reynolds.github.io/quaternions/2017/05/02/QuatQuantPart1.html#fnref:pos:3
https://github.com/Marc-B-Reynolds/Stand-alone-junk/blob/559bd78893a3a95cdee1845834c632141b945a45/src/Posts/quatquant0.c#L898


server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -56,7 +56,7 @@ class Tracker @JvmOverloads constructor(
 	 * Has solved position + rotation (Vive trackers)
 	 */
 	val isComputed: Boolean = false,
-	val imuType: IMUType? = null,
+	var imuType: IMUType? = null, // TrackersHID has to assign after the device is created
```
ImUrX, line 59: but a tracker should be initialized when the tracker info is received, not when the device is created...

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -199,64 +212,197 @@ class TrackersHID(name: String, private val trackersConsumer: Consumer<Tracker>)
 				if (dataReceived.isNotEmpty()) {
 					// Process data
 					// TODO: make this less bad
-					if (dataReceived.size % 20 != 0) {
+					if (dataReceived.size % 16 != 0) { // TODO: packet size 20 is valid (old)
```
ImUrX, line 215: use a `const` for this `PACKET_SIZE` please

ImUrX (changes requested):

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
sctanf: i can wait for the info packet before creating the tracker, but that depends if the tracker has time to send the info packet, it must be awake but not moving

ImUrX: this sounds like something that should be fixed on esb protocol.
idk, i would like more opinions if we should let trackers be lazily updated or for a tracker to be initialized it should contain the info for its initialization

sctanf: i can update it to wait for the info packet, but it means the tracker may not get picked up quickly
a user probably does not constantly move around right after a device is started, so it is ok
devices were allowed to override its data, like MCUType, only for trackers it cannot be overridden without modifying base tracker

sctanf:
some latency issue was appearing on receiver reconnect, maybe it is some buffer, so switched to using readAll

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
sctanf: after some more use, it doesnt seem like it is much of an issue as i thought

Eirenliel (approved):

PR #1204, opened 2024-10-18
Fix style for tracker "shadow" & add to all cases

ButterscotchV:
#1170 broke the tracker shadow when not in dev mode, the style should have `rgb` around `var`. This PR also applies the same change to all cases of the tracker shadow.


ImUrX (approved):

ButterscotchV:
Doesn't look very good on the trans theme, every other screen looks alright, but a little low contrast
image

ImUrX (approved):

Erimelowo (approved):
lgtm

unrelated, but the "shadow" fades out much faster without dev mode than with dev mode.

PR #1206, opened 2024-10-21
Fix typo on `OBTAINING_MAC_ADDRESS`

ImUrX:
Requires https://github.com/SlimeVR/SolarXR-Protocol/pull/150


ButterscotchV (approved):
I think the WiFi onboarding might also be broken in general because of the MAC step?

ImUrX:
yeah, i fixed that in the solarxr side, it was because the step was added in the middle of the enum instead of appended

Erimelowo (approved):

PR #1213, opened 2024-10-26
Add randomized theme

ButterscotchV:
idk

### DIFF
--- a/gui/src/AppLayout.tsx
+++ b/gui/src/AppLayout.tsx
@@ -6,6 +6,46 @@ export function AppLayout() {
   const { loading, config } = useConfig();
   const navigate = useNavigate();
 
+  const r = () => {
+    return Math.random() * 255;
+  };
+
+  let style = document.getElementById('random-style');
+
+  if (!style) {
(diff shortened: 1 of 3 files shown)


ImUrX:
idk if this is a good thing, its a theme that changes per page maybe not great for ppl susceptible to seizures.
i just wanted to commit to it as i made the `ThemeSelector` button be animated which looks very cool, the problem is that it uses !important and a selector check to take priority from Tailwind

Eirenliel:
Can it be randomized on boot?

ImUrX:
it could

loucass003:
@ButterscotchV What do we do about this?
Do we do the randomized on boot thingy and send it? 
Or just close it?

ButterscotchV:
> @ButterscotchV What do we do about this?
> Do we do the randomized on boot thingy and send it? 
> Or just close it?

I'm not sure, I don't want to work on this myself

ImUrX:
randomized on boot thing sounds fine, i can do that instead

PR #1214, opened 2024-10-26
Clarify AutoBone error after #1185

ButterscotchV:
The current error message doesn't really convey the issue as the error limit was just patched in. Let's make it say something more useful.

I don't think this is the best message possible since it doesn't really mention anything about an error value limit, though the end user knows nothing about this supposed error value, so maybe it would just be more confusing? I'd love thoughts and suggestions here.

### DIFF
--- a/gui/public/i18n/en/translation.ftl
+++ b/gui/public/i18n/en/translation.ftl
@@ -970,7 +970,7 @@ onboarding-automatic_proportions-done-title = Body measured and saved.
 onboarding-automatic_proportions-done-description = Your body proportions' calibration is complete!
 onboarding-automatic_proportions-error_modal-v2 =
     <b>Warning:</b> There was an error while estimating proportions!
-    This is likely a mounting calibration issue, make sure your tracking works first.
+    This is likely a mounting calibration issue. Make sure your tracking works properly before trying again.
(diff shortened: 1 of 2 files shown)


ImUrX (approved):

ImUrX (changes requested):
please change language key by appending -v2 to it and update any code trying to use the key

ImUrX (approved):

gui/public/i18n/en/translation.ftl
```diff
@@ -968,8 +968,9 @@ onboarding-automatic_proportions-verify_results-redo = Redo recording
 onboarding-automatic_proportions-verify_results-confirm = They're correct
 onboarding-automatic_proportions-done-title = Body measured and saved.
 onboarding-automatic_proportions-done-description = Your body proportions' calibration is complete!
-onboarding-automatic_proportions-error_modal =
-    <b>Warning:</b> An error was found while estimating proportions!
+onboarding-automatic_proportions-error_modal-v2 =
+    <b>Warning:</b> There was an error while estimating proportions!
+    This is likely a mounting calibration issue, make sure your tracking works first.
```
Erimelowo, line 973:
```suggestion
    This is likely a mounting calibration issue. Make sure your tracking works properly before trying again.
```

Erimelowo (approved):

PR #1216, opened 2024-10-29
Fix Drift compensation prediction description

Erimelowo:
more English


Erimelowo:
Actually maybe just “horizontally”? Is a “yaw axis” even a thing? Would like feedback…

Eirenliel (approved):

PR #1222, opened 2024-10-30
Skeleton Constraints

Stermere:
Adds constraints to the joints of the skeleton allowing for unrealistic joint angles to be clamped to more reasonable values and for extreme drift to be compensated for.


server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
```diff
@@ -0,0 +1,183 @@
...
+				ConstraintType.HINGE -> hingeConstraint
+				ConstraintType.LOOSE_HINGE -> looseHingeConstraint
+			}
+
+		private fun decompose(
+			rotation: Quaternion,
+			twistAxis: Vector3,
+		): Pair<Quaternion, Quaternion> {
+			val projection = rotation.project(twistAxis)
+			val twist = Quaternion(sqrt(1.0f - projection.xyz.lenSq()) * if (rotation.w >= 0f) 1f else -1f, projection.xyz).unit()
```
ButterscotchV, line 76: I can't think about this right now, but this feels sus doing `1f - xyx.lenSq()`, can I get an explanation on why this logic is correct? There are also other such uses like this within the code.

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
```diff
@@ -0,0 +1,183 @@
...
+		// Constraint function for TwistSwingConstraint
+		private val twistSwingConstraint: ConstraintFunction =
+			{ rotation: Quaternion, thisBone: Bone, swingRad: Float, twistRad: Float, _: Float ->
+				if (thisBone.parent == null) {
+					rotation
+				} else {
+					val parent = thisBone.parent!!
+					val localRotationOffset = parent.rotationOffset.inv() * thisBone.rotationOffset
+					val rotationLocal =
+						(parent.getGlobalRotation() * localRotationOffset).inv() * rotation
```
ButterscotchV, line 128: This code doesn't make sense being here and it is duplicated between constraint functions, perhaps this would do better in the **Bone** class. I could see this being used in `getLocalRotation()` if the **TransformNode** is `TransformNode.localRotation == false`? I would like an opinion on this.

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
```diff
@@ -0,0 +1,183 @@
...
+			twistAxis: Vector3,
+		): Pair<Quaternion, Quaternion> {
+			val projection = rotation.project(twistAxis)
+			val twist = Quaternion(sqrt(1.0f - projection.xyz.lenSq()) * if (rotation.w >= 0f) 1f else -1f, projection.xyz).unit()
+			val swing = (rotation * twist.inv()).unit()
+			return Pair(swing, twist)
+		}
+
+		private fun constrain(rotation: Quaternion, angle: Float): Quaternion {
+			val magnitude = sin(angle * 0.5f)
```
ButterscotchV, line 82: I would like a comment explaining this if possible, I do not understand why sine is used here or why we multiply by 0.5f. What is the magnitude of?

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
```diff
@@ -0,0 +1,183 @@
...
+					hingeAxisRot = constrain(hingeAxisRot, min, max, Vector3.NEG_X)
+					nonHingeRot = constrain(nonHingeRot, nonHingeDeviation)
+
+					(parent.getGlobalRotation() * localRotationOffset * (nonHingeRot * hingeAxisRot)).unit()
+				}
+			}
+
+		// Constraint function for CompleteConstraint
+		private val completeConstraint: ConstraintFunction = { _: Quaternion, thisBone: Bone, _: Float, _: Float, _: Float ->
+			thisBone.getGlobalRotation()
```
ButterscotchV, line 180: Should we not return the initial rotation provided? I think using the bone is beyond the necessary scope.

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
```diff
@@ -0,0 +1,183 @@
...
+	 */
+	var allowModifications = true
+
+	var initialRotation = Quaternion.IDENTITY
+
+	/**
+	 * Apply rotational constraints and if applicable force the rotation
+	 * to be unchanged unless it violates the constraints
+	 */
+	fun applyConstraint(rotation: Quaternion, thisBone: Bone): Quaternion = constraintFunction(rotation, thisBone, swingRad, twistRad, allowedDeviationRad).unit()
```
ButterscotchV, line 41: Perhaps we could remove Bone from the scope of these constraints entirely and replace `rotation` with `localRotation: Quaternion` as the current bone's local rotation and `thisBone` could be replaced with `parentRotation: Quaternion` as the parent bone's global rotation? I don't think any of the constraints use anything else besides `COMPLETE`, but that could be handled in another way (or maybe we just take current global rotation instead? it could also be handled earlier in the helper functions, see below).

We could instead use helper functions to accept a global current rotation and a Bone, that way we have minimal input for constraints. This would be beneficial for unit testing and reduce complexity and duplication.

server/core/src/main/java/dev/slimevr/tracking/processor/Bone.kt
```diff
@@ -58,6 +58,14 @@ class Bone(val boneType: BoneType) {
 		headNode.update()
 	}
 
+	/**
+	 * Computes the rotations and positions of this bone.
+	 */
+	fun updateThisNode() {
+		headNode.updateThisNode()
+		tailNode.updateThisNode()
+	}
```
ButterscotchV, line 67: This is unsafe in regards to the bone hierarchy. Please see my earlier comment in **HumanSkeleton** regarding a refactor of how constraints are applied.

ButterscotchV (changes requested):
Looking really good, overall good quality code. I think my suggestions could help to keep things maintainable for the future.

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
Stermere: I refactored to get very close to what you suggested here but annoyingly bone still needs to be an input to applyConstraint for Complete constraints to work properly in all use cases. This is because when these constraints are being used in an IK solver the input rotation is not necessarily going to be the same rotation as the bone's rotation. In the way it's being used right now the rotation is the same but it's not always going to be the case.

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
Stermere: Yes, first of all, it's only correct when the quaternion is unitized, so in this case it is very much sus because the input is not unitized (I fixed it now but it was broken good catch). But the idea is that when you have a unitized quaternion the scalar part will be equal to the square root of one minus the vector parts length squared because the total length of the quaternion is one. You have to reconstruct the w component in this way because the vector part's length may have been changed when it was projected onto an axis and if the w component does not bring the length of the quaternion to 1 then the rotation will not be correct. Using just the projected rotation does work but there are some inputs that cause an invalid quaternion to be created which I want to avoid at all costs. I hope that explanation was understandable!

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
Stermere: Check the comment I added but the idea is that 0-180 degrees as input to sin(angle * 0.5) maps directly to the length of the vector part of a rotation rotating angle degrees (again assuming a unitized quaternion). So magnitude is the length the vector part of the rotation would have if it was rotating exactly angle degrees and unitized.

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
Stermere: The refactoring I have done did not move this to the bone class but I am happy to if you still think it is needed :)

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
Stermere: See my other comment on the input rotation not being the same as the bone's global rotation when constraints are being used in an IK solver. I did change to returning the input rotation but the completeConstraint function should never be called due to it being handled before the constraint function is called due to its need for the bone.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -34,45 +36,47 @@ class HumanSkeleton(
...
-	val leftShoulderBone = Bone(BoneType.LEFT_SHOULDER)
-	val rightShoulderBone = Bone(BoneType.RIGHT_SHOULDER)
-	val leftUpperArmBone = Bone(BoneType.LEFT_UPPER_ARM)
-	val rightUpperArmBone = Bone(BoneType.RIGHT_UPPER_ARM)
-	val leftLowerArmBone = Bone(BoneType.LEFT_LOWER_ARM)
-	val rightLowerArmBone = Bone(BoneType.RIGHT_LOWER_ARM)
-	val leftHandBone = Bone(BoneType.LEFT_HAND)
-	val rightHandBone = Bone(BoneType.RIGHT_HAND)
+	val leftUpperShoulderBone = Bone(BoneType.LEFT_SHOULDER, Constraint(ConstraintType.COMPLETE))
+	val rightUpperShoulderBone = Bone(BoneType.RIGHT_SHOULDER, Constraint(ConstraintType.COMPLETE))
```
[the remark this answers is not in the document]
Stermere: Yeah, I completely missed that thanks!

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -372,6 +381,39 @@ class HumanSkeleton(
...
+			val deltaRot = newRot * initialRot.inv()
+			val angle = deltaRot.angleR()
+			val tracker = getTrackerForBone(bone.boneType)
+			val parentTracker = getTrackerForBone(bone.parent?.boneType)
+			if ((angle > 0.01f) && (
+					tracker?.filteringHandler?.getFilteringImpact()
+						?: 1f
+					) < 0.01f && (
+					parentTracker?.filteringHandler?.getFilteringImpact()
+						?: 1f
```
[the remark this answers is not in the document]
Stermere: I'll have to test this, but I think you are right it will be beneficial to have more correction in the vast majority of tracker placement setups.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -372,6 +381,39 @@ class HumanSkeleton(
...
+					) < 0.01f
```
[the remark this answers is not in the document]
Stermere: It may be a bit restrictive but I found in testing that values generally stay far below this when not moving around very fast. I will test a bit more and get back to you

Stermere (commented):
Hope these changes were what you had in mind! If they are not don't hesitate to let me know.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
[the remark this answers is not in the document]
Stermere: I increased the value to 2 degrees (compared to 0.6 degrees previously), which should allow correction more often without sacrificing much accuracy.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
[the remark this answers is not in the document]
Stermere: It seems to be good in testing!

Stermere:
I tested not excluding hinge constraints from correction but it causes the lower leg bone to consistently become off-kilter by 5-15 degrees after a little moving around so I am leaving them excluded for now :(

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
ButterscotchV: Sounds good although the comment is cut off with "since a quaternions"

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
ButterscotchV: Thank you for the additional context, looks good then.

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
```diff
@@ -0,0 +1,190 @@
...
+		val rotationLocal = rotation * initialRotation.inv()
+		var (swingQ, twistQ) = decompose(rotationLocal, Vector3.NEG_Y)
+		swingQ = constrain(swingQ, maxDeviationFromTrackerRad)
+		twistQ = constrain(twistQ, maxDeviationFromTrackerRad)
+		return initialRotation * (swingQ * twistQ)
+	}
+
+	companion object {
+		const val ANGLE_THRESHOLD = 0.004f // == 0.25 degrees
+		const val FILTER_IMPACT_THRESHOLD = 0.0349 // == 2 degrees
```
ButterscotchV, line 76: Was this supposed to be a double? 👀
```suggestion
		const val FILTER_IMPACT_THRESHOLD = 0.0349f // == 2 degrees
```

ButterscotchV (changes requested):
Looks much better, see my previous comments as I've replied to some for why changes are requested. P.S. If possible, merging from main is preferred to update PRs after reviews as otherwise it clobbers history, if it's not possible then it's fine, just makes it a little difficult. Perhaps a quick beta can be done with this to ensure none of the changes have broken anything. TY for the fixes!

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
ButterscotchV: I'm happy with how it is now, I still think it could be pulled out to the Bone class but if you would rather it stays here then it's fine. Resolve this conversation once you've seen this. :3

server/core/src/main/java/dev/slimevr/tracking/processor/Constraint.kt
Stermere: fixed, idk how I pulled that one off.

Stermere:
I can merge after reviews in the future, I didn't realize it messed it up until after I did it. I'll make a beta soon, thanks for the suggestions they made the code heaps better!

ButterscotchV (approved):

Eirenliel:
Pl.ease rebase SolarXR

ButterscotchV:
> Looks much better, see my previous comments as I've replied to some for why changes are requested. P.S. If possible, merging from main is preferred to update PRs after reviews as otherwise it clobbers history, if it's not possible then it's fine, just makes it a little difficult. Perhaps a quick beta can be done with this to ensure none of the changes have broken anything. TY for the fixes!

Just updating on this, I did test the beta myself and it appears to be fine, personally I think it's good to be merged besides SolarXR issues? Unsure what the status of SolarXR stuff is because I know the spine yaw correct PR got merged first.

Stermere:
> Just updating on this, I did test the beta myself and it appears to be fine, personally I think it's good to be merged besides SolarXR issues? Unsure what the status of SolarXR stuff is because I know the spine yaw correct PR got merged first.

Is SolarXR an issue? I belive the unused values for spine yaw comp are just replaced with dummies as of currently.

ButterscotchV:
> > Just updating on this, I did test the beta myself and it appears to be fine, personally I think it's good to be merged besides SolarXR issues? Unsure what the status of SolarXR stuff is because I know the spine yaw correct PR got merged first.
> 
> Is SolarXR an issue? I belive the unused values for spine yaw comp are just replaced with dummies as of currently.

SolarXR should be fine yeah I didn't realize Eiren filled it in.

ImUrX (approved):

ImUrX:
please fix lint

Stermere:
> please fix lint

Done!

loucass003 (approved):

Stermere:
Conflicts resolved

ImUrX (approved):

PR #1234, opened 2024-11-11
Enables a tracker to forward log messages to the server, and the server to send the log messages to the UI.

jabberrock:
1. `PROTOCOL_LOG_SUPPORT` which determines if the server supports forwarding log messages
2. `PACKET_LOG (26)` which is the UDP packet that the tracker sends to the server
3. `Device.logMessages` which is the new field for the UI to receive log messages

Device logs are displayed in the GUI under the specified tracker, because there is no place to put device-specific information, nor is there a good way to separate tracker-specific logs from each other. This means that aux trackers will show the same device logs as the main tracker.

Related PRs:
https://github.com/jabberrock/SlimeVR-Tracker-ESP/pull/1
https://github.com/jabberrock/SlimeVR-Server/pull/1
https://github.com/jabberrock/SolarXR-Protocol/pull/1

image

https://github.com/user-attachments/assets/f592a870-3dff-4236-8826-9416aa17c488

### DIFF
--- a/gui/src/components/tracker/DeviceLogs.tsx
+++ b/gui/src/components/tracker/DeviceLogs.tsx
@@ -1,0 +1,54 @@
+import { useEffect, useRef, memo, Fragment } from 'react';
+
+type LogProps = {
+  messages: string[]
+};
+
+function Log({ messages } : LogProps) {
+  const preRef = useRef(null);
+  const lastScrollTopRef = useRef(0);
+
+  useEffect(() => {
+    const pre = preRef.current;
+    const lastScrollTop = lastScrollTopRef.current;
+    // Scroll to the latest message if either:
+    // - We were looking at the latest message previously; or
+    // - The scroll height shrunk, e.g. when logs are cleared due to device reconnecting
+    if (pre && (pre.scrollTop >= lastScrollTop || lastScrollTop >= pre.scrollHeight)) {
+      pre.scrollTop = pre.scrollHeight;
+      lastScrollTopRef.current = pre.scrollTop;
+    }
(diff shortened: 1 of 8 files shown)


loucass003:
It is cool, i mostly dont like how it looks. I would create another block below the tracker settings with a nice title 
Device Logs

And put the logs text behind a darker color. Similar to the actual serial console in the server.

loucass003:
From the look of the code it also looks like all trackers will be sending their logs to the server no matter what. Imo this is not good. The server should only send the logs through the solarxr protocol when needed. 

You would need a packet to Start listening for logs from the gui and stop listening. 

Also that pose another question. Should the server really be always listening for tracker logs without asking for it. this seem like a lot of packets for simply a debug feature.

Eirenliel:
agree, @loucass003

jabberrock:
**Tracker -> Server**
Each tracker sends about 30 messages at startup, and then 1 message every minute. I don't think that's going to overwhelm the server. When there's a problem with the tracker, you want to be see what happened in the past.

**Server -> UI**
Agreed that the UI shouldn't request the logs unless needed. Would requesting the logs when you're viewing a specific tracker be acceptable?

jabberrock:
I imagine this will simplify debugging problems for customers. It could be super useful for things like calibration. It may lead firmware to put in more logs, e.g. so that customers can figure out why a tracker's acting weird.

Eirenliel:
> **Server -> UI** Agreed that the UI shouldn't request the logs unless needed. Would requesting the logs when you're viewing a specific tracker be acceptable?

@jabberrock yeah

PR #1235, opened 2024-11-11
make HID Device a slime

sctanf:


Eirenliel:
Why doesn't it read manufacturer from the usb info?

Eirenliel:
Also a PR should clearly say what the change does, why isSlime matters? I have no idea what this change means for HID devices that aren't original slimes, for example

sctanf:
> Why doesn't it read manufacturer from the usb info?

This field is used in GUI as `Type`, so I used `HID Device` as the type. If it should be changed to manufacturer or to something else, this should be clearer

sctanf:
> Also a PR should clearly say what the change does, why isSlime matters? I have no idea what this change means for HID devices that aren't original slimes, for example

I should have added, isSlime is how the GUI filters when you select the option to only show HMD and Slime trackers

so there is already DIY slimes or any slime made by third-party, is this an original slime? I don't know what type/manufacturer is appropriate, or if this is a wrong implementation for the intended outcome

Eirenliel (approved):

ImUrX (approved):

ButterscotchV (approved):

PR #1243, opened 2024-11-16
Spine Yaw Compensation to keep spine trackers from drifting

jabberrock:
Prevents the spine trackers from drifting, by nudging them to stay aligned to the headset.

Players are usually standing facing forwards or walking forwards, where the spine trackers are aligned to the yaw of the headset. However due to yaw drift, the spine trackers will eventually rotate to the side, requiring a reset. This feature adds a small centering rotation to compensate for the yaw drift. (For now, this compensation only activates when your trackers are pointing "up", e.g. when you're standing or sitting straight up.)

The correction amount should be approximately the maximum yaw drift of your gyroscope (0.3 degrees/sec is a good default).

image


Eirenliel:
Please rebase SolarXR, it was merged.

Eirenliel:
I think this can be enabled for legs too since it only works when standing up! Ideally, the setting would be part of Drift compensation setting that's now with 3 selectors: Off, Static/Simple (Old one), Dynamic/Advanced (this one). Then for this to have a setting of compensation speed, maybe a slider in 0.05deg/min intervals, and a selector for which parts to apply it: Upper body, Legs, Arms, Feet. There also needs to be an infomation that Dynamic only works with HMD and Static can work without it.

But this can be merged after review already, and settings can be added in the future. Very excited to RC this!

jabberrock:
@Eirenliel thanks taking a look! Definitely can look into putting this into the drift compensation setting later. (Or do you think I should just update it now?)

If this works well, I was going to add alignment between upper leg and lower leg.

But I'm not sure if upper leg can be aligned to the spine. For example, when I stand straight, I notice that my legs are twisted slightly away from my body. I was thinking maybe "balancing" the left and right legs so that they're at the same angle away from the spine...

Eirenliel:
> If this works well, I was going to add alignment between upper leg and lower leg. Maybe even foot?

Upper and lower leg should have the same yaw, so they're already averaged.

> But I'm not sure if upper leg can be aligned to the spine. For example, when I stand straight, I notice that my legs are twisted slightly away from my body. I was thinking maybe "balancing" the left and right legs so that they're at the same angle away from the spine...

It shouldn't matter if you reset the same way too. Also since it's a setting people can turn it off if it doesn't work for them and we can improve it in the future.

>  (Or do you think I should just update it now?)

If you want, we're not merging util Monday anyway.

Also pelase fix build errors (lint)

Eirenliel:
This needs to be Beta'd before it can be merged.

jabberrock:
@Eirenliel How do I get the build system to rebuild, and how does the beta process work? Sorry, it's my first PR. Thanks!

Erimelowo:
Since you're a first-time contributor, someone else needs to add their approval for the CI to run.

jabberrock:
Sigh oops I didn't realize there was a Java linter too

Eirenliel:
For Beta you need to create a post in beta-testing-forum on Discord! If you can't, ask some contributors

server/core/src/main/java/dev/slimevr/config/YawCorrectionConfig.kt
```diff
@@ -0,0 +1,9 @@
+package dev.slimevr.config
+
+class YawCorrectionConfig {
+	private var _amountInDegPerSec = 0.0f
```
Erimelowo, line 4: Could also add a toggle to allow simply toggling the option on/off without having to set it to 0. This would also allow the _amountInDegPerSec to have a sensible default value.

Also get/set are automatically generated in Kotlin and you don't need a private member + public propriety (see how other config classes were done)

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -519,6 +524,72 @@ class HumanSkeleton(
...
+		}
+
+		val deltaRot = trackerRot * parentTrackerRot.inv()
+		val deltaYawInRad = deltaRot.toEulerAngles(EulerOrder.YZX).y
+
+		// Amount of yaw should be roughly the maximum yaw bias of the gyroscope. If it
+		// is too small, the gyroscope will overpower the correction and the skeleton
+		// will drift. If it is too big, the player will notice that the skeleton is
+		// rotating when the player doesn't face forward for a long time.
+		val serverUpdateFreqInHz = 1000.0f; // FIXME: Use actual server update rate
```
Erimelowo, line 578: VRServer has `fpsTimer`

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -519,6 +524,72 @@ class HumanSkeleton(
...
+		for (tracker in trackersToAlign) {
+			if (tracker == null) {
+				continue;
+			}
+
+			if (parentTracker == null) {
+				parentTracker = tracker;
+				continue;
+			}
+
```
Erimelowo, line 545: Check if the tracker `isImu`. If it's not, don't adjust for drift.

Erimelowo (changes requested):

server/core/src/main/java/dev/slimevr/config/YawCorrectionConfig.kt
jabberrock: When JSON deserializes a missing field into the float (e.g. when loading a config from a previous version of the server), it sets it to `NaN`. `@JsonInclude(Include.NON_NULL)` did not seem to work.

Do you know how to fix this?

jabberrock: @Erimelowo could you review https://github.com/SlimeVR/SolarXR-Protocol/pull/155 when you have a chance? Thanks!

jabberrock: Addressed by adding enabled setting.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
jabberrock: Addressed by using `fpsTimer`

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
jabberrock: Addressed by adding `isImu` check before updating the yaw. I chose to allow an IMU tracker to be adjusted to match a non-IMU parent tracker.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -519,6 +523,76 @@ class HumanSkeleton(
...
 	}
 
+	private fun updateTrackerYawCorrections() {
+		val trackersToAlign = listOf(
+			headTracker,
+			upperChestTracker,
+			chestTracker,
+			waistTracker,
+			hipTracker,
+		)
```
ButterscotchV, line 533: This is a ton of allocations (run once per update, aka ~1000 times per second). Ideally we should avoid recreating this, but at the least it should be an array. Perhaps this array should be created in `setTrackersFromList()` and cached for use here?
```suggestion
		val trackersToAlign = arrayOf(
			headTracker,
			upperChestTracker,
			chestTracker,
			waistTracker,
			hipTracker,
		)
```

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -13,6 +13,7 @@ import dev.slimevr.tracking.trackers.TrackerRole
 import dev.slimevr.tracking.trackers.TrackerUtils.getFirstAvailableTracker
 import dev.slimevr.tracking.trackers.TrackerUtils.getTrackerForSkeleton
 import dev.slimevr.util.ann.VRServerThread
+import io.eiren.math.FloatMath.toRad
```
ButterscotchV, line 16: Generally in the code we use `deg * FastMath.DEG_TO_RAD`, this method is probably fine to use tho, so idk what people's opinion on this is. This method is a lot shorter to call which I like.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -519,6 +523,76 @@ class HumanSkeleton(
...
+			}
+
+			updateTrackerYawCorrection(tracker, parentTracker)
+			parentTracker = tracker
+		}
+	}
+
+	private fun isTrackerPointingUp(trackerRot: Quaternion, maxDeviationInRad: Float): Boolean {
+		val trackerUp = trackerRot.sandwich(POS_Y)
+		return trackerUp.angleTo(POS_Y) < maxDeviationInRad
```
ButterscotchV, line 557: Surely we have a cleaner way of doing this? It doesn't necessarily matter, but I figure I'd open a discussion on this.

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -327,6 +330,8 @@ class Tracker @JvmOverloads constructor(
 			filteringHandler.getTrackedRotation()
 		}
 
+		rot = EulerAngles(EulerOrder.YZX, 0.0f, yawCorrectionInRad, 0.0f).toQuaternion() * rot
```
ButterscotchV, line 333: `EulerAngles(EulerOrder.YZX, 0.0f, yawCorrectionInRad, 0.0f).toQuaternion()` should be cached on changes to `yawCorrectionInRad`.

ButterscotchV (changes requested):

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
jabberrock: This function's going away in the next change. What would be a cleaner way?

jabberrock:
@ButterscotchV , I remember you said I should move the `Tracker.yawCorrectionInRad` into `TrackerResetHandler` right?

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
jabberrock: Also, this needs to be applied *after* reset and mounting rotations

ButterscotchV:
> @ButterscotchV , I remember you said I should move the `Tracker.yawCorrectionInRad` into `TrackerResetHandler` right?

Probably should go there, yeah. The definitions of these things are kinda mangle together, but most of the drift corrective type stuff is around there.
Also maybe some of the code from this PR could be moved *out* of the HumanSkeleton class, as it's largely unrelated besides using the trackers from there?

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
ButterscotchV: tbh this is fine

jabberrock:
I realized in SolarXR I named it `yawCorrection`, so I'm going to rename this whole feature "Spine Yaw Correction" for consistency. I updated the names to match.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
jabberrock: Addressed by moving it into the `init` of `SpineYawCorrection` class. I also cleaned it up by removing null and non-IMU trackers. This is updated only when the config is changed from the UI.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
jabberrock: Addressed by switching to the constant.

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
jabberrock: Addressed by updating a private variable `spineYawCorrectionRot` when `spineYawCorrectionInRad` is set.

jabberrock:
Additional changes:
1. Extracted algorithm into `SpineYawCorrection` class
2. Moved applying the yaw correction rotation into `TrackerResetsHandler`
3. Renamed all references of "compensation" to "correction"

jabberrock:
Freshly tested:
1. When not enabled, no yaw correction happens
2. When enabled, and chest & hip trackers are pointing forward, yaw correction happens
3. When chest tracker is pointing forward, but hip tracker is pointing up, no yaw correction happens

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/SpineYawCorrection.kt
```diff
@@ -0,0 +1,76 @@
...
+) {
+
+	private val upperBodyTrackers = filterImuTrackers(
+		skeleton.headTracker,
+		skeleton.neckTracker,
+		skeleton.upperChestTracker,
+		skeleton.chestTracker,
+		skeleton.waistTracker,
+		skeleton.hipTracker,
+	)
```
ButterscotchV, line 24: This needs to be updated when the trackers are updated from `HumanSkeleton.setTrackersFromList()`

ButterscotchV (changes requested):

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/SpineYawCorrection.kt
ButterscotchV: Also does it need to be a list? It feels like an array would suffice still, though it matters less with this usage.

jabberrock: Nice catch! I kept it as a list so that I can use `zipWithNext`.

Addressed by changing `SpineYawCorrection` into a stateful class with `setTrackers` and `setConfig`.

ButterscotchV (approved):

jabberrock:
@Erimelowo , could you take another look? Thanks!

@Eirenliel, it looks like the general sentiment of the beta is positive. Most people have been testing v2 and v3, which is a more sophisticated version of this change. Would it be reasonable to merge this?

ButterscotchV (changes requested):
I've tested with the recommended changes and it seems to work as intended. I had weird results from AutoBone, but I still don't really know whether it's this PR causing it.

gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -184,6 +193,10 @@ export function GeneralSettings() {
     unitDisplay: 'narrow',
     maximumFractionDigits: 2,
   });
+  const degreeFormat = new Intl.NumberFormat(currentLocales, {
+    style: 'unit',
+    unit: 'degree',
+  });
```
ButterscotchV, line 199: We should maybe use our translation framework? Idk about this

gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -813,6 +836,55 @@ export function GeneralSettings() {
...
+                </Typography>
+                <Typography color="secondary">
+                  {l10n.getString(
+                    'settings-general-tracker_mechanics-spine_yaw_correction-amount-description'
+                  )}
+                </Typography>
+                <NumberSelector
+                  control={control}
+                  name="yawCorrectionSettings.amountInDegPerSec"
+                  valueLabelFormat={(value) => degreeFormat.format(value)}
```
ButterscotchV, line 881: This label should say in degrees per second, not just degrees. It is confusing as it is.

ButterscotchV (changes requested):

gui/src/components/settings/pages/GeneralSettings.tsx
ButterscotchV: @ImUrX Can I get your feedback on this 🐱

ImUrX: its fine, i discovered a better integrated way on Fluent recently but for now this is how we been doing it, and if its per second you can do `degree-per-second` for the `unit`

gui/src/components/settings/pages/GeneralSettings.tsx
ButterscotchV: See https://github.com/SlimeVR/SlimeVR-Server/pull/1243#discussion_r1866370498

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/SpineYawCorrection.kt
```diff
@@ -0,0 +1,82 @@
...
+		tracker.resetsHandler.spineYawCorrectionInRad += adjustYawInRad
+	}
+
+	private fun isTrackerPointingUp(trackerRot: Quaternion, maxDeviationInRad: Float): Boolean {
+		val trackerUp = trackerRot.sandwich(POS_Y)
+		return trackerUp.angleTo(POS_Y) < maxDeviationInRad
+	}
+
+	companion object {
+		private fun filterImuTrackers(vararg trackers: Tracker?): List<Tracker> = trackers.filterNotNull().filter { it.isImu() }.toList()
```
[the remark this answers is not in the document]
ButterscotchV: Actually, this probably shouldn't filter null trackers either? I'm not sure you would want to correct a hip tracker with the head tracker when every other tracker is missing. If this is intended behaviour then that is fine.

jabberrock:
Slightly updated version coming soon. Will create a new PR.

PR #1247, opened 2024-11-21
SolarXR IPC Socket

rcelyte:
This patch introduces an IPC interface for communicating with the server via SolarXR ー exposing bone data not available through the existing Protobuf API while remaining more reliable and performant compared to the WebSocket API. Monado will require this interface to implement [XR_FB_body_tracking](https://registry.khronos.org/OpenXR/specs/1.0/man/html/XR_FB_body_tracking.html) while maintaining latency parity with the OpenVR driver (see https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2253).

It also fixes a one line bug in the WebSocket API where messages got clobbered on send, resulting in all but the first SolarXR connection receiving mangled data.

### DIFF
--- a/server/desktop/src/main/java/dev/slimevr/desktop/Main.kt
+++ b/server/desktop/src/main/java/dev/slimevr/desktop/Main.kt
@@ -5,10 +5,11 @@ package dev.slimevr.desktop
 import dev.slimevr.Keybinding
 import dev.slimevr.SLIMEVR_IDENTIFIER
 import dev.slimevr.VRServer
-import dev.slimevr.bridge.ISteamVRBridge
+import dev.slimevr.bridge.Bridge
 import dev.slimevr.desktop.firmware.DesktopSerialFlashingHandler
 import dev.slimevr.desktop.platform.SteamVRBridge
 import dev.slimevr.desktop.platform.linux.UnixSocketBridge
+import dev.slimevr.desktop.platform.linux.UnixSocketRpcBridge
 import dev.slimevr.desktop.platform.windows.WindowsNamedPipeBridge
 import dev.slimevr.desktop.serial.DesktopSerialHandler
 import dev.slimevr.desktop.tracking.trackers.hid.TrackersHID
@@ -119,8 +120,7 @@ fun main(args: Array<String>) {
 		val configDir = resolveConfig()
 		LogManager.info("Using config dir: $configDir")
 		val vrServer = VRServer(
-			::provideSteamVRBridge,
-			::provideFeederBridge,
+			::provideBridges,
 			{ _ -> DesktopSerialHandler() },
 			{ _ -> DesktopSerialFlashingHandler() },
 			configPath = configDir,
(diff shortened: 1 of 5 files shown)


rcelyte:
Marked as draft since I'm not sure I have `try` blocks in all the right places

Eirenliel (approved):

rcelyte:
alright

server/core/src/main/java/dev/slimevr/protocol/ProtocolAPI.java
```diff
@@ -31,6 +32,10 @@ public ProtocolAPI(VRServer server) {
 	}
 
 	public void onMessage(GenericConnection conn, ByteBuffer message) {
+		if (message.position() != 0)
+			message = ByteBuffer
+				.wrap(Arrays.copyOfRange(message.array(), message.position(), message.limit()));
```
ButterscotchV, line 37: Can we not use `ByteBuffer.slice()` here? If required, we can also further do `ByteBuffer.asReadOnlyBuffer()` but I believe it should be fine without.
```suggestion
		if (message.position() != 0)
			message = message.slice();
```

ButterscotchV (commented):
Looks fine

server/core/src/main/java/dev/slimevr/protocol/ProtocolAPI.java
rcelyte: I should add a comment explaining why this is necessary. The issue isn't mutability, it's Flatbuffers generated code failing to account for starting position properly. By using `Arrays.copyOfRange()` I allocate a new backing array without the data offset.

ButterscotchV: Slice should set the starting position to 0 though, so it should work the same? I don't think flatbuffers ever directly accesses the backing array.

rcelyte: Hmm... I know enough to know I don't know enough. What I do know is that packets fail to parse without the `copyOfRange()` ー blocking Monado from starting entirely ー and that `slice()`, `duplicate()`, and `asReadOnlyBuffer()` don't help.

server/desktop/src/main/java/dev/slimevr/desktop/Main.kt
```diff
@@ -151,90 +151,89 @@ fun main(args: Array<String>) {
...
+					server,
+					"steamvr_feeder",
+					"SteamVR Feeder Bridge",
+					"""\\.\pipe\SlimeVRInput""",
+					FastList(),
+				),
 			)
 		}
 
 		OperatingSystem.LINUX -> {
```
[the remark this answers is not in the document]
rcelyte: Yes, but the SolarXR driver specifically hasn't been ported yet.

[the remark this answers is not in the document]
rcelyte: MacOS support is still in development, with a lot of outdated work needing rebasing (tracked at https://gitlab.freedesktop.org/monado/monado/-/issues/318). There also hasn't been much interest in getting real hardware working on non-Linux platforms yet, so most drivers (including the SolarXR bridge) remain Linux exclusive.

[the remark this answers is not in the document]
ImUrX: oki!

server/core/src/main/java/dev/slimevr/protocol/ProtocolAPI.java
Eirenliel: Okay but this needs to be investigated, otherwise it will create so much garbage 😢

rcelyte: Turns out this was just a skill issue on my part. Changing `dst.slice()` to `dst.duplicate()` in `UnixSocketConnection.read()` fixed it.

Eirenliel (approved):

server/desktop/src/main/java/dev/slimevr/desktop/platform/linux/UnixSocketConnection.java
```diff
@@ -0,0 +1,120 @@
...
+
+	public void next() {
+		System
+			.arraycopy(
+				dst.array(),
+				dst.arrayOffset() + dst.limit(),
+				dst.array(),
+				dst.arrayOffset(),
+				remainingBytes
+			);
```
ButterscotchV, line 116: This feels like a weird thing to have to do with ByteBuffer, though I can understand why it's done. Just to confirm, this is to shift the buffer so you don't hit the limit, correct?

ButterscotchV (approved):

server/desktop/src/main/java/dev/slimevr/desktop/platform/linux/UnixSocketConnection.java
rcelyte: Correct. `channel.read()` could return partial packets or more than one packet, so trailing data needs to be shifted over.

I'm considering rewriting the function as this:
```java
public void next() {
	dst.position(dst.limit());
	dst.limit(dst.limit() + remainingBytes);
	dst.compact();
	dst.limit(dst.capacity());
}
```
Would that be equivalent to the current code?

ButterscotchV (approved):
Looks great now

ImUrX (approved):

PR #1242, opened 2024-11-15
Flex resistance and angle support

Erimelowo:
I got tired of maintaining the old branch that got fucked up with 60+ commits. It is still there with old git history if needed: https://github.com/SlimeVR/SlimeVR-Server/pull/1095

Implements flex sensor support for UDP trackers.
This will be able to be used for any joint with a single (more or less...) degree of freedom such as the knee, fingers or even shoulders. Main use is for fingers.
GizmoQC's working repo for flex gloves firmware: https://github.com/Guizmo12/gizmoglovesmocap

Needs https://github.com/SlimeVR/SolarXR-Protocol/pull/149  
Based off https://github.com/SlimeVR/SlimeVR-Server/pull/1091
Closes https://github.com/SlimeVR/SlimeVR-Server/issues/961

PR #1251, opened 2024-11-26
Update to latest SolarXR-Protocol

jabberrock:
I just fixed the broken identifiers, and added a dummy `YawCorrectionSettings` entry.

@Erimelowo I think the parent/child ordering of the thumb segments are wrong. It should be metacarpal -> proximal -> distal. Some of the identifiers like `body:left_thumb_intermediate` need to be updated too. I didn't know how to fix it properly since I didn't review the original change.

https://github.com/SlimeVR/SolarXR-Protocol/pull/154
https://github.com/SlimeVR/SolarXR-Protocol/pull/155
https://github.com/SlimeVR/SolarXR-Protocol/pull/156


Erimelowo:
This got updated in the flex branch…

Erimelowo:
implemented in https://github.com/SlimeVR/SlimeVR-Server/pull/1242

PR #1255, opened 2024-12-07
Don't filter non-temporal trackers

ButterscotchV:
This broke AutoBone, we should not modify the rotation from it, though this may not be desired depending on what we want to define computed trackers as.


Eirenliel (approved):

ImUrX:
@Erimelowo can u check this pls

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -329,9 +331,11 @@ class Tracker @JvmOverloads constructor(
 		var rot = if (allowFiltering && filteringHandler.filteringEnabled) {
```
Erimelowo, line 331:
```suggestion
		var rot = if (allowFiltering && filteringHandler.filteringEnabled && !isComputed) {
```

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -361,9 +365,11 @@ class Tracker @JvmOverloads constructor(
 		var rot = if (filteringHandler.filteringEnabled) {
```
Erimelowo, line 365:
```suggestion
		var rot = if (filteringHandler.filteringEnabled && !isComputed) {
```

Erimelowo (changes requested):
might want to add those conditions for consistency's sake?

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
ButterscotchV: This would block smoothing on Vives in entirety though wouldn't it? Maybe this should be a separate flag? Or it could be linked to `allowFiltering` instead perhaps.

Eirenliel:
@Erimelowo can you recheck it please?

ButterscotchV:
Once again changing how this PR works lol, I think this is the safest way of doing this, just a simple added flag and it explains what it's for in the comment :3

ImUrX (approved):

Erimelowo (approved):
lgtm

Eirenliel (approved):

PR #1261, opened 2024-12-11
Refactor + Check for Env variables

### LINKED ISSUE #1260: JVM Environment Variables can crash our JVM
We should seriously check for `_JAVA_OPTIONS` and `JAVA_TOOL_OPTIONS` environment variables in our GUI startup, and at a minimum warn that they are set. I have seen multiple cases of people having corrupt values set for those and it is preventing launching our JVM with no obvious reason.

VocalFan:
Fixes #1260 

Before attempting to launch the server, the GUI will check if `_JAVA_OPTIONS` or `JAVA_TOOL_OPTIONS` are set, and if so, make a warning message appear for the user, along with outputting to the log file. 

Since I was already working on that part of the code, I also refactored the main.rs file by making it where everything was in their own functions instead of messily shoved into the Main function.

Screenshot 2024-12-11 172945

image

### DIFF
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -104,10 +104,9 @@ fn setup_logger(context: &tauri::Context) -> Result<flexi_logger::LoggerHandle> 
 
 	// Based on https://docs.rs/tauri/2.0.0-alpha.10/src/tauri/path/desktop.rs.html#238-256
 	#[cfg(target_os = "macos")]
-	let path = dirs_next::home_dir().ok_or(Error::UnknownPath).map(|dir| {
-		dir.join("Library/Logs")
-			.join(&context.config().identifier)
-	});
+	let path = dirs_next::home_dir()
+		.ok_or(Error::UnknownPath)
+		.map(|dir| dir.join("Library/Logs").join(&context.config().identifier));
 
 	#[cfg(not(target_os = "macos"))]
 	let path = dirs_next::data_dir()
@@ -163,28 +162,24 @@ fn setup_webview2() -> Result<()> {
 }
 
 fn check_environment_variables() {
-	let java_options = env::var("_JAVA_OPTIONS").ok();
-	let java_tool_options = env::var("JAVA_TOOL_OPTIONS").ok();
-
-	if java_options.is_some() || java_tool_options.is_some() {
-		let mut env_warn_msg = String::new();
-
-		if let Some(env) = java_options {
-			string_append_list_style(&mut env_warn_msg, "_JAVA_OPTIONS");
-			log::warn!("_JAVA_OPTIONS is set: {}", env);
-		}
-
-		if let Some(env) = java_tool_options {
-			string_append_list_style(&mut env_warn_msg, "JAVA_TOOL_OPTIONS");
-			log::warn!("JAVA_TOOL_OPTIONS is set: {}", env);
-		}
-
-		// Display the warning dialog
+	use itertools::Itertools;
+	const ENVS_TO_CHECK: &[&str] = &["_JAVA_OPTIONS", "JAVA_TOOL_OPTIONS"];
+	let checked_envs = ENVS_TO_CHECK
+		.into_iter()
+		.filter_map(|e| {
+			let Ok(data) = env::var(e) else {
+				return None;
+			};
+			log::warn!("{e} is set to: {data}");
+			Some(e)
+		})
+		.join(", ");
+	if !checked_envs.is_empty() {
 		rfd::MessageDialog::new()
-            .set_title("SlimeVR")
-            .set_description(&format!("You have environment variables {} set, which may cause the SlimeVR Server to fail to launch properly.", env_warn_msg))
-            .set_level(rfd::MessageLevel::Warning)
-            .show();
+			.set_title("SlimeVR")
+			.set_description(&format!("You have environment variables {} set, which may cause the SlimeVR Server to fail to launch properly.", checked_envs))
+			.set_level(rfd::MessageLevel::Warning)
+			.show();
 	}
 }
 
@@ -399,12 +394,4 @@ fn tauri_build_result(
 			show_error(&error.to_string());
 		}
 	}
-}
-
-// Allows text to be added in a list style format.
-fn string_append_list_style(string: &mut String, append: &str) {
-	if !string.is_empty() {
-		string.push_str(", ");
-	}
-	string.push_str(append);
 }
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -127,7 +127,7 @@ fn setup_logger(context: &tauri::Context) -> Result<flexi_logger::LoggerHandle> 
 		.write_mode(WriteMode::BufferAndFlush)
 		.start()?)
 }
-
+#[cfg(windows)]
 fn setup_webview2() -> Result<()> {
 	use crate::util::webview2_exists;
 	use win32job::{ExtendedLimitInfo, Job};
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -156,7 +156,6 @@ fn setup_webview2() -> Result<()> {
 		if confirm == MessageDialogResult::Ok {
 			open::that("https://docs.slimevr.dev/server-setup/installing-and-connecting.html#install-the-latest-slimevr-installer").unwrap();
 		}
-		return Ok(());
 	}
 	Ok(())
 }
@@ -174,6 +173,7 @@ fn check_environment_variables() {
 			Some(e)
 		})
 		.join(", ");
+
 	if !checked_envs.is_empty() {
 		rfd::MessageDialog::new()
 			.set_title("SlimeVR")
@@ -336,7 +336,6 @@ fn tauri_build_result(
 	exit_flag: Arc<AtomicBool>,
 	backend: Arc<Mutex<Option<CommandChild>>>,
 ) {
-	let backend_termination = backend.clone();
 	match build_result {
 		Ok(app) => {
 			app.run(move |app_handle, event| match event {
@@ -350,7 +349,7 @@ fn tauri_build_result(
 						Err(e) => log::error!("failed to save window state: {}", e),
 					}
 
-					let mut lock = backend_termination.lock().unwrap();
+					let mut lock = backend.lock().unwrap();
 					let Some(ref mut child) = *lock else { return };
 					let write_result = child.write(b"exit\n");
 					match write_result {
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -177,7 +177,7 @@ fn check_environment_variables() {
 	if !checked_envs.is_empty() {
 		rfd::MessageDialog::new()
 			.set_title("SlimeVR")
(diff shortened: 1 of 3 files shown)


gui/src-tauri/src/main.rs
```diff
@@ -68,88 +69,128 @@ fn main() -> Result<()> {
...
+		if confirm == MessageDialogResult::Ok {
+			open::that("https://docs.slimevr.dev/server-setup/installing-and-connecting.html#install-the-latest-slimevr-installer").unwrap();
+		}
+		return Ok(());
+	}
+	return Ok(());
+}
+
+fn check_environment_variables() {
+	let java_options = var("_JAVA_OPTIONS").ok();
```
ImUrX, line 165: why import `var`? whats wrong with importing env and doing env::var which is much more readable

gui/src-tauri/src/main.rs
```diff
@@ -68,88 +69,128 @@ fn main() -> Result<()> {
...
+		.rotate(
+			Criterion::Age(Age::Day),
+			Naming::Timestamps,
+			Cleanup::KeepLogFiles(2),
+		)
+		.duplicate_to_stderr(Duplicate::All)
+		.write_mode(WriteMode::BufferAndFlush)
+		.start()?)
+}
+#[cfg(windows)]
```
ImUrX, line 130:
```suggestion

#[cfg(windows)]
```

gui/src-tauri/src/main.rs
```diff
@@ -159,19 +200,29 @@ fn main() -> Result<()> {
...
+			show_error(&format!(
+                "Couldn't find a compatible Java version, please download Java {} or higher",
+                MINIMUM_JAVA_VERSION
+            ));
+			return Ok(None);
 		};
 
 		log::info!("Using Java binary: {:?}", java_bin);
-		Some((java_bin, p))
+		return Ok(Some((java_bin, p)));
```
ImUrX, line 211: is this return needed?

gui/src-tauri/src/main.rs
```diff
@@ -68,88 +69,128 @@ fn main() -> Result<()> {
...
+			log::warn!("JAVA_TOOL_OPTIONS is set: {}", env);
+		}
+
+		// Display the warning dialog
+		rfd::MessageDialog::new()
+            .set_title("SlimeVR")
+            .set_description(&format!("You have environment variables {} set, which may cause the SlimeVR Server to fail to launch properly.", env_warn_msg))
+            .set_level(rfd::MessageLevel::Warning)
+            .show();
+	}
```
ImUrX, line 187: not a big fan of this, maybe via filter map it would look nicer? let's try

```suggestion
	const ENVS_TO_CHECK: &[&str] = &["_JAVA_OPTIONS", "JAVA_TOOL_OPTIONS"]; 
	let checked_envs = ENVS_TO_CHECK.filter_map(|e| {
		let Some(data) = var(e) else { return None; };
		log::warn!("{e} is set to: {data}");
		Some(e)
	}).join(", ");

	if !checked_envs.is_empty() {
		rfd::MessageDialog::new()
			.set_title("SlimeVR")
			.set_description(&format!("You have environment variables {checked_envs} set, which may cause the SlimeVR Server to fail to launch properly."))
			.set_level(rfd::MessageLevel::Warning)
			.show();
	}
```

gui/src-tauri/src/main.rs
```diff
@@ -68,88 +69,128 @@ fn main() -> Result<()> {
...
+			.set_description("Couldn't find WebView2 installed. You can install it with the SlimeVR installer")
+			.set_buttons(MessageButtons::OkCancel)
+			.set_level(MessageLevel::Error)
+			.show();
+		if confirm == MessageDialogResult::Ok {
+			open::that("https://docs.slimevr.dev/server-setup/installing-and-connecting.html#install-the-latest-slimevr-installer").unwrap();
+		}
+		return Ok(());
+	}
+	return Ok(());
```
ImUrX, line 161: again return not needed
```suggestion
	Ok(())
```

ImUrX (changes requested):

gui/src-tauri/src/main.rs
VocalFan: It is:
```
mismatched types
expected enum `Result<std::option::Option<_>, ErrReport>`
   found enum `std::option::Option<_>`
```

VocalFan: Ah, no, if I remove the return, we're good XD

ImUrX: That's the type, not the `return ` part

VocalFan: Yeah I realized that lol

gui/src-tauri/src/main.rs
VocalFan: That code doesn't even compile @w@;

VocalFan: Well first I get

```
no method named `filter_map` found for reference `&'static [&'static str]` in the current scope
`&'static [&'static str]` is not an iterator
```
Which adding .into_iter() `ENVS_TO_CHECK.into_iter().filter_map` fixes

Then `Some(data)` gives me this:

```
mismatched types
expected enum `Result<std::string::String, VarError>`
   found enum `std::option::Option<_>`
```

Then `.join` gives me this:

```
no method named `join` found for struct `FilterMap` in the current scope
method not found in `FilterMap<Iter<'_, &str>, {closure@main.rs:167:58}>`
```

ImUrX: yeah sadly the join part requires itertools, i forgor, well pull it as a dependency, its pretty useful for iterating

ImUrX: and this is with the other errors fixed
```rs
	const ENVS_TO_CHECK: &[&str] = &["_JAVA_OPTIONS", "JAVA_TOOL_OPTIONS"];
	let checked_envs = ENVS_TO_CHECK.into_iter().filter_map(|e| {
		let Ok(data) = env::var(e) else { return None; };
		log::warn!("{e} is set to: {data}");
		Some(e)
	}).join(", ");
	if !checked_envs.is_empty() {
		rfd::MessageDialog::new()
			.set_title("SlimeVR")
			.set_description(&format!("You have environment variables {checked_envs} set, which may cause the SlimeVR Server to fail to launch properly."))
			.set_level(rfd::MessageLevel::Warning)
			.show();
	}
```

VocalFan: > yeah sadly the join part requires itertools, i forgor, well pull it as a dependency, its pretty useful for iterating

That moment when rust is... *Terrible joke here* Illiterate.

VocalFan:
image
image
image
@ImUrX Well, seems like all 3 cases of env variables work out -w-

VocalFan:
Easy way to dev test this:
```rs
unsafe {
    env::set_var("_JAVA_OPTIONS", "Test!");
}
```

gui/src-tauri/src/main.rs
```diff
@@ -339,6 +394,4 @@ fn main() -> Result<()> {
 			show_error(&error.to_string());
 		}
 	}
-
-	Ok(())
-}
+}
```
ImUrX, line 397:
```suggestion
}

```

gui/src-tauri/src/main.rs
```diff
@@ -159,19 +196,29 @@ fn main() -> Result<()> {
 			.then(|| jre.into_os_string())
 			.or_else(|| valid_java_paths().first().map(|x| x.0.to_owned()));
 		let Some(java_bin) = java_bin else {
-			show_error(&format!("Couldn't find a compatible Java version, please download Java {} or higher", MINIMUM_JAVA_VERSION));
-			return Ok(());
+			show_error(&format!(
```
ImUrX, line 199: use `formatcp!` instead as it can be done constant-time

gui/src-tauri/src/main.rs
```diff
@@ -281,7 +328,15 @@ fn main() -> Result<()> {
...
-		.build(tauri_context);
+		.build(context)
+}
+
+fn tauri_build_result(
+	build_result: Result<tauri::App, tauri::Error>,
+	exit_flag: Arc<AtomicBool>,
+	backend: Arc<Mutex<Option<CommandChild>>>,
+) {
+	let backend_termination = backend.clone();
```
ImUrX, line 339: there is no need to clone this as u just use it on one part of the code

gui/src-tauri/src/main.rs
```diff
@@ -68,88 +69,124 @@ fn main() -> Result<()> {
...
+		.into_iter()
+		.filter_map(|e| {
+			let Ok(data) = env::var(e) else {
+				return None;
+			};
+			log::warn!("{e} is set to: {data}");
+			Some(e)
+		})
+		.join(", ");
+	if !checked_envs.is_empty() {
```
ImUrX, line 177:
```suggestion

	if !checked_envs.is_empty() {
```

gui/src-tauri/src/main.rs
```diff
@@ -68,88 +69,124 @@ fn main() -> Result<()> {
...
+		let confirm = MessageDialog::new()
+			.set_title("SlimeVR")
+			.set_description("Couldn't find WebView2 installed. You can install it with the SlimeVR installer")
+			.set_buttons(MessageButtons::OkCancel)
+			.set_level(MessageLevel::Error)
+			.show();
+		if confirm == MessageDialogResult::Ok {
+			open::that("https://docs.slimevr.dev/server-setup/installing-and-connecting.html#install-the-latest-slimevr-installer").unwrap();
+		}
+		return Ok(());
```
ImUrX, line 159:
```suggestion
```
this is not needed as it will leave the if and also return `Ok`

ImUrX (changes requested):

gui/src-tauri/src/main.rs
VocalFan: Yeah... Found that part of the original code kinda weird.

gui/src-tauri/src/main.rs
VocalFan: Uh, you know formatcp! is a compile-time thing, right? But since we use a runtime variable, that won't exactly... work.

VocalFan:
```
the trait bound `std::string::String: From<&&str>` is not satisfied
required for `&&str` to implement `Into<std::string::String>`

attempt to use a non-constant value in a constant
```

ImUrX: its because you doing `&formatcp!` instead of just `formatcp!`

VocalFan: This is what I get for waking up 20 mins ago

VocalFan: @ImUrX That... Still gets you `attempt to use a non-constant value in a constant` because of `checked_envs`

ImUrX: only use formatcp on formats with only constant values

gui/src-tauri/src/main.rs
```diff
@@ -186,6 +186,7 @@ fn check_environment_variables() {
 fn execute_server(
 	cli: Cli,
 ) -> Result<Option<(std::ffi::OsString, std::path::PathBuf)>> {
+	use shadow_rs::formatcp;
```
ImUrX, line 189:
```suggestion
	use const_format::formatcp;
```

VocalFan:
...Yey :3

loucass003 (approved):

PR #1262, opened 2024-12-14
Improve log file behaviour

ButterscotchV:
I hate how logs are. Now they're better™??

## Changes
- Implemented our own FileHandler-like system
- Logs are now `tag_date_time_count.log`
  - `tag` is specified when making the FileLogHandler instance (`slimevr-server` to be more descriptive)
  - `date_time` is the date and time from when you started the server, this could be changed but I think this makes the most sense
  - `count` is the file chunk from the specified limit
- Logs can be limited any variation of:
  - Current file size (default: `25000000` (25 MB))
  - Collective file count (default: `2`)
  - Collective file size (default: `-1` (disabled))
- Logs over the limits are deleted by the earliest file, found by parsing file names
- No more `log_last_0.log.lck`, `log_last_0.log.1`, or `log_last_0.log.1.lck` left sitting around
- No more losing old logs the second you launch a new server instance!
  - Collective file size could be very useful for storing a reasonably sized history of logs to help with debugging user issues, perhaps we could try using it?

### Splitting with 2 count limit
image
### Cumulative file size with splitting
image-1


ImUrX:
why is it on java?

ButterscotchV:
> why is it on java?

Code surrounding it is Java, I can make the added class Kotlin if you want

ImUrX:
its okay, just asking

server/core/src/main/java/io/eiren/util/logging/FileLogHandler.java
```diff
@@ -0,0 +1,265 @@
+package io.eiren.util.logging;
```
Eirenliel, line 1: probably should move it all to dev.slimevr package tbh

Eirenliel (approved):

ImUrX (approved):

PR #1265, opened 2024-12-16
Fix "trackers still on" when closing via tray

### LINKED ISSUE #1264: SlimeVR believes trackers are still connected despite all trackers being turned off.
image
Despite all the trackers being disconnected/turned off, along with SteamVR being fully closed (and I have tried PC restarts)

image
When trying to exit, SlimeVR warns me I have trackers still on.

JovannMC:
When the user chooses to minimize the app to the system tray and tries to exit the app, the "Trackers still on" warning would still appear even if all trackers were disconnected. Updates the tray code to filter the trackers correctly.

Also adds `v0.13.2` to the metainfo xml file

Fixes #1264.


ImUrX (approved):

VocalFan:
So I DID find an actual bug! Nice- Oh, and it does work :3

JovannMC:
Haha yeah, I was very confused and tried literally every single possible combination to reproduce it and couldn't. Ended up looking into the code itself and found the inconsistency

dev.slimevr.SlimeVR.metainfo.xml
loucass003, line 1: Why is this changed?

VocalFan: To be fair, seems like it was just kinda... Forgotten to be added when the release first came out.

ImUrX: i update it when i remember or when we release new version lol

JovannMC: i noticed that 0.13.2 was missing from the file, so i just added it in along with the fix
should i remove it?

ImUrX: i dont really care, it should be done on another PR technically

ButterscotchV (commented):
This shouldn't be needed. Why are there still trackers timed out in the list even if it's properly disconnected? We don't necessarily *need* to show the warning, but it feels like just timed out should still be considered as "connected".

FluffyOMC:
> This shouldn't be needed. Why are there still trackers timed out in the list even if it's properly disconnected? We don't necessarily *need* to show the warning, but it feels like just timed out should still be considered as "connected".

Idk, I mean, all I do is flick the power switch to turn off the official slimes and... Yeah...

Eirenliel (approved):

loucass003 (approved):

Issue #1124, opened 2024-07-26
Hip snaps around in some specific angles


kruemmelbande:
Sometimes the hip snaps around at odd angles. Here i am reproducing it by just moving the hip tracker by hand, although one report suggests that this is happening when sitting down/laying down as well.

Original report in game:
https://github.com/user-attachments/assets/7bf82317-67cd-4a22-807b-0bb1a56e4c6c

My reproduction:
https://github.com/user-attachments/assets/9de00966-6ed0-45b6-8bc0-aa987ba57736

JovannMC:
I can confirm I have this issue as well, just had it again today. I also experience it sometimes when I sit/lay - one session I had in VR had me being literally unable to lay down because my body would snap upwards, where I ended up finding it was the hip tracker causing it and disconnected it.

I thought it was something to do with my setup (I'm using the HaritoraX trackers via SlimeTora), but glad to see this isn't just an issue with me.

kruemmelbande:
Hmm, weird. Ive seen this happen just for me for a little bit, but i never thought much of it until i heard somebody complain in the discord about it actually happening during gameplay... i wonder how many people actually have something like that...

Erimelowo:
That's due to interpolation. Normally, it should allow 270 degrees foward and 90 degrees backwards. This issue can be solved completely by properly keeping track of rotation, which I'm doing in the `fingers` branch...

PR #1277, opened 2025-01-02
Fix over 180 degrees rotations in HumanSkeleton

Erimelowo:
Before we had the constraint that quaternions could only go up to 180 degrees, and we had a hack to make it 270 degs forward and 90 degs backwards in certain situations in HumanSkeleton.
However, now that quaternions are temporarly tracked, we don't have that constraint anymore so we can't just remove the hack.

@ButterscotchV would this be a problem with Autobone? May need testing with that.

closes https://github.com/SlimeVR/SlimeVR-Server/issues/1124


ButterscotchV (approved):
Will test with AutoBone later, should be fine though

Eirenliel (approved):
The code looks fine, but can't check the logic since I have no idea what's going on here 😂 Good job though.

ImUrX (approved):

ImUrX:
guess we merge

ButterscotchV:
Small issue with extended spine model... Didn't do anything specific to cause this, just randomly happened when I was fairly upright. This only happens with the aformentioned option on, a full reset fixes it.

image
image

PR #1278, opened 2025-01-07
Attempt to fix moving average quaternions not resetting properly

Erimelowo:
- Fixed the rotBuffer not being cleared in QuaternionMovingAverage:resetQuats
- Refactored and cleaned up old code since we don't need filtered and non-filtered rotation at the same time (should improve performance)


ButterscotchV (approved):

ImUrX (approved):
I hate the removal of lateinit

Erimelowo:
> I hate the removal of lateinit

I don’t think lateinit makes sense here since it’s not always initialized and can make debugging harder since the IDE won’t warn you about null safety

PR #1283, opened 2025-01-18
Reinstantiate QuaternionMovingAverage on reset

Erimelowo:
Try to fix the centaur issue by reinstantiating the QuaternionMovingAverage class intead of resetting its properties


Eirenliel:
I beg you to put some description.

jabberrock:
I'm applying this to Stay Aligned v6.0 to see if the wider audience can repro the bug. I was able to repro it once, so I don't think it's fully fixed, but it seems to be much rarer.

jabberrock:
For context, when yaw/full resetting, sometimes the skeleton turns into a centaur. This used to happen quite consistently when going through Autobone (which does a full reset at the end). Now it seems to be rarer, but not completely fixed.

image

Erimelowo:
Sure, this didn’t fix anything anyways it seems

PR #1290, opened 2025-01-28
Implement skeleton constraint toggles

ButterscotchV:
Looks like this was missed in #1222, whoops! Tested this on my own with a rotating chest tracker and stationary hip tracker, it seems to work as intended.


Stermere (approved):
Oops, I am surprised this didn't get caught earlier!

Eirenliel (approved):

PR #1294, opened 2025-02-02
Improve reset sounds

loucass003:
Change reset sounds so you have clues at the begining and end of the reset action.
During my last vr session i kept messing up my reset because you would only have a long sound clue at the begining of the reset. Implying that the end of the sound was the end of the reset sequence. And it is not! the reset sequence ends 1s after the sound.


ButterscotchV:
I like the idea but it's hard to differentiate the different resets just from the sounds, perhaps a two note "completed" tone for the full and mounting resets would help?

Issue #1297, opened 2025-02-03
Skeleton constraint bugs


ButterscotchV:
~~I found one issue when testing, which is if you are at the edge of a constraint already and pushing against it, then you reset, it will treat you as still at the edge of a constraint. You can test this super easily by having a HMD and rotating a hip tracker at a constant speed in some direction, you'll see it reach the constraint then if you yaw reset, it will remain as though it is constrained.~~
Edit: Can't seem to reproduce anymore

The second thing is there may be an issue with the correction not actually being cleared on reset, or there may be an issue because correction is not cleared when the feature is disabled. This needs to be looked into, current workaround is to restart SlimeVR.

PR #1303, opened 2025-02-09
Disable correct with constraints server side and hide option in GUI

Stermere:
Should ideally fix https://github.com/SlimeVR/SlimeVR-Server/issues/1297.


ButterscotchV (approved):
Tested last night for a couple hours and this seems good to me. Did not encounter any issues.

Issue #1298, opened 2025-02-04
Body proportions revert to old proportions when trackers connect/disconnect


jabberrock:
1. Manual Proportions
2. Check the arm length (A)
3. Change arm length to be longer (B)
4. Turn on or turn off tracker
5. Skeleton changes arm length back to the original value (A)
6. vrconfig.yml still shows (B)

Something internally must be saving a copy of the proportions, and overwriting it when a tracker connects/disconnects

PR #1304, opened 2025-02-10
Updates nodeOffset when body proportions change

jabberrock:
After a user changes the body proportions, when a tracker connects or disconnects, the skeleton reverts back to the original skeleton. This is because nodeOffset is not updated when the user changes the body proportions, and stores the value at startup. When a tracker connects or disconnects, updateNodeOffsetsInSkeleton overwrites the skeleton's node offsets using the original values. This causes the skeleton to revert back to the original skeleton.

Fixes https://github.com/SlimeVR/SlimeVR-Server/issues/1298

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/processor/config/SkeletonConfigManager.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/processor/config/SkeletonConfigManager.kt
@@ -156,14 +156,8 @@ class SkeletonConfigManager(
 	}
 
 	protected fun setNodeOffset(nodeOffset: BoneType, x: Float, y: Float, z: Float) {
-		var offset = nodeOffsets[nodeOffset]
-
-		if (offset == null) {
-			offset = Vector3(x, y, z)
-			nodeOffsets[nodeOffset] = offset
-		} else {
-			offset = Vector3(x, y, z)
-		}
+		val offset = Vector3(x, y, z)
+		nodeOffsets[nodeOffset] = offset
 
 		// Updates in skeleton
 		humanPoseManager?.updateNodeOffset(nodeOffset, offset)


Eirenliel (approved):
This was probably an error from migration to kotlin, it probably was offset.set(x)

ButterscotchV (approved):
Looks good, seems like this comes from a mix of legacy code (old Vector lib), not updating SkeletonConfigManager correctly, and other skeleton code being changed in behaviour significantly. This change appears to be my original intention with this function.

I believe this bug only started presenting with my changes here: https://github.com/SlimeVR/SlimeVR-Server/commit/518fe8d2efa434b315a2737fe25b181a385dfbba#diff-a7fed76c138eff6f5a44be7e41e654c6560a093f8585ae6a23ae1f4bb52d378dR302

ButterscotchV:
> Looks good, seems like this comes from a mix of legacy code (old Vector lib), not updating SkeletonConfigManager correctly, and other skeleton code being changed in behaviour significantly. This change appears to be my original intention with this function.
> 
> I believe this bug only started presenting with my changes here: [518fe8d#diff-a7fed76c138eff6f5a44be7e41e654c6560a093f8585ae6a23ae1f4bb52d378dR302](https://github.com/SlimeVR/SlimeVR-Server/commit/518fe8d2efa434b315a2737fe25b181a385dfbba#diff-a7fed76c138eff6f5a44be7e41e654c6560a093f8585ae6a23ae1f4bb52d378dR302)

To further "paper-trail" this, this is the original change that caused this issue: https://github.com/SlimeVR/SlimeVR-Server/commit/d4068e7ccb6e6ba32aa8c09f4764c90fa5a0777a#diff-47a2e674419e78de69cfdb92d5d570d1f04de7c2c96ab84454b1dcd491b3beaeL194

PR #1324, opened 2025-03-04
Remove Vive Emulation

ButterscotchV:
I think this joke is a bit tired and Vives don't really do this anymore. People are also starting to be confused by this option for some reason even though it very explicitly says what it is, maybe we should just remove it?

Related to https://github.com/SlimeVR/SolarXR-Protocol/pull/161


Erimelowo:
Agreed and I would be tempted to remove it from solarxr as well while we can

ButterscotchV:
I've created a PR on SolarXR at https://github.com/SlimeVR/SolarXR-Protocol/pull/161 if we want to go down that route. If we do, then I can modify this PR to remove the SolarXR part too.

Eirenliel:
Let's remove it from SolarXR and also make the server completely not run it even if it's left in the config.

Eirenliel (approved):

PR #1329, opened 2025-03-06
Full height for manual configuration

ButterscotchV:
Manual height config is changed to use full height, then is converted to headset height by multiplying by `0.936`. We should probably introduce a constant for that somewhere.

For HMD based height config, it now targets specifically the head tracker if available, so controllers should not interfere with measurement.

I also cleaned up a bit of the wording for the "reset all proportions" warning.


gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
```diff
@@ -123,6 +123,16 @@ export function ManualHeightStep({
...
               />
             </form>
+            <div className="flex flex-col self-center items-center justify-center">
+              <Typography>
+                {l10n.getString(
+                  'onboarding-scaled_proportions-manual_height-estimated_height'
+                )}
+              </Typography>
+              <Typography>
+                {mFormat.format(control._formValues.height * 0.936)}
```
ImUrX, line 133: you should use the [watch](https://react-hook-form.com/docs/useform/watch) function from the `useForm`

ImUrX (changes requested):

gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
ButterscotchV: ty, is this better?

ImUrX (approved):

loucass003 (approved):

gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
```diff
@@ -65,13 +66,14 @@ export function ManualHeightStep({
   );
 
   const submitHmdHeight = (values: HeightForm) => {
-    setHmdHeight(values.height);
+    const newHeight = values.height * 0.936;
```
Eirenliel, line 69: I kinda want to know where this magic multiplier came from? Either way, it should be a const value somewhere and properly named. Do not magic numbers, please!

Eirenliel (approved):

Eirenliel (changes requested):
oops wrong button

gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
ButterscotchV: Pulled it to a constant and added a comment (plus updated the existing one on the server side a little)

Eirenliel (approved):

PR #1335, opened 2025-03-08
Adjust GUI values of floor and HMD height by 0.936 to match user's real height

bappitybup:
2 line change, title self explanatory.

(not sure if the floor height change should be included, let me know)


bappitybup:
ALREADY DONE
https://github.com/SlimeVR/SlimeVR-Server/pull/1329

PR #1339, opened 2025-03-10
Scaled proportions as default

loucass003:


gui/src/components/onboarding/pages/body-proportions/ProportionsChoose.tsx
```diff
@@ -1,428 +0,0 @@
...
-                        >
-                          Description for autobone
-                        </Typography>
-                      </Localized>
-                    </div>
-                  </div>
-                  <Button
-                    variant="primary"
-                    // Check if we are in dev mode and just let it be used
-                    disabled={beneathFloor && import.meta.env.PROD}
```
ButterscotchV, line 281: We still need to consider these for the button to AutoBone on manual proportions now. It should be disabled if there is no headset present or its position is too low. There's also the check if it's a dev environment (for me to test functionality without needing a headset, though I could work around it).

ButterscotchV (changes requested):

ButterscotchV (approved):

PR #1340, opened 2025-03-10
Better Error page when connection lost + Open logs folder btn

loucass003:


ButterscotchV (approved):
Looks good, small note that the "Open the logs folder" button won't work at all on Android right now. There's not really anything we can do about that for now though.

ButterscotchV (approved):

PR #1342, opened 2025-03-10
Fix manual height not submitting form value

ButterscotchV:
It seems like we forgot to actually submit the form correctly. Changed to actually consume the form value.


gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
```diff
@@ -133,20 +142,7 @@ export function ManualHeightStep({
...
-                null,
-                null,
-                null,
-                new SkeletonHeightT(hmdHeight, 0)
-              );
-              sendRPCPacket(RpcMessage.ChangeSettingsRequest, settingsRequest);
-              nextStep();
-            }}
-          >
+          <Button variant="primary" onClick={handleSubmit(onSubmit)}>
```
loucass003, line 149: This is not the right way of doing it imo.
The form should wrap arround this button and have th onSubmit event call handle submit
Then the button can be set to type submit

loucass003: You can look at WifiCredsPage for an example

loucass003 (changes requested):

loucass003 (approved):

PR #1343, opened 2025-03-11
Fix moving average quaternions

ButterscotchV:
Synchronizes `resetQuats()` to prevent race conditions, tracks quaternion polarity in `addQuaternion()` to prevent tracker tick vs server tick race conditions, and no longer tracks the polarity of output trackers.

~~Not entirely sure why we track the polarity of the output trackers, pls let me know if there's a reason. To me, I don't see why they wouldn't just inherit the rotation of the bones, which inherit the rotation from the input trackers?~~

### DIFF
--- a/server/core/src/main/java/dev/slimevr/filtering/QuaternionMovingAverage.kt
+++ b/server/core/src/main/java/dev/slimevr/filtering/QuaternionMovingAverage.kt
@@ -68,7 +68,7 @@ class QuaternionMovingAverage(
 				val amt = predictFactor * fpsTimer.timePerFrame
 
 				// Slerps the target rotation to that predicted rotation by amt
-				filteredQuaternion = filteredQuaternion.interpR(quatBuf, amt)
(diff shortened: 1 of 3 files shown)


ButterscotchV:
I think we need to track polarity after tracker resets, that way we can ensure all trackers are in a consistent tracking space and the shortest distance between them will be shared, I am not yet sure how to handle yaw resets to retain the polarity between trackers

ButterscotchV:
SO! It turns out, there is no possible way for this code to function. Quaternions have a rotation limit of 360 degrees, and even assuming a limited joint rotation of 180 degrees, it would take a mere 2 joints to hit the limit of our precision... We cannot feasibly track rotation this way, we would be better off tracking local rotations on the skeleton and assuming positive rotations (180 degrees joint limit), then comparing between joints, reaching the maximum 360 degree limit. I don't think it's actually possible for the human body to go beyond 180 degrees for any joint, but if we wanted to, then we could go even further and use either skeleton constraints or temporal tracking within the local rotations.

tl;dr: It's not reasonably possible to track rotations globally

PR #1350, opened 2025-03-15
Unit test toe snap + fix leg tweaks & default proportions

ButterscotchV:
- Adds a toe snap unit test
  - Should fail on main and the commit introducing it, should pass on the head of this PR
- Fixes leg tweaks rotations not applying due to tracker ticks with moving average quaternions
  - Finally! Any rotation applied would actually only be applied after both a `dataTick()` and a regular `tick()`, this desync should be solved later!
- Fixes default body proportions not actually being applied when using HumanPoseManager without VRServer or proportions as input
  - Doesn't really affect much, AutoBone is not technically affected due to setting the configs on its own terms, and I don't think anything else uses this, but this is good to fix!
- Also removes this one useless math operation in leg tweaks (weird huh)

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
@@ -135,6 +135,7 @@ class HumanPoseManager(val server: VRServer?) {
 		isInternal = true,
 		isComputed = true,
 		allowFiltering = false,
+		// Do not track polarity, moving avg de-syncs ticks and breaks leg tweaks
 		trackRotDirection = false,
 	)
 
--- a/server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
@@ -124,160 +124,108 @@ class HumanPoseManager(val server: VRServer?) {
 
 	// #endregion
 	// #region private methods
+	private fun mkTrack(name: String, display: String, pos: TrackerPosition): Tracker = Tracker(
+		null,
+		getNextLocalTrackerId(),
+		name,
+		display,
+		pos,
+		hasPosition = true,
+		hasRotation = true,
+		isInternal = true,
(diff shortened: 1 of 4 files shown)


server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
```diff
@@ -110,174 +110,123 @@ class HumanPoseManager(val server: VRServer?) {
...
 		}
+		// This computes all node offsets, so the defaults don't need to be
+		// explicitly loaded into the skeleton (no need for `computeAllNodeOffsets()`)
 		skeletonConfigManager.setOffsets(offsetConfigs)
 		skeletonConfigManager.updateSettingsInSkeleton()
 	}
 
 	// #endregion
 	// #region private methods
+	private fun mkTrack(name: String, display: String, pos: TrackerPosition): Tracker = Tracker(
```
jabberrock, line 127: Maybe something like `makeComputedTracker`?

jabberrock:
> Fixes leg tweaks rotations not applying due to tracker ticks with moving average quaternions

I might be missing something, but I don't see the code that would fix this.

ButterscotchV:
> > Fixes leg tweaks rotations not applying due to tracker ticks with moving average quaternions
> 
> I might be missing something, but I don't see the code that would fix this.

I disabled moving average tracking for the HumanPoseManager trackers, this uses the direct provided raw rotation and doesn't rely on any ticking
https://github.com/SlimeVR/SlimeVR-Server/pull/1350/commits/82639a612e4c11352a1c50f1542a2dac529697b0#diff-8b83d50451df38c606da51887a5b4fa55676a1b0fc58d5ed2c5f4b2fbe9cb224R138

server/core/src/main/java/dev/slimevr/tracking/processor/HumanPoseManager.kt
ButterscotchV: 👍 Was only named mkTrack 'cause that was my quick test code, I should name it better

Eirenliel (approved):

PR #1352, opened 2025-03-17
Apply filtering post-reset, reapply 180° fix removal & fix getYawQuaternion polarity

Erimelowo:
Basically 3 things going on in this PR (1 per commit)
1. Apply filtering *after* reset instead of before (based off the reset-adjusted rotation instead of the raw rotation)
2. Revert the revert of the 180 degrees skeleton fix (so that there's no snapping but could be centaur)
3. Fix the polarity of the `getYawQuaternion(Quaternion): Quaternion` method in TrackerResetsHandler, making sure the polarity of the return Quaternion is the same as the one of the parameter Quaternion. 

This needs testing, especially for other places that use getYawQuaternion such as drift compensation and VRTrackers.
Also obviously need to make sure centaur doesn't trigger at all anymore.


ButterscotchV (approved):
Looks good but needs to be tested thoroughly, let's merge after we release v0.14.0 if possible

Eirenliel (approved):

Erimelowo:
> Looks good but needs to be tested thoroughly, let's merge after we release v0.14.0 if possible

Sure, I’m good with that

ButterscotchV:
Rad. Just to note for functionality testing, in theory this PR should extend tracking of polarity for up to 360 degrees relative distance between each tracker within the adjusted space (after full reset, mounting, etc). This only affects interpolation and angular distance within the quaternion space. I don't think there's any situation where there should be more than 360 degrees rotation between any tracker on the body, but we should make sure to test with more extreme poses and heavy use with dancing and network issues.

To develop unit tests (and for general testing), we should ensure that the relative quaternion space is retained for rotations over 360 degrees, over varying body poses, and with full/yaw/mounting resets. The most obvious sign of this failing will be the interpolated waist bone interpolating in the wrong direction (commonly called "centaur bug"). I could potentially develop unit tests branching off my code in #1349 that check the expected polarity of the resulting trackers against what we have been doing previously that we know usually works, though we would need to stay within the bounds of our assumed body restrictions for those tests, it should be able to at least confirm that our current functionality is retained at a minimum.

Pls lemme know if I got anything wrong with this, but I think we should be able to prove that this works :3

PR #1355, opened 2025-03-19
Optimise re-renders

loucass003:
Not all of it is done but this is prob a good start

a lot of work for not so much imo 
idk if it is worth it
there is still some big spots not optimised and i really feel like 50% of the problem is also react
Having a ui with everything changing based on websockets cause so much renders everywhere
unless you start deeply nesting component this kind of stuff just have to happen
but nesting more means pain to maintain

Solidjs would really help in this case....


ButterscotchV (approved):

gui/src/store/app-store.ts
```diff
@@ -0,0 +1,98 @@
...
+} from 'solarxr-protocol';
+import { selectAtom } from 'jotai/utils';
+import { isEqual } from '@react-hookz/deep-equal';
+
+export interface FlatDeviceTracker {
+  device?: DeviceDataT;
+  tracker: TrackerDataT;
+}
+
+export const ignoredTrackersAtom = atom(new Set());
```
ImUrX, line 17:
```suggestion
export const ignoredTrackersAtom = atom(new Set<string>());
```

gui/src/components/tracker/TrackerSettings.tsx
```diff
@@ -524,7 +526,10 @@ export function TrackerSettingsPage() {
                     RpcMessage.ForgetDeviceRequest,
                     new ForgetDeviceRequestT(macAddress)
                   );
-                  dispatch({ type: 'ignoreTracker', value: macAddress });
+                  setIgnoredTracker((state) => {
+                    if (!state.has(macAddress)) state.add(macAddress);
```
ImUrX, line 530: unnecesary check
```suggestion
                    state.add(macAddress);
```

gui/src/components/UnknownDeviceModal.tsx
```diff
@@ -91,9 +92,9 @@ export function UnknownDeviceModal() {
         <Button
           variant="tertiary"
           onClick={() => {
-            dispatch({
-              type: 'ignoreTracker',
-              value: currentTracker as string,
+            setIgnoredTracker((state) => {
+              if (!state.has(currentTracker)) state.add(currentTracker);
```
ImUrX, line 96: unnecesary check
```suggestion
              state.add(currentTracker);
```

gui/src/components/UnknownDeviceModal.tsx
```diff
@@ -28,7 +29,7 @@ export function UnknownDeviceModal() {
         ['/onboarding/connect-trackers', '/settings/firmware-tool'].includes(
           pathname
         ) ||
-        state.ignoredTrackers.has(macAddress as string) ||
+        ignoredTrackers.has(macAddress as string) ||
```
ImUrX, line 32: not needed if ignoredTrackers is `Set<string>`
```suggestion
        ignoredTrackers.has(macAddress) ||
```

ImUrX (changes requested):
good stuff

PR #1359, opened 2025-03-29
Interpolation tests & better skeleton test class

ButterscotchV:
Skeleton intepolation handling is a bit of a mess right now, this PR adds some extra unit tests and a new function for joint interpolation. No longer interferes with #1352.

Branched off from #1349 but without any of the actual functionality changes. This should be safe to merge at any time for additional testing and a cleaner unit test setup without breaking anything! :3

Also just a small note, we will likely remove the joint interpolation stuff entirely, but the function added here can be used to verify functionality in that regard, especially to ensure that we don't have functionality regression.


Eirenliel (approved):

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -1736,11 +1736,48 @@ class HumanSkeleton(
...
+			val isLower = (ref * UP_ADJ).angleToR(extended) > FastMath.HALF_PI
+
+			return if (isBack && isLower) {
+				// Select longest rotation
+				if (extended.dot(ref) >= 0f) -extended else extended
+			} else {
+				// Select shortest rotation
+				extended.twinNearest(ref)
+			}
+		}
```
ImUrX, line 1781: maybe we should move this to `Quaternion`?

server/core/src/test/java/dev/slimevr/unit/TestTrackerSet.kt
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * All the trackers in the set as a list.
+	 */
+	val setL = set.asList()
+
+	/**
+	 * All the trackers in the set plus the headset.
+	 */
+	val all = set.plus(head)
```
ImUrX, line 44:
```suggestion
	val all = set + head
```

server/core/src/test/java/dev/slimevr/unit/TrackerTestUtils.kt
```diff
@@ -58,4 +59,13 @@ object TrackerUtils {
...
 			FastMath.isApproxEqual(q1.z, q2.z, tolerance)
+
+	fun randFloat(random: Random, minInclusive: Float, maxExclusive: Float): Float = minInclusive + (random.nextFloat() * (maxExclusive - minInclusive))
+
+	fun randQuat(random: Random): Quaternion = Quaternion(
+		randFloat(random, -1f, 1f),
+		randFloat(random, -1f, 1f),
+		randFloat(random, -1f, 1f),
+		randFloat(random, -1f, 1f),
+	).unit()
```
ImUrX, line 70: just a comment but QuaternionTest has a guassian distribution-based random if interested

ImUrX (approved):

server/core/src/test/java/dev/slimevr/unit/TrackerTestUtils.kt
ButterscotchV: I did see that, I figured this would be ideal to keep a seed per test rather than global? I'm not sure what is the best option

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
ButterscotchV: Hmm perhaps... This is very skeleton specific, but also that would make more logical sense, I think it's a good idea

ButterscotchV: I think this is good yeah? I also realized I can just use `twinNearest` directly lol this code is so much simpler

server/core/src/test/java/dev/slimevr/unit/TrackerTestUtils.kt
Eirenliel: The best option is not to use randoms in tests, but to cover fixed edge cases.

Eirenliel (approved):

server/core/src/test/java/dev/slimevr/unit/TrackerTestUtils.kt
ButterscotchV: Not randomized anymore :3

Eirenliel (approved):

PR #1364, opened 2025-04-03
Stop using Twemoji and just use SVGs for it

ImUrX:
I replaced the ~2MB of emoji fonts that we had with normal SVG imports which reduced the total size of the binary and we don't have to make any special hack for fonts in some OSes.


ButterscotchV (approved):

Eirenliel (commented):
But does it work in other places? Like if we put emoji in a changelog in firmware?

ImUrX:
No it doesn't work with other places as it's all manually replaced, the reason why I don't like the current way of doing it is that it's kind of a hack that works in different ways depending on the browser and doesn't even work in Linux.

We could use a library that replaces all emojis with an svg automatically but it might be heavier in size than the two color fonts combined, but I'm not sure

Eirenliel (approved):
Aight, no emojis for changelogs i guess

ImUrX:
Emojis will still work, it will just use the os default emoji font

PR #1369, opened 2025-04-07
Clean-up PoseStreamer and PoseRecorder

ButterscotchV:
Converts all the PoseStreamer classes to Kotlin, then cleans up functionality, making a generic TickReducer class instead of implementing ticking twice manually. This also decouples the ticking from system time, so it should be much nicer to unit test now.

Warning: Not yet tested


server/core/src/main/java/dev/slimevr/posestreamer/ServerPoseStreamer.kt
```diff
@@ -0,0 +1,27 @@
...
+import dev.slimevr.tracking.processor.skeleton.HumanSkeleton
+import dev.slimevr.util.ann.VRServerThread
+import java.util.function.Consumer
+
+class ServerPoseStreamer(val server: VRServer) : TickPoseStreamer(server.humanPoseManager.skeleton) {
+
+	init {
+		// Register callbacks/events
+		server.addSkeletonUpdatedCallback(Consumer { skeleton: HumanSkeleton? -> this.onSkeletonUpdated(skeleton) })
+		server.addOnTick(Runnable { this.tick() })
```
ImUrX: cant you just do lambda syntax in here directly?

ButterscotchV: Yes :3

ButterscotchV:
Tested lightly and it works really well, got around 48-50 TPS easily from the main tick function

PR #1382, opened 2025-04-14
Simplify mounting reset status checking

ButterscotchV:
Currently we're over-detecting on mounting reset status checking, hopefully only checking trackers that will actually be reset can reduce the false positive? Also let's just check the tracker directly, as we really shouldn't rely on the status system in that way.

This whole thing has turned into quite a mess, maybe the status system isn't a good idea? We could probably just replace it with direct SolarXR packets that evaluate the current state rather than trying to track the state separately.


loucass003:
Status system code is getting removed by the flightlist sytem. something a bit more isolated that just check for global state of the trackers and server insead of code spread everywhere in the server

PR #1383, opened 2025-04-14
Scale arms on proportions reset

ButterscotchV:
Sort of a bug? Arms aren't being scaled with the rest of the proportions, so we get unusually short arms for tall people and unusually long arms for short people lol. Here's a simple fix that just sets up scaling based on the default proportions.


ButterscotchV:
Do we also need to scale shoulders in any way? Or hands? I put the arms for now 'cause I think that's most reasonable based on what we've been doing so far

ImUrX (approved):

loucass003 (approved):

PR #1389, opened 2025-04-15
Make Firmware update more reliable, prevent deadlocks of ui

loucass003:
Fixes: https://github.com/SlimeVR/SlimeVR-Server/issues/1377

### DIFF
--- a/gui/src/components/firmware-update/FirmwareUpdate.tsx
+++ b/gui/src/components/firmware-update/FirmwareUpdate.tsx
@@ -286,11 +286,10 @@ export function FirmwareUpdate() {
   };
 
   const startUpdate = () => {
-    const selectedDevices = Object.keys(selectedDevicesForm)
-      .filter((d) => selectedDevicesForm[d])
-      .reduce((curr, id) => {
+    const selectedDevices = Object.keys(selectedDevicesForm).reduce(
+      (curr, id) => {
+        if (!selectedDevicesForm[id]) return curr;
         const device = devices.find(({ id: dId }) => id === dId?.id.toString());
-
         if (!device) return curr;
         return [
           ...curr,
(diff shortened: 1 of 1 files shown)


gui/src/components/firmware-update/FirmwareUpdate.tsx
```diff
@@ -280,25 +281,29 @@ export function FirmwareUpdate() {
...
+        if (!device) return curr;
+        return [
+          ...curr,
+          {
+            type: FirmwareUpdateMethod.OTAFirmwareUpdate,
+            deviceId: id,
+            deviceNames: deviceNames(device, l10n),
+          },
+        ];
+      }, [] as SelectedDevice[]);
```
ImUrX, line 303: the `if(!device)` check should be on filter function and this should be a `map`

ImUrX (changes requested):

gui/src/components/firmware-update/FirmwareUpdate.tsx
ImUrX: edited the review

loucass003: Why reduce is the perfect case to perform both filter and map at the same time. it is literally made for it

loucass003: wait youre right tho ... this is cleaner

loucass003: i need device both for the filter and the content of the map. so no reduce is the best case

ImUrX (approved):

gui/src/components/firmware-update/FirmwareUpdate.tsx
```diff
@@ -280,25 +281,29 @@ export function FirmwareUpdate() {
...
       ),
     });
-    queueFlashing(devices);
+    setStatus({});
+    clear();
   };
 
   const startUpdate = () => {
     const selectedDevices = Object.keys(selectedDevicesForm)
       .filter((d) => selectedDevicesForm[d])
```
ImUrX, line 290: can you add this check inside the reduce at least, it will reduce needed iterations

PR #1395, opened 2025-04-17
Load configured height for manual height config

ButterscotchV:
Currently we only load the configured height for automatic height config, we should do the same for manual.

This PR is still a draft because it doesn't seem to actually apply the default. This is probably because of retaining the initial state before the config is loaded.

### DIFF
--- a/gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
+++ b/gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
@@ -46,10 +46,10 @@ export function ManualHeightStep({
 
   // Load the last configured height
   useEffect(() => {
-    const height = currentHeight();
     reset({
       height:
-        (currentHeight && currentHeight / EYE_HEIGHT_TO_HEIGHT_RATIO) || DEFAULT_FULL_HEIGHT,
+        (currentHeight && currentHeight / EYE_HEIGHT_TO_HEIGHT_RATIO) ||
+        DEFAULT_FULL_HEIGHT,
     });
   }, [currentHeight]);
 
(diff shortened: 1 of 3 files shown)


gui/src/hooks/height.ts
```diff
@@ -44,7 +30,9 @@ export function useProvideHeightContext(): HeightContext {
     }
   });
 
-  return { hmdHeight, setHmdHeight, floorHeight, setFloorHeight, validateHeight };
+  const currentHeight = () => computeHeight(hmdHeight, floorHeight);
```
loucass003, line 33:
```suggestion
  const currentHeight = useMemo(() => computeHeight(hmdHeight, floorHeight), [hmdHeight, floorHeight])
```

gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
```diff
@@ -31,14 +35,24 @@ export function ManualHeightStep({
...
   const { sendRPCPacket } = useWebsocketAPI();
   const { currentLocales } = useLocaleConfig();
   const height = watch('height');
 
+  // Load the last configured height
+  useEffect(() => {
+    const height = currentHeight();
+    reset({
+      height:
+        (height && height / EYE_HEIGHT_TO_HEIGHT_RATIO) || DEFAULT_FULL_HEIGHT,
```
loucass003, line 52:
```suggestion
        (currentHeight && currentHeight / EYE_HEIGHT_TO_HEIGHT_RATIO) || DEFAULT_FULL_HEIGHT,
```

gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
```diff
@@ -31,14 +35,24 @@ export function ManualHeightStep({
...
+    });
+  }, [currentHeight()]);
```
loucass003, line 54:
```suggestion
  }, [currentHeight]);
```

gui/src/hooks/height.ts
```diff
@@ -8,28 +8,14 @@ export interface HeightContext {
   setHmdHeight: React.Dispatch<React.SetStateAction<number | null>>;
   floorHeight: number | null;
   setFloorHeight: React.Dispatch<React.SetStateAction<number | null>>;
-  validateHeight: (
-    hmdHeight: number | null | undefined,
-    floorHeight: number | null | undefined
-  ) => boolean;
+  currentHeight: () => number | null;
```
loucass003, line 11:
```suggestion
  currentHeight: number | null;
```

loucass003 (approved):

PR #1401, opened 2025-04-21
differentiable reset sounds

sctanf:
original sounds between full reset and mounting could be differentiated by speed, so instead play a unique pattern
I am not sure this is needed, but maybe it is a good idea.


sctanf:
needs feedback on the reset start tones

ImUrX:
if using `F#` I would make the reset sound ends end in `G4`, will make it feel like it ended correctly

sctanf:
that sounds good and you can commit it; since reset sounds might be replaced again anyway i dont know if this should be worked on

ImUrX:
I mean until it gets changed why not

Polymoria (commented):
<3

Polymoria (approved):
<33

PR #1407, opened 2025-04-25
Session flightlist

loucass003:


gui/public/i18n/en/translation.ftl
```diff
@@ -1132,7 +1147,11 @@ onboarding-automatic_mounting-done-description = Your mounting calibration is co
 onboarding-automatic_mounting-done-restart = Try again
 onboarding-automatic_mounting-mounting_reset-title = Mounting Reset
 onboarding-automatic_mounting-mounting_reset-step-0 = 1. Squat in a "skiing" pose with your legs bent, your upper body tilted forwards, and your arms bent.
-onboarding-automatic_mounting-mounting_reset-step-1 = 2. Press the "Reset Mounting" button and wait for 3 seconds before the trackers' mounting orientations will reset.
+onboarding-automatic_mounting-mounting_reset-step-1 = 2. Press the "Mounting calibration" button and wait for 3 seconds before the trackers' mounting orientations will reset.
+
+onboarding-automatic_mounting-mounting_reset-feet-step-0 = 1. Stand on your toes, both feets pointing forward. Alternatively you can do it siting on a chair.
```
ButterscotchV:
```suggestion
onboarding-automatic_mounting-mounting_reset-feet-step-0 = 1. Stand on your toes with both feet pointing forward. Alternatively you can do it siting on a chair.
```

gui/public/i18n/en/translation.ftl
```diff
@@ -1522,3 +1543,58 @@ error_collection_modal-description_v2 = { settings-interface-behavior-error_trac
...
+tracking_checklist-status-complete = You are prepared to use SlimeVR!
+tracking_checklist-MOUNTING_CALIBRATION = Perform a mounting calibration
+tracking_checklist-FEET_MOUNTING_CALIBRATION = Perform a feet mounting calibration
+tracking_checklist-FULL_RESET = Perform a full Reset
+tracking_checklist-FULL_RESET-desc = Some Trackers need a reset to be performed
+tracking_checklist-STEAMVR_DISCONNECTED = SteamVR not running
+tracking_checklist-STEAMVR_DISCONNECTED-desc = SteamVR is not running. Are you using it for vr?
+tracking_checklist-STEAMVR_DISCONNECTED-open = Launch SteamVR
+tracking_checklist-TRACKERS_REST_CALIBRATION = Calibrate your trackers
+tracking_checklist-TRACKERS_REST_CALIBRATION-desc = You didnt perform the tracker calibration. Please let your slimes, highlited in yellow, rest on a static surface for a few secconds
```
ButterscotchV:
```suggestion
tracking_checklist-TRACKERS_REST_CALIBRATION-desc = You didn't perform tracker calibration. Please let your Slimes (highlighted in yellow) rest on a stable surface for a few secconds.
```

gui/public/i18n/en/translation.ftl
```diff
@@ -1522,3 +1543,58 @@ error_collection_modal-description_v2 = { settings-interface-behavior-error_trac
...
+          This is not recommended for SlimeVR to function properly.
+          <PublicFixLink>See how to fix it here.</PublicFixLink>
+    *[many] Some of your network adapters are set to public:
+            {$adapters}
+            This is not recommended for SlimeVR to function properly.
+            <PublicFixLink>See how to fix it here.</PublicFixLink>
+}
+tracking_checklist-NETWORK_PROFILE_PUBLIC-open = Open Control Panel
+tracking_checklist-STAY_ALIGNED_CONFIGURED = Configure Stay Aligned
+tracking_checklist-STAY_ALIGNED_CONFIGURED-desc = Record the stay aligned poses for an improved imu drift
```
ButterscotchV:
```suggestion
tracking_checklist-STAY_ALIGNED_CONFIGURED-desc = Record the Stay Aligned poses for an improved imu drift
```

server/core/src/main/java/dev/slimevr/config/TapDetectionConfig.kt
```diff
@@ -14,20 +14,16 @@ class TapDetectionConfig {
...
-			field = yawResetTaps
 		}
 	var fullResetTaps = 3
 		set(fullResetTaps) {
 			field = FastMath.clamp(fullResetTaps.toFloat(), 2f, 10f).toInt()
-			field = fullResetTaps
 		}
 	var mountingResetTaps = 3
 		set(mountingResetTaps) {
 			field = FastMath.clamp(mountingResetTaps.toFloat(), 2f, 10f).toInt()
```
ButterscotchV: We can also just use `coerceIn` for these to not convert to floats and stuff
See: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.ranges/coerce-in.html
(Might need an import)

```suggestion
			field = yawResetTaps.coerceIn(2, 10)
		}
	var fullResetTaps = 3
		set(fullResetTaps) {
			field = fullResetTaps.coerceIn(2, 10)
		}
	var mountingResetTaps = 3
		set(mountingResetTaps) {
			field = mountingResetTaps.coerceIn(2, 10)
```

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
ButterscotchV, line 221: These should all really be `val`, not `var`.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
```diff
@@ -1619,45 +1616,70 @@ class HumanSkeleton(
...
 			}
 		}
 		legTweaks.resetBuffer()
 		localizer.reset()
+
+		if (humanPoseManager.server != null) {
+			humanPoseManager.server.configManager.vrConfig.resetsConfig.lastMountingMethod =
+				MountingMethods.AUTOMATIC
+			if (!humanPoseManager.server.trackingChecklistManager.resetMountingCompleted) {
+				humanPoseManager.server.trackingChecklistManager.resetMountingCompleted = bodyParts.any { it ->
```
ButterscotchV, line 1640: Is `any` the right choice here? We're essentially just checking if one of the body parts is matching, not all.

ButterscotchV (changes requested):

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
loucass003: it is a bit broader bc imo its a bigger deal if the status don't clear and get stuck bc of something wierd. bc the list on the sever could mismatch the one in the GUI. I hate the current state of things .... we should have not kept that

ButterscotchV: Fair, definitely not ideal. If this is what needs to be done then it can be resolved.

server/core/src/main/java/dev/slimevr/tracking/processor/skeleton/HumanSkeleton.kt
loucass003: it does not want it

Issue #1112, opened 2024-07-16
Battery percentage should not be shown when charging


Vyolex:
I've seen a lot of new users get confused by the UI showing 100% charge while the tracker is being charged, thinking it is fully charged regardless of what the actual battery percentage is at.

I feel like the UI should just show a charging indicator and no percentages (that are meaningless and incorrect anyway) as it would be less confusing for new users. Would be nice if later hardware revisions can have current sensing to more accurately display the charge level when charging.

Related to #669 for more clear/better battery reporting.

kruemmelbande:
For the most part the tracker doesnt know if its charging. It literally cant tell the difference between being at 100% and being on the charger

Vyolex:
Correct, still the server displays a charging indicator which is at >4.3V atm I believe. 

There are cases where it gets mistakenly displayed as charging while it is just fully charged (I have two official dev trackes that do so). I still believe it will be less confusing to remove the percentage compared to what is happening now. But maybe we need to fine-tune the charging threshold value too?

Minimewatson99:
my diy slimes say 84% when charging anyways, so getting a consistent result that works across all trackers, is somewhat impossible anyways. 
An improvement could be made for if the tracker is above a certain voltage, but it wouldn't work across all slimes. the usual voltage of a full battery is around 4.2v, and my official slimes read 4.5 when charging, so you could potentially hide battery percentage when the tracker is above 4.3v, which would probably be what you are after. 

Also, because the way the battery percentage is worked out, the percentage of the tracker while charging is impossible to get, as the tracker simply reports the voltage of the battery, and if it is above the maximum voltage that the battery should give, it reports it as charging, so unless estimating the charging time of a tracker, the percentage while charging is not possible to get with the current design of trackers

kruemmelbande:
i literally get the exact same value when my tracker is empty but charging and fully charged. The hardware simply doesnt allow you to tell reliably if you are charging or not. If you lower the threashold, itll just start reporting as charging when its not, or itll be inconsistent.

Minimewatson99:
> Correct, still the server displays a charging indicator which is at >4.3V atm I believe.
> 
> There are cases where it gets mistakenly displayed as charging while it is just fully charged (I have two official dev trackes that do so). I still believe it will be less confusing to remove the percentage compared to what is happening now. But maybe we need to fine-tune the charging threshold value too?

[...]

PR #1410, opened 2025-04-25
Better battery and wifi statuses behavior

### LINKED ISSUE #669: Dead battery doesn't show up
The tracker with low battery (blinking constantly but still connected) doesn't show battery indicator in the server GUI. If put on charge it will show 100% and charging, but when unplugged the battery icon will just disappear along with percentage and voltage. Not sure if it's GUI, Server or firmware problem...

sctanf:
Battery will continue to show even if the level is 0%, and the level will be hidden if in a charge state. Ping will be hidden if it is null. Firmware updater will also treat 0% level as low battery.

Fix #669 and #1112

### DIFF
--- a/gui/src/components/tracker/TrackerBattery.tsx
+++ b/gui/src/components/tracker/TrackerBattery.tsx
@@ -31,38 +31,21 @@ export function TrackerBattery({
     <div className="flex gap-2">
       <div className="flex flex-col justify-around">
         <BatteryIcon
-          value={((voltage || 0) <= 4.3 && value) || 0}
+          value={value}
           disabled={disabled}
-          charging={false}
+          charging={(voltage || 0) > 4.3}
         />
       </div>
-      {((voltage || 0) > 4.3 && (
-        <div className="flex flex-col justify-center w-10">
-          <svg
-            width="18"
-            height="16"
-            viewBox="0 0 14 10"
-            xmlns="http://www.w3.org/2000/svg"
-          >
-            <path
-              className={'fill-status-warning'}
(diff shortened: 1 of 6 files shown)


sctanf:
Fixes null rssi and white wifi icon when rssi value causes "percentage" of 0 and incorrect "y" calculation

gui/src/components/tracker/TrackerBattery.tsx
```diff
@@ -36,16 +36,24 @@ export function TrackerBattery({
...
-        <Typography color={textColor}>
-          {percentFormatter.format(value)}
-        </Typography>
-        {voltage && config?.debug && (
-          <Typography color={textColor}>
-            {voltageFormatter.format(voltage)}V
-          </Typography>
-        )}
-      </div>
+      {(((voltage || 0) <= 4.3 || (voltage && config?.debug)) && (
```
loucass003, line 39: can you move this into a variable?
so it is a bit cleaner 🙏

gui/src/components/tracker/TrackersTable.tsx
```diff
@@ -290,9 +290,9 @@ export function TrackersTable({
           (device?.hardwareStatus?.rssi != null ||
             device?.hardwareStatus?.ping != null) && (
```
loucass003, line 291: if this check is here.
None of the ping !=null and rssi != null should matter?

loucass003: my bad cant read. But i dont think this condition is needed anymore

sctanf: The wifi section is disabled if neither rssi nor ping exist, which covers the case when a tracker has just connected but sent no statuses. This is the same as battery, and is also the original behavior, so I did not need to change it

PR #1418, opened 2025-04-29
Stay Aligned

jabberrock:
Keeps your trackers aligned by slowly adjusting the yaw of your trackers to correct for drift.

Every tick, Stay Aligned adjusts the yaw of one tracker. It looks at the player's current pose, and adjusts the yaw of the tracker by a tiny amount, in the direction which is closest to the relaxed pose that the player set up. Over time, this has the effect of keeping all the trackers in a human-like pose.

1. When a tracker is at rest, we "lock" the tracker and try to maintain its yaw. This is very useful when sitting or lying down because the player is rarely moving. They shouldn't drift at all.
2. When a tracker is moving, we try to match the tracker to the player's current pose. For example, let's say the player is standing and talking to friends. On average, we expect the player to return to their relaxed standing pose. So, we try to rotate the trackers towards the relaxed standing pose. However, we also try to balance a tracker between its above and below trackers, so that the skeleton remains realistic when, for example, the player has their legs in a wide stance.

Main flow
1. `HumanSkeleton` calls `StayAligned.adjustNextTracker` every server tick
3. `StayAligned.adjustNextTracker` picks the next tracker, and calls `AdjustTrackerYaw.adjust`
4. `AdjustTrackerYaw.adjust` calculates the error between the tracker's rotation, and its ideal rotation, then uses gradient descent to adjust a yaw correction to reduce the error
5. Calculating the error is handled by one of `LockedErrorVisitor`, `CenterErrorVisitor` or `NeighborErrorVisitor`
6. `Tracker.getRotation/getIdentityAdjustedRotation` applies the yaw correction to the raw rotation, before applying all the other full and mounting adjustments. This is because drift happens in the tracker's raw rotation space.

Home
image

Stay Aligned Settings
image
image

Setup Stay Aligned
image
image
image
image

SolarXR-Protocol: https://github.com/SlimeVR/SolarXR-Protocol/pull/170


Sacred-Shadow:
Relaxed Lying?

I showed the picture to three other people, and all of them also interpreted it as a sitting position.
The four of us associate lying with something different.

jabberrock:
> Relaxed Lying?
> 
> I showed the picture to three other people, and all of them also interpreted it as a sitting position. The four of us associate lying with something different.

You're right, the text was from an earlier version and I forgot to update it. Should be "Sitting on floor" like on the settings page.

Sacred-Shadow:
One more thing – I talked about the feature with a few friends and recommended it. What I often heard was that the most important positions – lying on the back or on the side – are missing.

Many mentioned that drift tends to happen more often in those positions.

Is that something that might be added later on?

jabberrock:
> One more thing – I talked about the feature with a few friends and recommended it. What I often heard was that the most important positions – lying on the back or on the side – are missing.
> 
> Many mentioned that drift tends to happen more often in those positions.
> 
> Is that something that might be added later on?

1. Sitting on ground and lying on back are handled by the same pose. I'll update the text.
2. Stay Aligned can't do much for lying on side, beyond locking non-moving trackers. You can curl up to any angle, so there's no "expected" pose that Stay Aligned can adjust the trackers.

PR #1419, opened 2025-04-29
TrackersHID fix exception from null magStatus

sctanf:


jabberrock:
Makes sense, but why would we be receiving invalid `mag_id` from the tracker? Is there a bug somewhere in the parsing code?

sctanf:
> Makes sense, but why would we be receiving invalid `mag_id` from the tracker? Is there a bug somewhere in the parsing code?

technically the tracker can send whatever it wants to the server, and server just has to deal with it
this interface is temporary anyway, since trackers should be provisioned by serial not hid

jabberrock:
> > Makes sense, but why would we be receiving invalid `mag_id` from the tracker? Is there a bug somewhere in the parsing code?
> 
> technically the tracker can send whatever it wants to the server, and server just has to deal with it this interface is temporary anyway, since trackers should be provisioned by serial not hid

Agreed that server should be defensive. But there's probably a hidden bug somewhere. Either tracker firmware is sending bad data, or there's a bug with the parsing.

jabberrock (approved):

sctanf:
> > > Makes sense, but why would we be receiving invalid `mag_id` from the tracker? Is there a bug somewhere in the parsing code?
> > 
> > 
> > technically the tracker can send whatever it wants to the server, and server just has to deal with it this interface is temporary anyway, since trackers should be provisioned by serial not hid
> 
> Agreed that server should be defensive. But there's probably a hidden bug somewhere. Either tracker firmware is sending bad data, or there's a bug with the parsing.

also nope, i assume it is just corruption from packet send because i do not currently time any radio transmission, and the small packet size makes it easy to cause errors

ButterscotchV (approved):

loucass003 (approved):

PR #1439, opened 2025-05-20
[Stay Aligned] Move reset stay aligned after filtering

jabberrock:
Stay Aligned correction should only be reset after filtering, so that filtering can make use of the corrected rotation.


ButterscotchV (approved):
Probably doesn't matter, but I figure Stay Aligned probably wants the most up to date and accurate rotation, so we should probably reset the moving average quaternion stuff first. Looks good :3

jabberrock:
@ButterscotchV or @Eirenliel feel free to merge this. I don't have permissions! Thanks

PR #1440, opened 2025-05-21
Add a proper tracker name for gestures trackers

9il:


ImUrX:
Should probably use the manufacturer hardware info for saying `Gestures`

sctanf:
manufacturer is hard coded for all device depending on how it connects, since it is actually referred as type field in gui

Eirenliel (approved):

PR #1441, opened 2025-05-21
Allow partial/separate resets for feet and fingers

Erimelowo:
This adds an array on every reset used to specify which trackers should be reset. 
If the array is empty, the server will decide which trackers will be reset (all except fingers and feet during mounting basically).
This deprecates the feet reset feature since I believe it is much better to keep it a separate mounting reset.
This is also backwards compatible.

Needs solar (need edit)

### DIFF
--- a/gui/src/components/home/ResetButton.tsx
+++ b/gui/src/components/home/ResetButton.tsx
@@ -1,6 +1,7 @@
 import { useLocalization } from '@fluent/react';
 import { useEffect, useMemo, useRef, useState } from 'react';
 import {
+  BodyPart,
   ResetRequestT,
   ResetType,
   RpcMessage,
@@ -26,12 +27,14 @@ import classNames from 'classnames';
 export function ResetButton({
   type,
   size = 'big',
+  bodyPartsToReset = 'default',
   className,
   onReseted,
 }: {
   className?: string;
   type: ResetType;
   size: 'big' | 'small';
+  bodyPartsToReset?: 'default' | 'feet' | 'fingers';
   onReseted?: () => void;
 }) {
   const { l10n } = useLocalization();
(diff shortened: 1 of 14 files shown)


loucass003 (approved):

gui/src/components/home/ResetButton.tsx
```diff
@@ -50,9 +53,53 @@ export function ResetButton({
...
   const reset = () => {
     const req = new ResetRequestT();
     req.resetType = type;
+    if (bodyPartsToReset == 'default') {
+      // Default (server handles it)
+      req.bodyParts = [];
+    } else if (bodyPartsToReset == 'feet') {
+      // Feet
+      req.bodyParts = feetBodyParts;
+    } else if (bodyPartsToReset == 'fingers') {
```
ImUrX, line 99:
```suggestion
    if (bodyPartsToReset === 'default') {
      // Default (server handles it)
      req.bodyParts = [];
    } else if (bodyPartsToReset === 'feet') {
      // Feet
      req.bodyParts = feetBodyParts;
    } else if (bodyPartsToReset === 'fingers') {
```

gui/src/components/home/ResetButton.tsx
```diff
@@ -75,13 +122,22 @@ export function ResetButton({
   const text = useMemo(() => {
     switch (type) {
       case ResetType.Yaw:
-        return l10n.getString('reset-yaw');
+        return l10n.getString(
+          'reset-yaw' +
+            (bodyPartsToReset != 'default' ? '-' + bodyPartsToReset : '')
```
ImUrX, line 127:
```suggestion
            (bodyPartsToReset !== 'default' ? '-' + bodyPartsToReset : '')
```

gui/src/components/home/ResetButton.tsx
```diff
@@ -75,13 +122,22 @@ export function ResetButton({
...
+        );
       case ResetType.Mounting:
-        return l10n.getString('reset-mounting');
+        return l10n.getString(
+          'reset-mounting' +
+            (bodyPartsToReset != 'default' ? '-' + bodyPartsToReset : '')
```
ImUrX, line 132:
```suggestion
            (bodyPartsToReset !== 'default' ? '-' + bodyPartsToReset : '')
```

gui/src/components/home/ResetButton.tsx
```diff
@@ -75,13 +122,22 @@ export function ResetButton({
...
+        );
       case ResetType.Full:
-        return l10n.getString('reset-full');
+        return l10n.getString(
+          'reset-full' +
+            (bodyPartsToReset != 'default' ? '-' + bodyPartsToReset : '')
```
ImUrX, line 137:
```suggestion
            (bodyPartsToReset !== 'default' ? '-' + bodyPartsToReset : '')
```

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerPosition.kt
```diff
@@ -135,5 +135,99 @@ enum class TrackerPosition(
...
+					it == TrackerPosition.RIGHT_MIDDLE_DISTAL ||
+					it == TrackerPosition.RIGHT_RING_PROXIMAL ||
+					it == TrackerPosition.RIGHT_RING_INTERMEDIATE ||
+					it == TrackerPosition.RIGHT_RING_DISTAL ||
+					it == TrackerPosition.RIGHT_LITTLE_PROXIMAL ||
+					it == TrackerPosition.RIGHT_LITTLE_INTERMEDIATE ||
+					it == TrackerPosition.RIGHT_LITTLE_DISTAL
+			}
+			return false
+		}
```
ImUrX, line 231: maybe this should be a member function on the enum?
ofc it wont be usable with null object but it will be much easier to read
```kotlin
TrackerPosition.isLeftArm(tracker.trackerPosition)
// or
tracker.trackerPosition != null && 
(tracker.trackerPosition.isLeftArm() || tracker.trackerPosition.isLeftFinger()) 
```

or if you dont want to check for null, just make it an extension function
```kotlin
fun TrackerPosition?.isLeftArm(): boolean {
  this?.let {
    return true
  }
  return false
}

tracker.trackerPosition.isLeftArm()
```

ImUrX (changes requested):

server/core/src/main/java/dev/slimevr/tracking/trackers/TrackerPosition.kt
Erimelowo: Done with extension functions in https://github.com/SlimeVR/SlimeVR-Server/pull/1441/commits/2a6870e5a91462a0124276e9026f33eec97a507a

Erimelowo:
needs solarxr

PR #1444, opened 2025-05-24
TrackersHID only reopen if no data received

sctanf:


ButterscotchV (approved):
I don't know much on the HID side but the code looks fine

ImUrX (approved):

sctanf:
testing seems to show the issues caused before are fixed

PR #1447, opened 2025-05-28
stay aligned fixes

loucass003:


jabberrock (approved):
Thanks for addressing all my comments so quickly! Tested in VR with HMD and it looks good.

PR #1449, opened 2025-05-30
Manual proportions visual fixes

Polymoria:
Changes:

- Added an info icon that you can hover over or click on mobile to display an info tooltip. Explanation: Now you won't get a tooltip if your mouse hovered on the space in the middle of the block, which looked a bit random, especially without a reaction of the body part name to the hover. Now you see the tooltip only if you need info about that body part.
- Opacity of the incrementation buttons was changed: you don't need to have them that transparent when they are out of hover, it messes with the contrast, and you need to hover on each of them if you want to read them comfortably. Now you don't need to hover them for that, but opacity changes on hover anyway to show subtly that blocks are hovered :3
- I'm changing the scale of the "more proportion" down arrow when you hover, I think it's cool, instead of when nothing happens.

Also, I didn't add an SVG for that icon because I don't think I need SVG for an I and border around it :D

### DIFF
--- a/gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
+++ b/gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
@@ -128,28 +128,32 @@ function ProportionItem({
             'duration-300'
           )}
         >
-          <Tooltip
-            content={
-              <Localized id={`${part.label}-desc`}>
-                <Typography
-                  variant="standard"
-                  whitespace="whitespace-pre-wrap"
-                ></Typography>
-              </Localized>
-            }
-            preferedDirection="bottom"
-            mode="corner"
-          >
-            <div className="flex flex-grow" onClick={toggleOpen}>
-              <Typography variant="section-title" bold>
-                {l10n.getString(part.label)}
-              </Typography>
-            </div>
-          </Tooltip>
+          <div className="flex flex-grow" onClick={toggleOpen}>
+            <Typography variant="section-title" bold>
+              {l10n.getString(part.label)}
+            </Typography>
+            <Tooltip
+              content={
+                <Localized id={`${part.label}-desc`}>
+                  <Typography
+                    variant="standard"
+                    whitespace="whitespace-pre-wrap"
+                  ></Typography>
+                </Localized>
+              }
+              preferedDirection="bottom"
+              mode="corner"
+            >
+              <div className="info-icon hover:opacity-100 opacity-65 ml-1 scale-[0.65] border-2 border-solid text-xs w-5 h-5 flex justify-center items-center rounded-full">
+                i
+              </div>
+            </Tooltip>
+          </div>
+
           <div className="flex gap-4 items-center mobile:justify-center mobile:w-full">
             <div
               className={classNames(
-                'flex items-center gap-2 my-2 opacity-10',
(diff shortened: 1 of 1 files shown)


Polymoria:
image

Polymoria:
@loucass003 @ImUrX Should I fix something here?

gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
```diff
@@ -128,28 +128,32 @@ function ProportionItem({
...
+                  <Typography
+                    variant="standard"
+                    whitespace="whitespace-pre-wrap"
+                  ></Typography>
+                </Localized>
+              }
+              preferedDirection="bottom"
+              mode="corner"
+            >
+              <div className="info-icon hover:opacity-100 opacity-65 ml-1 scale-[0.65] border-2 border-solid text-xs w-5 h-5 flex justify-center items-center rounded-full">
```
ImUrX:
```suggestion
              <div className="hover:opacity-100 opacity-65 ml-1 scale-[0.65] border-2 border-solid text-xs w-5 h-5 flex justify-center items-center rounded-full">
```
that class doesnt exist

gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
```diff
@@ -128,28 +128,32 @@ function ProportionItem({
...
+                i
+              </div>
+            </Tooltip>
+          </div>
+
           <div className="flex gap-4 items-center mobile:justify-center mobile:w-full">
             <div
               className={classNames(
-                'flex items-center gap-2 my-2 opacity-10',
+                'flex items-center gap-2 my-2 opacity-75',
```
ImUrX: i honestly think opacity on these should be always 100

gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
```diff
@@ -182,7 +186,7 @@ function ProportionItem({
             </div>
             <div
               className={classNames(
-                'flex items-center gap-2 my-2 opacity-10',
+                'flex items-center gap-2 my-2 opacity-75',
```
ImUrX: opacity should probably be 100

ImUrX (changes requested):
I think the `IncrementButton` should have `bg-background-60` and when its inside the ratio dropdown it should be `bg-background-40`

gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
Polymoria: <s>why their opacity is separate xD</s>

Polymoria: Alright, I don't like how contrast between bg of the button and number looks with 100 opacity xD going to try to fix, need to think a bit 
image

Polymoria: that is not a <p> semantically 
image

gui/src/components/onboarding/pages/body-proportions/BodyProportions.tsx
Polymoria: image

ImUrX (approved):

PR #1450, opened 2025-05-30
Smaller paddings in alerts

Polymoria:
I like yellow and purple, but this time i think its time to reduce amount of yellow/purple empty space


ImUrX:
maybe make the padding be modifiable then?

Polymoria:
Weird boolean name fixed to bgDark :3 Sorry for no comment on commit

PR #1459, opened 2025-06-02
Add support for Linux registry checking for VRChat

ImUrX:
Requires testing in both operating systems, but Windows should work as expected


loucass003 (approved):
Sick good job

kruemmelbande:
Works fine for me umu

server/desktop/src/main/java/dev/slimevr/desktop/games/vrchat/RegEdit.kt
```diff
@@ -0,0 +1,149 @@
...
+			LogManager.severe("[VRChatRegEdit] Error reading Values from VRC registry: ${e.message}")
+		}
+		registry = map
+		return keysMap
+	}
+
+	companion object {
+		const val USER_REG_SUBPATH = "Steam/steamapps/compatdata/438100/pfx/user.reg"
+		val USER_REG_PATH = System.getenv("XDG_DATA_HOME")?.let { Path(it, USER_REG_SUBPATH) }
+			?: System.getenv("HOME")?.let { Path(it, ".local", "share", USER_REG_SUBPATH) }
```
ImSapphire: I believe the Steam path is special for Flatpak and Debian, on debian the datadir is at `$HOME/.steam/debian-installation/`, on Flatpak the datadir is at `$HOME/.var/app/com.valvesoftware.Steam/data/Steam/`.

ImUrX: fun

ImUrX: the problem is what if a user has multiple installations aka flatpak and via package manager :p

ImSapphire: I would probably prioritise as such:
first `$HOME/.steam/steam/` since it's the normal configuration,
then `$HOME/.steam/debian-installation/`,
and then finally `$HOME/.var/app/com.valvesoftware.Steam/data/Steam/`.

Maybe if there are multiple present, a warning could also be presented in the UI just in case it picks the wrong one? Probably too much effort for what it's worth though.

ImUrX: I will probably use the first config folder presented that I find in `~/.config/openvr/openvrpaths.vrpath` as thats probably what steam is being used for VR

ImSapphire: That usually won't be filled in when not using SteamVR and instead using Monado (through Envision) or WiVRn with OC. WiVRn only sets `runtime` and `version` fields since those are the only ones necessary for Proton to be happy, not sure what Envision does.

ImUrX: :sob:

ImUrX: should be fixed now

PR #1462, opened 2025-06-03
Fix missing part of reg key for VRChat

ImUrX:
https://discord.com/channels/817184208525983775/948579740225261598/1379204024514908393


ButterscotchV:
I see both `VRC_IK_LEGACY` and `VRC_IK_LEGACY_CALIBRATION` in my registry 👀

ImUrX:
Can you switch `VRC_IK_LEGACY` in yours? I was only testing this on Desktop VRChat so maybe this setting is not visible at all without SteamVR on

ImUrX:
not needed at the end i think

PR #1465, opened 2025-06-06
Cpu fixes

Philip6294:


Philip6294:
this should not have been opened against the upstream branch, sorry

PR #1466, opened 2025-06-09
Fix neck length not being able to be used for VMC

Erimelowo:
This basically uses the neck position instead of the head position for VMC when anchoring from the head.

This behaviour change is only relevent for people using positional trackers (or HMDs) for the head, as using a head tracker with no position will have the neck length (and head shift) set to 0 in the skeleton directly.

Relevant Discord thread: https://discord.com/channels/817184208525983775/1380933378425229513/1381713892971122891


ImUrX (approved):

ButterscotchV (approved):
I don't really like how neck height is calculated but it seems like the most sensible way

PR #1470, opened 2025-06-13
Check monitor size before recovering old window state

ImUrX:
I been trying [olekolek1000/wayvr-dashboard](https://github.com/olekolek1000/wayvr-dashboard) and discovered that SlimeVR will recover the old size anyways even though it's bigger than the screen itself, so I basically added a check for the display size.

I'm not sure if this bug affects Windows but at least it will fix this in Linux wayland compositors at least.


gui/src-tauri/src/state.rs
```diff
@@ -80,17 +80,35 @@ impl WindowState {
...
+			let monitor_size = *monitor.size();
+			let window_size = size
+				.map(|s| s.to_physical(monitor.scale_factor()))
+				.unwrap_or(window.outer_size()?);
+			window.set_size(PhysicalSize::new(
+				u32::min(monitor_size.width, window_size.width),
+				u32::min(monitor_size.height, window_size.height),
+			))?;
+
+			if is_old {
```
ButterscotchV, line 111: `is_old` isn't very descriptive, should probably rename or add a comment explaining it

ImUrX: that better?

ButterscotchV (approved):
Happy happy happyyy

PR #1479, opened 2025-06-22
Support inches for simple tasks

ImUrX:
I added support for inches in the places that I thought were simple to start with, please do tell if I forgot any apart from the manual proportions which is a little more complicated.


gui/src/components/onboarding/pages/body-proportions/scaled-steps/ManualHeightStep.tsx
```diff
@@ -104,17 +103,23 @@ export function ManualHeightStep({
...
-              min={MIN_HEIGHT}
-              max={4}
-              step={0.01}
-              showButtonWithNumber
-              doubleStep={0.1}
+              min={convert(MIN_HEIGHT, 'm').to(currentUnit)}
+              max={convert(4, 'm').to(currentUnit)}
+              step={currentUnit === 'm' ? 0.01 : 1}
+              showButtonWithNumber={currentUnit === 'm' ? 10 : 12}
+              doubleStep={currentUnit === 'm' ? 0.1 : 12}
```
ButterscotchV: All this conversion feels super jank ngl, I feel like we should just keep the internal representation as meters and convert to feet and inches for the display. We could probably switch the increment and do some rounding to imperial units tho, I'm really not sure what's best.

ImUrX: Already tried this, because of floating point error you can add 1 inches and one moment you will end up adding 2 inches imprecisely.

The best kind of rounding would be half decimal rounding against every multiplier of an inch in meters

PR #1482, opened 2025-06-25
Add ARM builds in CI

ImUrX:
this bumps the node version for CI to v22 which is the latest LTS, didnt break anything from the testings.


.github/workflows/gradle.yaml
```diff
@@ -367,11 +374,11 @@ jobs:
           cp server/desktop/build/libs/slimevr.jar ./SlimeVR/slimevr.jar
           cp server/core/resources/* ./SlimeVR/
           cp target/release/slimevr.exe ./SlimeVR/
-          7z a -tzip SlimeVR-win64.zip ./SlimeVR/
+          7z a -tzip "SlimeVR-$BUILD_ARCH.zip" ./SlimeVR/
```
ButterscotchV, line 377: This will break the installer, it is expecting `SlimeVR-win64.zip`, see https://github.com/SlimeVR/SlimeVR-Installer/blob/main/windows/web/slimevr_web_installer.nsi#L402.

ButterscotchV (changes requested):

PR #1483, opened 2025-06-25
Add save file prompt for BVH recordings

ImUrX:
Needs https://github.com/SlimeVR/SolarXR-Protocol/pull/175


ImUrX:
needs testing on windows and macOS

loucass003:
Have we asked the mocap people that this is what they want? 
I would rather have a setting in the gui or server that says if you want to specify the path before saving or use a generated name

ImUrX:
The current way is plain wrong, files should not be created next to the binary in program files

loucass003:
I agree on that but then why not selecting a better default path?

ImUrX:
Then it would make it more unexpected because you can't find where BVH files are unless you google the new folder.
I'm ok with adding an option for choosing a default folder but choosing where to save the file itself sounds more intuitive.

loucass003:
I agree, let's add an option to set the default folder in the settings. and put the dialog by default. so we don't have to provide a folder by default on our end

ButterscotchV (approved):

PR #1488, opened 2025-07-02
include udev rules in the linux installers

ImUrX:


kruemmelbande (approved):
Can confirm, has udev rules in the linux packages umu

loucass003 (approved):

PR #1490, opened 2025-07-04
Yaw reset smoothing fix

ButterscotchV:
Fixes reset smoothing counting for filtering since filtering was moved to *after* resets are performed. This PR moves reset smoothing to after filtering and creates a standalone class that handles the interpolation.

I tested this quickly with one tracker and it seems to work fine.


loucass003 (approved):
LGTM

PR #1491, opened 2025-07-08
Staggered fw upate system

loucass003:


Eirenliel (approved):

ImUrX:
this is filled with iterator methods, breaks supports with linux and old macOS

PR #1494, opened 2025-07-09
Fix deploy data fetching

loucass003:


gui/src/hooks/firmware-update.ts
```diff
@@ -40,31 +40,34 @@ const processDeployData = (sortedMap: [string, string][]) => {
...
-  if (!deployAssetUrl) return false;
-  const deployDataJson: DeployData | null = await fetch(deployAssetUrl)
-    .then((res) => res.json())
-    .catch(() => null);
-  if (!deployDataJson) return false;
+const checkUserCanUpdate = (body: string) => {
+  const matches = [...body.matchAll(/```json\r?\n([\s\S]*?)```/g)];
+  if (matches.length === 0) return true;
+  const deployDataJson = JSON.parse(matches.at(-1)?.[1] ?? 'null');
+  if (!deployDataJson) return true;
```
ImUrX: this regex is scary

loucass003: it is what it is ....
i cant grab a file from the assets of the release with the gui so. 
Unless you find a better regex screw it

ImUrX: you can grab it tbh, but I think it's less network heavy to use the body itself
```js
fetch(release.assets.find(x => x.name === "deploy.json").browser_download_url)
```

loucass003:
@ImUrX Please confirm compatibility with linux / macos

gui/src/hooks/firmware-update.ts
loucass003: NOPE
CORS error

ImUrX: ah yeah makes sense

loucass003:
@ImUrX i switched to the tauri plugin for http calls. Seemed like the better option.
I had to bump the rust toolchain. if you can double check 🙏

gui/src/hooks/firmware-update.ts
```diff
@@ -87,18 +78,15 @@ export async function fetchCurrentFirmwareRelease(): Promise<FirmwareRelease | n
...
     if (version.charAt(0) === 'v') {
       version = version.substring(1);
     }
 
-    const userCanUpdate = !deployAsset
-      ? true
-      : await checkUserCanUpdate(deployAsset?.browser_download_url);
-
+    const deployAsset = firstAsset(release.assets, 'deploy.json');
+    const userCanUpdate = await checkUserCanUpdate(deployAsset?.browser_download_url);
```
ImUrX: i have an eslint rule on the eslint PR failing from having an await inside this for loop, can you use `Promise.all()`?

gui/src/hooks/firmware-update.ts
```diff
@@ -26,47 +24,40 @@ const hash = (str: string) => {
...
-  }
-  return deployData;
-};
-
-const checkUserCanUpdate = async (deployAssetUrl: string) => {
-  if (!deployAssetUrl) return false;
-  const deployDataJson: DeployData | null = await fetch(deployAssetUrl)
+const checkUserCanUpdate = async (url: string) => {
+  if (!url) return true;
+  const deployDataJson = await fetch(url)
```
ImUrX: not gonna cache the request?

ImUrX (changes requested):

gui/src/hooks/firmware-update.ts
loucass003: i really dont want to. it would be really annoying to filter out

gui/src/hooks/firmware-update.ts
loucass003: There is almost no rate limit on raw.github. Should be fine

loucass003: Added cache anyway

gui/src/hooks/firmware-update.ts
ImUrX: ok i see

ImUrX (approved):

PR #1501, opened 2025-07-17
Use scroll snap at the end for serial

ImUrX:
Add scroll snapping to the end of the serial so it only scrolls when you are at the end of the console.

Breaks mouse scroll in Linux webkit when the overflow is high enough, but I think being able to use scrollbar is good enough


ButterscotchV (approved):
Tested on Windows 11 and appears to work as intended

PR #1506, opened 2025-07-26
Show new mounting resets if they have related body part assigned

ImUrX:
There is so many reset buttons now, we need to hide them if they are not really needed, please test if they work as I can't make my typical slimevr-wrangler test binary work right now :sob:

### DIFF
--- a/gui/src/components/home/ResetButton.tsx
+++ b/gui/src/components/home/ResetButton.tsx
@@ -23,6 +23,57 @@ import {
 } from '@/components/commons/icon/ResetIcon';
 import { useStatusContext } from '@/hooks/status-system';
 import classNames from 'classnames';
+import { atom, useAtomValue } from 'jotai';
+import { assignedTrackersAtom } from '@/store/app-store';
+
+const FEET_BODY_PARTS = [BodyPart.LEFT_FOOT, BodyPart.RIGHT_FOOT];
+const FINGER_BODY_PARTS = new Set([
+  BodyPart.LEFT_THUMB_METACARPAL,
+  BodyPart.LEFT_THUMB_PROXIMAL,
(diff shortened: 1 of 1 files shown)


gui/src/components/home/ResetButton.tsx
```diff
@@ -173,6 +192,13 @@ export function ResetButton({
     };
   }, []);
 
+  if (
+    (bodyPartsToReset === 'fingers' && !hasFingerTrackers) ||
+    (bodyPartsToReset === 'feet' && !hasFeetTrackers)
+  ) {
+    return null;
+  }
```
ButterscotchV: This will be weird if we have some kinda step that uses this button and the button just vanishes lol, maybe the parent should control it? or like have it disabled by this?

loucass003: i would go for a disable or the button in the Component code. 
And if it needs to be hidden it should be handled by the parent component i agree

gui/src/components/home/ResetButton.tsx
```diff
@@ -23,6 +23,57 @@ import {
...
+  BodyPart.RIGHT_RING_PROXIMAL,
+  BodyPart.RIGHT_RING_INTERMEDIATE,
+  BodyPart.RIGHT_RING_DISTAL,
+  BodyPart.RIGHT_LITTLE_PROXIMAL,
+  BodyPart.RIGHT_LITTLE_INTERMEDIATE,
+  BodyPart.RIGHT_LITTLE_DISTAL,
+]);
+
+const fingerAssignedTrackers = atom((get) =>
+  get(assignedTrackersAtom).some(
```
loucass003: i would keep the atoms in the store file.
Easier to find common branches

PR #1508, opened 2025-07-29
Fix mounting reset not working for feet

ButterscotchV:
This fixes the headset reference not being used for resets that don't include the head. Also changes the tap reset default back to not including feet.
Since we don't have the foot reset toggle anymore, the GUI must be used to reset, we can maybe fix this in a better way after this PR?


loucass003 (approved):

Erimelowo (approved):

Erimelowo:
>  Since we don't have the foot reset toggle anymore, the GUI must be used to reset, we can maybe fix this in a better way after this PR?

Doing that

PR #1509, opened 2025-07-29
Bring back resetMountingFeet toggle for mounting resets

Erimelowo:
- This brings back resetMountingFeet, which was removed in https://github.com/SlimeVR/SlimeVR-Server/pull/1441 in favor of separate resets for feet and the rest of the body. Users may still want to do all mounting rests in one so this setting was added back. Now, users can either do feet mounting reset separately or at the same time. Separate feet reset is still available even if resetMountingFeet is true (could be useful if feet are borked but the rest of the body is fine).
- Added some comments in the code because of how jank all of this is. All mounting resets ask for feet to be reset but they will only be reset if:
1. Feet are the only body parts that are asked to be mounting reset (separate feet reset button) or
2. resetMountingFeet is true (in this case, individual trackers' resets handlers will handle that. This is the previous behaviour.)
- I also changed the text for resetMountingFeet in GUI to "Force feet mounting reset" instead of "Enable feet mounting reset".
- Layout of mounting resets buttons was changed so that the "Clear Reset Mounting" button is at the end and icons were changed to feet and hands for the corresponding mounting resets
slimevr_2Bm0q2CDh8
slimevr_D4Z5LsG6M1


Erimelowo:
Made a beta on Discord

ButterscotchV:
> Maybe a silly question, but are those two buttons only shown if you actually use the feature?
> I mean, if someone doesn’t track their fingers or doesn’t have trackers on their feet, they wouldn’t need those buttons, right?

The buttons could be hidden, yes, see #1506

PR #1519, opened 2025-08-21
Foot snapping fix

ButterscotchV:
The current way foot plant isolates the yaw from the tracker orientation does not work with VRChat IK. It is technically correct for the quaternion, but the euler yaw resulting from it causes snapping. This PR uses a method that does *not* cause snapping in VRChat.

Please feel free to recommend a better way of fixing this, as this was just the first thing I came up with. One alternative could be using the project function of the quaternion, though that may have its own issues.


Erimelowo (approved):
lgtm

ImUrX (approved):

Stermere (approved):

PR #1520, opened 2025-08-24
Fix endian docs for smol HID

9il:


ButterscotchV (approved):

ButterscotchV:
Actually might be better to describe as *from* what *to* what because this is still a little confusing

9il:
everything in smol is little :)

sctanf (approved):

PR #1521, opened 2025-08-25
Calibration tutorial fix settling

sctanf:
Not using hasCompletedRestCalibration yet, but fixes settling detection for trackers that will rely on old behavior

### DIFF
--- a/gui/src/components/onboarding/pages/CalibrationTutorial.tsx
+++ b/gui/src/components/onboarding/pages/CalibrationTutorial.tsx
@@ -45,16 +45,17 @@ export function CalibrationTutorialPage() {
   useEffect(() => {
     const accelLength = restCalibrationTrackers.every((x) => {
       if (
+        x.device?.id?.id === undefined ||
         x.tracker.trackerId?.trackerNum === undefined ||
         x.tracker.trackerId.deviceId?.id === undefined ||
         !x.tracker.linearAcceleration
       )
         return false;
 
-      const trackerId =
-        x.tracker.trackerId.trackerNum + (x.tracker.trackerId.trackerNum << 8);
-      const lastValues = lastValueMap.current.get(trackerId) ?? [];
-      lastValueMap.current.set(trackerId, lastValues);
+      const trackerId = x.device.id.id;
+//        x.tracker.trackerId.trackerNum + (x.tracker.trackerId.deviceId.id << 8);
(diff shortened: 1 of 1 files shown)


gui/src/components/onboarding/pages/CalibrationTutorial.tsx
```diff
@@ -32,42 +34,62 @@ export function CalibrationTutorialPage() {
...
+      abortCountdown();
+      startCountdown();
+    } else if (!accelLength && !settled && isCounting) {
+      abortCountdown();
+    } else if (!accelLength && settled) {
+      setSettled(false);
+    }
+
+    setRested(settled || restCalibrationTrackers.length === 0);
   }, [restCalibrationTrackers]);
```
ImUrX:
```suggestion
  }, [restCalibrationTrackers, settled, isCounting]);
```
need to add it as a dependency if you use it inside the effect

gui/src/components/onboarding/pages/CalibrationTutorial.tsx
```diff
@@ -32,42 +34,62 @@ export function CalibrationTutorialPage() {
...
         x.tracker.trackerId.deviceId?.id === undefined ||
         !x.tracker.linearAcceleration
       )
         return false;
 
-      const trackerId =
-        x.tracker.trackerId.trackerNum + (x.tracker.trackerId.trackerNum << 8);
-      const lastValues = lastValueMap.current.get(trackerId) ?? [];
-      lastValueMap.current.set(trackerId, lastValues);
+      const trackerId = x.device.id.id;
```
ImUrX: this breaks extensions as its no longer using `trackerNum`

i see that it previously had a bug of using `trackerNum` two times? probably meant to have deviceId in the bitwise operation

ImUrX (changes requested):

gui/src/components/onboarding/pages/CalibrationTutorial.tsx
sctanf: it seems like it works correctly anyway, but i will change it to account for trackerNum
deviceId is the same as device.id, i will use the latter because i think it is a little clearer what it represents

ImUrX: yeah thats okay

ImUrX (approved):

Eirenliel (approved):

PR #1532, opened 2025-09-04
HID support on Android

### LINKED ISSUE #1525: Support HID trackers on Android
This should be able to be done using Android's USB host API (https://developer.android.com/develop/connectivity/usb/host)

ButterscotchV:
Adds basic HID dongle support for Android.
Closes #1525

## Features
- Moves HID packet handling to SlimeVR core to share code between PC and Android
- Prompts to open app when dongle is connected
- Uses Android's USB events to request permission immediately with minimal CPU impact
  - Cleanly handles disconnects and reconnects


server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -0,0 +1,292 @@
...
+		}.toMutableList()
+		synchronized(devicesByHID) {
+			// Work on devicesByHid and add/remove as necessary
+			val removeList: MutableList<UsbDevice> = devicesByHID.keys.toMutableList()
+			removeList.removeAll(hidDeviceList)
+			for (device in removeList) {
+				removeDevice(device)
+			}
+
+			// Quickly reattaching a device may not be detected, so always try to open existing devices
```
sctanf: was this still needed for android as well? if reattaching receivers quickly works, then it probably can be removed

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -0,0 +1,292 @@
...
+import dev.slimevr.tracking.trackers.hid.HIDCommon.Companion.PACKET_SIZE
+import dev.slimevr.tracking.trackers.hid.HIDDevice
+import io.eiren.util.logging.LogManager
+import java.nio.ByteBuffer
+import java.util.function.Consumer
+
+const val ACTION_USB_PERMISSION = "dev.slimevr.USB_PERMISSION"
+
+/**
+ * Receives trackers data by UDP using extended owoTrack protocol.
```
Eirenliel: Nitpick: comment is wrong...

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -0,0 +1,292 @@
...
+ */
+class TrackersHID(
```
Eirenliel: Maybe rename it to something like "AndroidHIDService", currently the name is not matching the purpose and is confusing.

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/UsbDeviceHID.kt
```diff
@@ -0,0 +1,90 @@
...
+
+import android.hardware.usb.UsbConstants
+import android.hardware.usb.UsbDevice
+import android.hardware.usb.UsbDeviceConnection
+import android.hardware.usb.UsbEndpoint
+import android.hardware.usb.UsbInterface
+import android.hardware.usb.UsbManager
+import java.io.Closeable
+
+class UsbDeviceHID(hidDevice: UsbDevice, usbManager: UsbManager) : Closeable {
```
Eirenliel: Rename to something like "AndroidHIDDevice"

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -1,18 +1,15 @@
...
-import com.jme3.math.FastMath
 import dev.slimevr.VRServer
 import dev.slimevr.tracking.trackers.Device
 import dev.slimevr.tracking.trackers.Tracker
 import dev.slimevr.tracking.trackers.TrackerStatus
-import dev.slimevr.tracking.trackers.udp.BoardType
-import dev.slimevr.tracking.trackers.udp.IMUType
-import dev.slimevr.tracking.trackers.udp.MCUType
-import dev.slimevr.tracking.trackers.udp.MagnetometerStatus
+import dev.slimevr.tracking.trackers.hid.HIDCommon
```
Eirenliel: Maybe also rename the class to DesktopHIDManager

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -0,0 +1,292 @@
...
+
+			val oldConn = this.connByHID.remove(hidDevice)
+			val serial = oldConn?.serialNumber ?: "Unknown"
+			oldConn?.close()
+
+			LogManager.info("[TrackerServer] Linked HID device removed: $serial")
+		}
+	}
+
+	private fun deviceIdLookup(hidDevice: UsbDevice, deviceId: Int, deviceName: String? = null, deviceList: MutableList<Int>): HIDDevice? {
```
Eirenliel: Can this also be moved to common code?

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
ButterscotchV: Yeah you're right, that can be removed now :3

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
```diff
@@ -0,0 +1,283 @@
...
+			for ((hidDevice, deviceList) in devicesByHID) {
+				val dataReceived = ByteArray(64)
+				val conn = connByHID[hidDevice]!!
+				val dataRead = conn.deviceConnection.bulkTransfer(conn.endpointIn, dataReceived, dataReceived.size, 0)
+
+				// LogManager.info("[TrackerServer] HID data read ($dataRead bytes): ${dataReceived.contentToString()}")
+
+				devicesPresent = true // Even if the device has no data
+				if (dataRead > 0) {
+					// Process data
```
Eirenliel: Can this also be moved to common code?

server/core/src/main/java/dev/slimevr/tracking/trackers/hid/HIDCommon.kt
```diff
@@ -0,0 +1,288 @@
...
+				}
+			}
+
+			val tracker: Tracker? = device.getTracker(trackerId)
+			if (tracker == null) { // not registered yet
+				return
+			}
+
+			// Packet data
+			var batt: Int? = null
```
Eirenliel, line 141: I think it's time we use a struct/class for this... @sctanf ?

Eirenliel (changes requested):

server/android/src/main/AndroidManifest.xml
```diff
@@ -26,7 +28,11 @@
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
+                <action android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED" />
             </intent-filter>
+
+            <meta-data android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED"
+                android:resource="@xml/device_filter" />
```
ButterscotchV, line 35: One note about this, it seems that the "Open SlimeVR" popup shows up regardless of whether the app is already running, kinda annoying ngl

server/android/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
+    <uses-feature android:name="android.hardware.usb.host" />
```
ButterscotchV: I'm not sure if specifying this means that the app *requires* this API to function. On the docs, they say that this indicates that the app *uses* the APIs, but not what the consequence of specifying this does.

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
ButterscotchV: Yeah I just copied from the desktop code, I can clean up both, I just wanted to keep as few differences as possible to make the PR clearer

Eirenliel: Fair, but let's at least update naming to have easier development and refactoring going further.

server/android/src/main/AndroidManifest.xml
Eirenliel: I don't think it's bad?

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
ButterscotchV: It was kinda a pain because it uses the OS specific USB device class extensively in this part, so it was left OS specific

server/android/src/main/AndroidManifest.xml
ImUrX: it means [it needs it for it to run](https://developer.android.com/guide/topics/manifest/manifest-intro#uses-feature)

ButterscotchV: > For example, if your app can't achieve basic functionality on a device without a compass sensor ...

So this probably shouldn't be specified then, got it

ImUrX: you can make it optional by adding a `android:required="false"` inside it and probably checking for the feature being available via [`hasSystemFeature()`](https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_USB_HOST)

server/core/src/main/java/dev/slimevr/tracking/trackers/hid/HIDCommon.kt
sctanf: i am not sure, since the data here can be partial/aggregate from multiple packets, and it is only used here to update the device or tracker struct below

server/android/src/main/java/dev/slimevr/android/tracking/trackers/hid/TrackersHID.kt
ButterscotchV: Updated the comments on both sides to reflect the functionality

server/android/src/main/AndroidManifest.xml
ButterscotchV: > you can make it optional by adding a `android:required="false"` inside it and probably checking for the feature being available via [`hasSystemFeature()`](https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_USB_HOST)

ty, made it non-required, hopefully that should be good enough

server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/TrackersHID.kt
ButterscotchV: I went with "DesktopHIDService" to match the Android side, idk if we prefer "Service" or "Manager", I think of it more as a service since it's kinda separate from the rest of the code? Though it technically manages HID stuff sooo idk?

ButterscotchV: Actually maybe "Manager" *is* better since service usually implies it is providing access to something, whereas this is just accessing things.

server/android/src/main/AndroidManifest.xml
ImUrX: have u tried this on quest? apparently it depends on the system if it crashes or not
image

Eirenliel (approved):

PR #1533, opened 2025-09-04
Fix Android serial support

### LINKED ISSUE #1334: Serial only works periodically on Android
Serial may not be functioning completely on Android. This has been an issue for a while and supposedly did not happen before. Sometimes only works after rebooting device or app, only functioning once. Needs further testing and research.

ButterscotchV:
Fixes the major issues with serial on Android:
- Uses Android USB events to enumerate serial devices rather than polling
- Clears permission request on device removal or permission approval
  - Subsequentent connections now re-request instead of blocking due to the previous request
- Closes serial port on disconnect
  - No longer hangs after removal or on reconnection

Fixes #1334

### DIFF
--- a/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
+++ b/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
@@ -223,7 +223,7 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
 	@Synchronized
 	private fun writeSerial(serialText: String, print: Boolean = false) {
 		try {
-			usbIoManager?.writeAsync("${serialText}\n".toByteArray())
+			currentPort?.port?.write("${serialText}\n".toByteArray(), 0)
 			if (print) {
 				addLog("-> $serialText\n")
 			}
@@ -272,7 +272,7 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
 	}
 
 	override fun write(buff: ByteArray) {
-		usbIoManager?.writeAsync(buff)
+		currentPort?.port?.write(buff, 0)
 	}
 
 	@Synchronized
--- a/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
+++ b/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
@@ -287,6 +287,6 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
 	override fun onRunError(e: java.lang.Exception?) {}
 
 	companion object {
-		private val ACTION_USB_PERMISSION = "dev.slimevr.USB_PERMISSION"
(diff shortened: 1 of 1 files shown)


ImUrX (approved):
looks great :3

server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
```diff
@@ -43,12 +44,11 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
 	private var usbIoManager: SerialInputOutputManager? = null
 
 	private val listeners: MutableList<SerialListener> = CopyOnWriteArrayList()
-	private val getDevicesTimer = Timer("GetDevicesTimer")
-	private var watchingNewDevices = false
 	private var lastKnownPorts = setOf<SerialPortWrapper>()
 	private val manager = activity.getSystemService(Context.USB_SERVICE) as UsbManager
 	private var currentPort: SerialPortWrapper? = null
 	private var requestingPermission: String = ""
+	private var readBuffer: StringBuilder = StringBuilder(4096)
```
ImUrX:
```suggestion
	private var readBuffer: StringBuilder = StringBuilder(1024)
```
i think 1024 chars is enough?

server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
```diff
@@ -251,14 +290,21 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
...
 	override fun onNewData(data: ByteArray?) {
 		if (data != null) {
-			val s = StandardCharsets.UTF_8.decode(ByteBuffer.wrap(data)).toString()
-			addLog(s, false)
+			// Collect serial in a buffer until newline (or character limit)
+			// This is somewhat of a workaround for Android serial buffer being smaller
+			//  than on desktop, so we don't read full lines and it causes parsing issues
+			readBuffer.append(StandardCharsets.UTF_8.decode(ByteBuffer.wrap(data)))
+
+			if (readBuffer.contains('\n') || readBuffer.length >= 4096) {
```
ImUrX:
```suggestion
			if (readBuffer.contains('\n') || readBuffer.length >= 1024) {
```

server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
ButterscotchV: I originally set it to 512, but changed it to 4096 to match the desktop implementation. I could change it to 1024 though, maybe it would be better as that.

ImUrX: sounds better to have it smol for android devices with not many resources

ButterscotchV: Using 1024 now

Eirenliel (approved):

Issue #1534, opened 2025-09-04
HumanSkeleton race condition


ButterscotchV:
There seems to be a race condition with HumanSkeleton. I experienced a NullPointerException after a null check in `Constraint`, this is likely due to the skeleton being disassembled in `HumanSkeleton.assembleSkeletonArms()` from `HumanSkeleton.updateToggleState()` with `SkeletonConfigToggles.FORCE_ARMS_FROM_HMD`, called by `RPCSettingsHandler.onChangeSettingsRequest()` on the RPC thread.

<img width="987" height="291" alt="Image" src="https://github.com/user-attachments/assets/3c539b19-5479-4c48-928c-ed81fc1c517b" />

ButterscotchV:
Can be reproduced with relative ease on Android by repeatedly switching between the settings and home tabs

Eirenliel:
If something wants to modify skeleton it should send the task to the main thread. It also seems like the skeleton gets modified every time you close settings, which is obviously incorrect, you should check for modifications. I think it's a trivial fix and at least it needs to be documented somewhere before we accept the hacky fix...

#1535 is a hack, not a fix. So this will be open.

PR #1535, opened 2025-09-04
Race condition patch for constraints

ButterscotchV:
This should work around #1534 so it at least doesn't crash, but we *really* shouldn't be having any sort of race conditions like this. We will need to touch on this in more depth later.

During testing, I was not able to reproduce the crash after this change.


Eirenliel (approved):

Eirenliel:
I don't like it 😢 I had an idea for how multi-threading should work in this, and read-write locks are a very very bad concurrency tool. You shouldn't be using them, that's how you shoot yourself in the leg.

If something wants to modify skeleton it should send the task to the main thread. It also seems like the skeleton gets modified every time you close settings, which is obviously incorrect, you should check for modifications. I think it's a trivial fix and at least it needs to be documented somewhere before we accept the hacky fix...

PR #1546, opened 2025-09-17
Add Positional UDP packet

SummerSigh:


ImUrX:
Mind making a PR to this file? https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/network/packets.h#L56

Just reserving the enum would be fine

SummerSigh:
@ImUrX https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/471

ButterscotchV (approved):

ImUrX (approved):

server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
```diff
@@ -576,6 +576,13 @@ class TrackersUDPServer(private val port: Int, name: String, private val tracker
 				tracker.dataTick()
 			}
 
+			is UDPPacket27Position -> {
+				tracker = connection?.getTracker(packet.sensorId)
+				if (tracker == null) return
+				tracker.position = packet.position
+				// dont call dataTick here as this is just position update
```
Eirenliel, line 583: Why we don't do data tick here? We need to update stuff based on new position though?

We also should probably have a PositionRotation packet that has both to save on a bit of data

SummerSigh: me and butter talked about this a bit. essentially data tick has filtering stuff (specific to rotation data) that happens when it's called and is otherwise only really used to calculate TPS. PositionRotation is something that could be added for sure, but i was just going to use packet bundling anyway to do rot + pos in a single packet. Can add to this PR if needed 
CC @ButterscotchV

PR #1547, opened 2025-09-17
Disable update prompt on Android

### LINKED ISSUE #1523: Android: Hide server update modal
The new update available modal should not show on Android and update should be handled by play store

ImUrX:
Fixes #1523


ButterscotchV (approved):
Looks good but needs to be tested, I am not entirely sure the BVH button is hidden correctly anymore.

Issue #1106, opened 2024-07-11
Stop allowing people to set any port (osc or vmc) to the tracker port


kruemmelbande:
I see people time and time again messing around with the ports and setting for example the osc input to 6969. I think the gui shouldnt allow you to set any port to the tracker port, since doing so will prevent trackers from connecting.

ZRock35:
I use other ports to route things for VTubing and Mocap purposes routinely. If anything a warning might be in order, but disabling the ability to set ports is a no-go for me.

kruemmelbande:
no, id still want the port to be settable, just not to the tracker port, since setting any port to be the tracker port makes the server stop accepting any trackers at all. A warning might be fine though. If there really is a need to route something through port 6969, you could still change the tracker port in the config, but i guess just adding a warning might be fine too.

PR #1563, opened 2025-09-26
Don't allow the same value for port in and port out

### LINKED ISSUE #1561: Setting OSC in and out ports to the same port causes softlock
If you set both the OSC ports to the same thing, it will cause the software to essentially completely lock up and prevent you from changing the config any more.

gorbit99:
Fixes #1561, also applies the fix to the osc router page. Also disallows the use of ports 6969 and 21110, as that can break the server too (#1106).


gui/src/components/settings/pages/OSCRouterSettings.tsx
```diff
@@ -165,7 +186,26 @@ export function OSCRouterSettings() {
                 <Input
                   type="number"
                   control={control}
-                  rules={{ required: true }}
+                  rules={{
+                    required: true,
+                    validate: (portOut) => {
+                      console.log(portOut, formValues);
```
loucass003: oof console log

gui/src/components/settings/pages/OSCRouterSettings.tsx
```diff
@@ -45,9 +47,61 @@ export function OSCRouterSettings() {
...
+                    'settings-osc-router-network-ports_match_error'
+                  ),
+                  (port, context) => port != context.parent.portOut
+                )
+                .notOneOf(bannedPorts, (context) =>
+                  l10n.getString(
+                    'settings-osc-router-network-port_banned_error',
+                    { port: context.originalValue }
+                  )
+                ),
```
ButterscotchV: Seems like a lotta duplicate code

loucass003: Yep thats fair. you can separate that whole section in a const ouside the component even

gorbit99: Separated into a hook

ButterscotchV (approved):

PR #1566, opened 2025-09-26
Fingertracking, Input and driver version

Erimelowo:
This is a WIP but is almost done.

Driver PR: https://github.com/SlimeVR/SlimeVR-OpenVR-Driver/pull/54


Erimelowo:
I gave up on this as I don’t have enough time to develop it.

We could use OpenGloves instead.

PR #1568, opened 2025-09-27
add CH343 support

H114514191981:
Added support for CH343

### DIFF
--- a/server/core/src/main/java/dev/slimevr/serial/SerialHandler.kt
+++ b/server/core/src/main/java/dev/slimevr/serial/SerialHandler.kt
@@ -29,6 +29,8 @@ abstract class SerialHandler {
 			Pair(0x1A86, 0x7523),
 			// CH341
 			Pair(0x1A86, 0x5523),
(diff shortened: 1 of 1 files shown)


server/core/src/main/java/dev/slimevr/serial/SerialHandler.kt
```diff
@@ -29,6 +29,8 @@ abstract class SerialHandler {
 			Pair(0x1A86, 0x7523),
 			// CH341
 			Pair(0x1A86, 0x5523),
+			// CH343
+			pair(0x1A86, 0x55D3)
```
unlogisch04:
```suggestion
			Pair(0x1A86, 0x55D3),
```

unlogisch04 (commented):
Please also update the udev rules when updating these in:
gui/src-tauri/69-slimevr-devices.rules

loucass003:
@H114514191981 We need more changes from you or im gonna have to close the pr

Issue #1531, opened 2025-09-03
Graph live tracker data


ButterscotchV:
It would be *very* useful to be able to see a graph of selected tracker data in real time, like seeing accel or position over the last minute. Should be pretty much just for dev stuff.

PR #1571, opened 2025-09-29
Add tracker graph to tracker settings page

gorbit99:
Implements #1531 

This PR adds a graph the user can show or hide on the tracker settings page that graphs either the linear acceleration for IMU trackers or the position for positional trackers.

kép

The PR pulls in chart.js and react-chartjs-2 as new dependencies.


Eirenliel:
We really need an ability to graph other stuff too. Like wifi connection, TPS, packet loss, or all of them together...

loucass003:
Does not look really nice tho. Can we spend some time to make it prettier. Also this is super tall. And front colors are not white

gorbit99:
@loucass003 do you have a list of suggestions? I'll apply the ones you already mentioned

gorbit99:
Updated the picture in the original comment

PR #1573, opened 2025-09-30
Delete target web-gui before copying

ButterscotchV:
Currently we just leave all the old files from the web gui in the builds, this makes them progressively bigger each time... Bad bad. Not sure if this will cause issues with failing to load (trying to load from cache or smth?).


loucass003 (approved):
Prob fine

PR #1574, opened 2025-10-01
Add restCalibrationStatus logic to the calibration tutorial

gorbit99:
Depends on https://github.com/SlimeVR/SolarXR-Protocol/pull/182

This PR implements the rest calibration status flag present in the SlimeVR firmware since version 0.5.4 into the server as well. Specifically the desk calibration tutorial now uses this flag if all of the trackers connected to the server support it. If that's not the case, the old behaviour is used as a fallback.

This flag could also be used down the line to warn the user in case their trackers haven't been calibrated yet, but this PR doesn't implement that behaviour.


loucass003:
Conflitcts with #1407 this pr already gives validity check for restCalibration i think we should use the flag coming from the flightlist system instead of duplicating it in the protocol.

Also there is debates about removing that page completely as it wont have a lot of impact during first setup

PR #1604, opened 2025-10-30
Make socket recover if slimevr crashed

loucass003:
Fixes an issue where if slimevr server crash, the deleteOnExit is not necessarly called. Resulting in the server being unable to start anymore unless you delete the socket files manually.

This PR will try to see if the socket is not already used and if not will delete the old one upon restart preventing the deadlock.


notpeelz:
The server probably should also clean up the socket files on SIGINT (ctrl-c) or SIGTERM.

loucass003:
Yes i can confirm that using sigint and sigterm works as intended and remove the socket files.

PR #1612, opened 2025-11-03
Skip starting server from GUI when server is already running

notpeelz:
Skips starting the server if the WebSocket port is in use, allowing running the server via a systemd service without showing an error message every time the GUI opens.
Also adds a command-line argument to skip starting the server.

### DIFF
--- a/gui/src-tauri/src/main.rs
+++ b/gui/src-tauri/src/main.rs
@@ -181,6 +181,8 @@ fn find_server_jar(cli: &Cli) -> Option<PathBuf> {
 				f
 			})
 			.ok(),
+		// For development
+		#[cfg(debug_assertions)]
 		Some(PathBuf::from(env!("CARGO_MANIFEST_DIR"))),
 		// For flatpak container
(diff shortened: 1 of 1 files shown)


gui/src-tauri/src/main.rs
```diff
@@ -309,7 +337,38 @@ fn setup_tauri(
...
+						MINIMUM_JAVA_VERSION
+					));
+					return None;
+				};
+				log::info!("Using Java binary: {:?}", java_bin);
+
+				Some(ServerStartInfo {
+					java_bin,
+					server_jar,
+				})
```
ButterscotchV, line 360: This is fine as it is but we should probably just match the order of the struct
```suggestion
				Some(ServerStartInfo {
					server_jar,
					java_bin,
				})
```

ButterscotchV (approved):

gui/src-tauri/src/main.rs
notpeelz: Done

PR #1615, opened 2025-11-04
Various rc fixes

gorbit99:
This PR fixes a couple issues encountered on the 0.17.0 RC, specifically:

- Serial console scrolling didn't work
- The "Send" button was taller than the rest, and taller than the input next to it
- The serial console was crashing every time someone selected a serial port other than "Auto"
- The Scaled proprtions page was taller than the viewport, and caused it to scroll
- The skeleton preview after doing automatic proportions calibration wasn't visible


server/core/src/main/java/dev/slimevr/protocol/rpc/serial/RPCSerialHandler.java
ButterscotchV: This needs to be done *after* we try opening the serial port, it should be included in the queued task.

ButterscotchV (changes requested):

ButterscotchV (approved):

PR #1618, opened 2025-11-07
Check driver version

Erimelowo:
This will check the slimevr driver protocol version, starting at 1 and either display, for example:

`[ProtobufBridge] Driver protocol version matches the server protocol version: 1`
or 
`[ProtobufBridge] Driver protocol version (2) doesn't match the server protocol version (1)`
in console.

It can also be used in code to make sure the user has a matching driver for new features.

needs https://github.com/SlimeVR/SlimeVR-OpenVR-Driver/pull/59

### DIFF
--- a/server/desktop/src/main/java/dev/slimevr/desktop/platform/ProtobufBridge.kt
+++ b/server/desktop/src/main/java/dev/slimevr/desktop/platform/ProtobufBridge.kt
@@ -175,8 +175,9 @@ abstract class ProtobufBridge(@JvmField protected val bridgeName: String) : ISte
 	@VRServerThread
 	protected open fun versionReceived(versionMessage: Version) {
 		remoteProtocolVersion = versionMessage.protocolVersion
-		LogManager.info("[ProtobufBridge] Driver protocol version matches the server protocol version: $remoteProtocolVersion")
-		if (remoteProtocolVersion != PROTOCOL_VERSION) {
+		if (remoteProtocolVersion == PROTOCOL_VERSION) {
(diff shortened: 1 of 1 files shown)


server/desktop/src/main/java/dev/slimevr/desktop/platform/ProtobufBridge.kt
```diff
@@ -167,6 +172,15 @@ abstract class ProtobufBridge(@JvmField protected val bridgeName: String) : ISte
 		}
 	}
 
+	@VRServerThread
+	protected open fun versionReceived(versionMessage: Version) {
+		remoteProtocolVersion = versionMessage.protocolVersion
+		LogManager.info("[ProtobufBridge] Driver protocol version matches the server protocol version: $remoteProtocolVersion")
+		if (remoteProtocolVersion != PROTOCOL_VERSION) {
+			LogManager.warning("[ProtobufBridge] Driver protocol version ($remoteProtocolVersion) doesn't match the server protocol version ($PROTOCOL_VERSION)")
+		}
```
ButterscotchV: Seems like an if else gone wonk here, we probably shouldn't say it matches and then say it actually doesn't match

ButterscotchV (changes requested):

server/desktop/src/main/java/dev/slimevr/desktop/platform/ProtobufBridge.kt
Erimelowo: Oops, good catch!

Erimelowo: Fixed

ButterscotchV (approved):

PR #1627, opened 2025-11-18
Remove the need for restart before OTA on newer firmware

loucass003:


ButterscotchV (approved):

gorbit99 (approved):
Tested it on an 0.5.4 version BNO slime, it asked for a restart, tested it on an 0.6.0 ICM slime, it didn't ask for a restart, both finished correctly.

Issue #1570, opened 2025-09-29
Tracker table list is missing the firmware update button


loucass003:
<img width="202" height="117" alt="Image" src="https://github.com/user-attachments/assets/25ca47c6-f312-4c0c-bc16-1720ae628413" />

<img width="970" height="138" alt="Image" src="https://github.com/user-attachments/assets/7ec8ad92-326b-47ce-958f-a050c5bf5c6a" />

if you work on this issue, please start from the #1407 branch as the table has been completely rewritten

PR #1634, opened 2025-11-24
Adding firmware update button to tracker table list

Aed-1:
Implements #1570 
requires #1407


Aed-1:
gonna remake this PR and point it to flightlist branch instead, since that is definately not supposed to happen

PR #1635, opened 2025-11-24
Adding firmware update button to tracker table list

Aed-1:
adds #1570
requires #1407

### DIFF
--- a/gui/src/components/tracker/TrackersTable.tsx
+++ b/gui/src/components/tracker/TrackersTable.tsx
@@ -124,9 +124,7 @@ function Header({
       })}
     >
       <div className={className}>
-        <div className="relative">
-          <Typography id={name} whitespace="whitespace-nowrap" />
-        </div>
+        <Typography id={name} whitespace="whitespace-nowrap" />
       </div>
     </th>
   );


Aed-1:
i think this should work, but i do not have esp slimes or officials to test it with

gui/src/components/tracker/TrackersTable.tsx
```diff
@@ -123,7 +124,9 @@ function Header({
       })}
     >
       <div className={className}>
-        <Typography id={name} whitespace="whitespace-nowrap" />
+        <div className="relative">
+          <Typography id={name} whitespace="whitespace-nowrap" />
+        </div>
```
loucass003: Are you sure this is sill needed?

gui/src/components/commons/FirmwareIcon.tsx
```diff
@@ -0,0 +1,95 @@
...
+        showUpdate !== 'unavailable' &&
+        showUpdate !== 'updated' && (
+          <UpdateIcon showUpdate={'can-update'} size={size} />
+        )}
+    </div>
+  );
+}
+//{showUpdate &&
+// showUpdate !== 'unavailable' &&
+//  showUpdate !== 'updated' && (
```
loucass003: oops

gui/src/components/tracker/TrackersTable.tsx
Aed-1: did not touch that part at all, but the text moves around when deleting it so it should probably stay

gui/src/components/commons/FirmwareIcon.tsx
Aed-1: oops indeed

gui/src/components/tracker/TrackersTable.tsx
loucass003: no you did you added relative parent

loucass003:
it's waiting for flightlist to be merged

PR #1638, opened 2025-12-01
Fix for hands trackers not lining up properly in SteamVR

Sebane1:
A somewhat simple fix for an issue I found during my investigations of using SlimeVR trackers as hand controllers.
I noticed a skeleton offset on the hands was causing them to not align properly in SteamVR, and they now line up correctly with this adjustment.

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/processor/config/SkeletonConfigManager.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/processor/config/SkeletonConfigManager.kt
@@ -290,7 +290,7 @@ class SkeletonConfigManager(
 				nodeOffset,
 				0f,
 				-getOffset(SkeletonConfigOffsets.HAND_Y),
-				-getOffset(SkeletonConfigOffsets.HAND_Z),
+				0f,
 			)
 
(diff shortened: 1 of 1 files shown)


ButterscotchV:
Wouldn't the proper solution be to just set the config to 0? I'm pretty sure the Z offset is for using controllers as hand trackers, which is the majority use-case.

Sebane1:
> Wouldn't the proper solution be to just set the config to 0? I'm pretty sure the Z offset is for using controllers as hand trackers, which is the majority use-case.

I don't think SlimeVR is typically handling the hands at all which is why the offset hasn't previously been problematic. Maybe the config "could" be zero'd out, but I didn't see the benefit of any config storage for this value. I haven't noticed any negative impact between continuing to use proper controllers, or using slimes in place of controllers with this change.

Semantically, the config could exist, but I'm not sure of the benefit.

loucass003:
What is the state of this? i have no expertise on this topic @ButterscotchV do we add it or not?

Erimelowo:
Btw there was a very good reason for the z offset existing being that it was the offset for the input controllers. I set the z offset to default to 0 for the rewrite tho.

PR #1661, opened 2025-12-11
Custom color theme

sctanf:


sctanf:
I think these range slilders should have styling
image

PR #1675, opened 2025-12-12
Make ignore buttons be session only in checklist

loucass003:


server/core/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
```diff
@@ -339,6 +339,12 @@ public ObjectNode convert(
 					}
 				}
 			}
+
+			if (version < 15) {
+				ObjectNode checklistNode = (ObjectNode) modelData.get("trackingChecklist");
+				ArrayNode ignoredStepsArray = (ArrayNode) checklistNode.get("ignoredStepsIds");
+				ignoredStepsArray.removeAll();
```
ImSapphire: these may be null

server/core/src/main/java/dev/slimevr/config/ConfigManager.java
```diff
@@ -1,3 +1,4 @@
+
```
ImSapphire: should be removed

server/core/src/main/java/dev/slimevr/config/CurrentVRConfigConverter.java
loucass003: nice catch

ButterscotchV (approved):
Maybe we should consider having *some* kind of hint that the permanent ignore option is in the little settings menu? Not sure how to make it hint, but not direct people to end up at the same problem as before. My only thought is to make it show some sort of message if they've repeatedly ignored it, but I also think that'd be a pain to implement.

PR #1701, opened 2025-12-31
Rework VRChat OSC address/port handling

### LINKED ISSUE #1309: Using yaw reset with OSC causes trackers to follow HMD
Relevant Discord thread: https://discord.com/channels/817184208525983775/1295244826781945856
For some reason, at some point this became an issue with OSC where the headset movement will affect tracker movement for a while after a yaw reset.

ButterscotchV:
Cleaned up all the OSC port initialization methods to make them clearer and handle match detection better. OSCQuery detection has also been changed to prioritize an OSCQuery entry that matches the configured IP and/or port. Also prioritized OSCQuery detection over using the manually configured address so there won't be duplicate data sent.

Should fix #1309


server/core/src/main/java/dev/slimevr/osc/VRCOSCHandler.kt
```diff
@@ -110,122 +111,217 @@ class VRCOSCHandler(
...
 
-	override fun updateOscReceiver(portIn: Int, args: Array<String>) {
-		// Stop listening
-		val wasListening = oscReceiver != null && oscReceiver!!.isListening
-		if (wasListening) {
-			oscReceiver!!.stopListening()
+	/**
+	 * Close/remove the OSCQuery sender
+	 */
+	fun closeOscQuerySender() {
```
Eirenliel, line 217: if your addOscQuerySender is synchronized, that should be too, probably.

Eirenliel (approved):
I mean it's not really thread safe even with this fix, but like... so is life.

server/core/src/main/java/dev/slimevr/osc/VRCOSCHandler.kt
ButterscotchV: Changed to just use VRServer thread, all the rest are on it anyways so we should probably just keep that up

loucass003:
There is conflict on these changes. Did it get fixed by another pr? What is the status on this

PR #1702, opened 2026-01-02
Acceleration based mounting

ButterscotchV:
Adds an acceleration based automatic mounting option (step mounting).


gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -813,46 +813,64 @@ export function GeneralSettings() {
...
-                    outlined
-                    control={control}
-                    name="resetsSettings.resetMountingFeet"
-                    label={l10n.getString(
-                      'settings-general-fk_settings-leg_fk-reset_mounting_feet-v1'
-                    )}
-                  />
-                </div>
-              </div>
+            <div className="grid sm:grid-cols-1 gap-3 pb-3">
```
loucass003, line 931: can we stop copy pasting without checking what it does. 
You are making a grid with only one item 🤷

PR #1709, opened 2026-01-09
add one euro filter

SummerSigh:
adds code for one euro filter


ButterscotchV (approved):
Could use `TimeSource.Monotonic` (https://kotlinlang.org/docs/time-measurement.html), but this is alright too (TimeSource uses nanoTime internally)

loucass003:
Why are we adding this without using it?

PR #1745, opened 2026-02-11
fix: use base64 BWIFI for WiFi credential provisioning

### LINKED ISSUE #1201: Cannot Set WIFI SSID that Contains Emoji
Using SlimeVR server, I'm unable to set the wifi when the network's SSID contains emoji. The SSID that is sent is `?` rather than the emoji. This works fine using the Butterscotch firmware tool so it appears to be a SlimeVR Server specific issue.

JeromeRGero:
SSIDs containing non-alphanumeric characters (periods, hyphens, etc.) fail to connect when sent via the plain-text SET WIFI serial command.

The firmware already supports SET BWIFI which accepts base64-encoded SSID and password, eliminating all parsing issues. This change switches both the desktop and Android serial handlers to use it.

### DIFF
--- a/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
+++ b/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
@@ -281,8 +281,11 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
 
 	@Synchronized
 	override fun setWifi(ssid: String, passwd: String) {
-		writeSerial("SET WIFI \"${ssid}\" \"${passwd}\"")
-		addLog("-> SET WIFI \"$ssid\" \"${passwd.replace(".".toRegex(), "*")}\"\n")
+		val encoder = java.util.Base64.getEncoder()
+		val b64ssid = encoder.encodeToString(ssid.toByteArray(StandardCharsets.UTF_8))
+		val b64passwd = encoder.encodeToString(passwd.toByteArray(StandardCharsets.UTF_8))
+		writeSerial("SET BWIFI $b64ssid $b64passwd")
+		addLog("-> SET BWIFI $b64ssid ${b64passwd.replace(".".toRegex(), "*")}\n")
 	}
 
 	override fun getCurrentPort(): SlimeSerialPort? = this.currentPort
--- a/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
+++ b/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
@@ -284,7 +284,7 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
 		val b64ssid = Base64.Default.encode(ssid.encodeToByteArray())
 		val b64passwd = Base64.Default.encode(passwd.encodeToByteArray())
 		writeSerial("SET BWIFI $b64ssid $b64passwd")
-		addLog("-> SET BWIFI $b64ssid ${b64passwd.replace(".".toRegex(), "*")}\n")
+		addLog("-> SET BWIFI $b64ssid ${"*".repeat(b64passwd.length)}\n")
 	}
 
 	override fun getCurrentPort(): SlimeSerialPort? = this.currentPort
--- a/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
+++ b/server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
@@ -20,6 +20,8 @@ import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.stream.Stream
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
 import kotlin.streams.asSequence
 import kotlin.streams.asStream
 import dev.slimevr.serial.SerialPort as SlimeSerialPort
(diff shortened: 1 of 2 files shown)


server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
```diff
@@ -281,8 +281,11 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
 
 	@Synchronized
 	override fun setWifi(ssid: String, passwd: String) {
-		writeSerial("SET WIFI \"${ssid}\" \"${passwd}\"")
-		addLog("-> SET WIFI \"$ssid\" \"${passwd.replace(".".toRegex(), "*")}\"\n")
+		val encoder = java.util.Base64.getEncoder()
+		val b64ssid = encoder.encodeToString(ssid.toByteArray(StandardCharsets.UTF_8))
+		val b64passwd = encoder.encodeToString(passwd.toByteArray(StandardCharsets.UTF_8))
```
ButterscotchV: [Kotlin's Base64](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io.encoding/-base64/) should probably be used instead of Java's, ex. `Base64.Default.encode(ssid.encodeToByteArray())`.

```suggestion
		val b64ssid = Base64.Default.encode(ssid.encodeToByteArray())
		val b64passwd = Base64.Default.encode(passwd.encodeToByteArray())
```

server/desktop/src/main/java/dev/slimevr/desktop/serial/DesktopSerialHandler.kt
```diff
@@ -199,9 +199,12 @@ class DesktopSerialHandler :
 		val os = currentPort?.outputStream ?: return
 		val writer = OutputStreamWriter(os)
 		try {
-			writer.append("SET WIFI \"").append(ssid).append("\" \"").append(passwd).append("\"\n")
+			val encoder = java.util.Base64.getEncoder()
+			val b64ssid = encoder.encodeToString(ssid.toByteArray(StandardCharsets.UTF_8))
+			val b64passwd = encoder.encodeToString(passwd.toByteArray(StandardCharsets.UTF_8))
```
ButterscotchV: [Kotlin's Base64](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io.encoding/-base64/) should probably be used instead of Java's, ex. `Base64.Default.encode(ssid.encodeToByteArray())`.

```suggestion
			val b64ssid = Base64.Default.encode(ssid.encodeToByteArray())
			val b64passwd = Base64.Default.encode(passwd.encodeToByteArray())
```

server/desktop/src/main/java/dev/slimevr/desktop/serial/DesktopSerialHandler.kt
```diff
@@ -199,9 +199,12 @@ class DesktopSerialHandler :
...
+			writer.append("SET BWIFI ").append(b64ssid).append(" ").append(b64passwd).append("\n")
 			writer.flush()
-			addLog("-> SET WIFI \"$ssid\" \"${passwd.replace(".".toRegex(), "*")}\"\n")
+			addLog("-> SET BWIFI $b64ssid ${b64passwd.replace(".".toRegex(), "*")}\n")
```
ButterscotchV: This is some old code and I'm okay with this being accepted, but we should probably just do something like this instead.
```suggestion
			addLog("-> SET BWIFI $b64ssid ${StringUtils.repeat('*', b64passwd.length)}\n")
```

server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
```diff
@@ -281,8 +281,11 @@ class AndroidSerialHandler(val activity: AppCompatActivity) :
...
+		writeSerial("SET BWIFI $b64ssid $b64passwd")
+		addLog("-> SET BWIFI $b64ssid ${b64passwd.replace(".".toRegex(), "*")}\n")
```
ButterscotchV: This is some old code and I'm okay with this being accepted, but we should probably just do something like this instead.
```suggestion
		addLog("-> SET BWIFI $b64ssid ${StringUtils.repeat('*', b64passwd.length)}\n")
```

ButterscotchV (commented):
Looks good

server/desktop/src/main/java/dev/slimevr/desktop/serial/DesktopSerialHandler.kt
ButterscotchV: Or alternatively, using Kotlin's repeat method.
```suggestion
			addLog("-> SET BWIFI $b64ssid ${"*".repeat(b64passwd.length)}\n")
```

server/android/src/main/java/dev/slimevr/android/serial/AndroidSerialHandler.kt
ButterscotchV: Or alternatively, using Kotlin's repeat method.
```suggestion
		addLog("-> SET BWIFI $b64ssid ${"*".repeat(b64passwd.length)}\n")
```

server/desktop/src/main/java/dev/slimevr/desktop/serial/DesktopSerialHandler.kt
```diff
@@ -199,9 +199,11 @@ class DesktopSerialHandler :
 		val os = currentPort?.outputStream ?: return
 		val writer = OutputStreamWriter(os)
 		try {
-			writer.append("SET WIFI \"").append(ssid).append("\" \"").append(passwd).append("\"\n")
+			val b64ssid = Base64.Default.encode(ssid.encodeToByteArray())
+			val b64passwd = Base64.Default.encode(passwd.encodeToByteArray())
+			writer.append("SET BWIFI ").append(b64ssid).append(" ").append(b64passwd).append("\n")
 			writer.flush()
-			addLog("-> SET WIFI \"$ssid\" \"${passwd.replace(".".toRegex(), "*")}\"\n")
+			addLog("-> SET BWIFI $b64ssid ${StringUtils.repeat('*', b64passwd.length)}\n")
```
ButterscotchV: My bad for making this inconsistent! Let's just use the Kotlin stuff.
```suggestion
			addLog("-> SET BWIFI $b64ssid ${"*".repeat(b64passwd.length)}\n")
```

ButterscotchV (approved):
Looks good, thanks for the PR! This was meant to happen a long while ago when `BWIFI` was added to firmware, but it seems to have been lost to time.

Re-iterating to make it clear, this will set the minimum firmware version for setting WiFi through the server to be *at least* v0.4.0.

loucass003:
I would still like the fallback for older firmware for this to be merged.

loucass003:
This will not be ported until the revrite comes out. i will close the PR.  #1798

PR #1747, opened 2026-02-12
Switch to Electron

loucass003:


gui/electron/main/index.ts
```diff
@@ -0,0 +1,410 @@
...
+  screen,
+  shell,
+  Tray,
+} from 'electron';
+import { IPC_CHANNELS } from '../shared';
+import path, { dirname, join } from 'node:path';
+import open from 'open';
+import trayIcon from '../ressources/icons/icon.png?asset';
+import appleTrayIcon from '../ressources/icons/appleTrayIcon.png?asset';
+import javaVersionFile from '../ressources/java-version/JavaVersion.jar?asset';
```
ImSapphire: nit: folder should be renamed 'resources'

gui/src/components/commons/SystemFileInput.tsx
```diff
@@ -5,9 +5,9 @@ import {
   UseControllerProps,
 } from 'react-hook-form';
 import { FileInputContentBlank, FileInputContentFile } from './FileInput';
-import { open } from '@tauri-apps/plugin-dialog';
+import { useElectron } from '@/hooks/electron';
 
-export function InnerTauriFileInput({
+export function InnerSytemFileInput({
```
ImSapphire: nit: typo
```suggestion
export function InnerSystemFileInput({
```

gui/src/components/commons/SystemFileInput.tsx
```diff
@@ -56,7 +71,7 @@ export function TauriFileInput({
       name={name}
       control={control}
       render={({ field: { onChange, value, ref } }) => (
-        <InnerTauriFileInput
+        <InnerSytemFileInput
```
ImSapphire: see above
```suggestion
        <InnerSystemFileInput
```

gui/electron/main/paths.ts
```diff
@@ -0,0 +1,43 @@
...
+import { getPlatform } from './utils';
+
+export const CONFIG_IDENTIFIER = 'dev.slimevr.SlimeVR';
+
+export const getGuiDataFolder = () => {
+  const platform = getPlatform();
+
+  switch (platform) {
+    case 'linux':
+      return join(app.getPath('home'), '.local/share', CONFIG_IDENTIFIER);
```
ImSapphire, line 20: should look in XDG_DATA_HOME
```suggestion
      if (process.env['XDG_DATA_HOME'])
        return join(process.env['XDG_DATA_HOME'], CONFIG_IDENTIFIER);
      return join(app.getPath('home'), '.local/share', CONFIG_IDENTIFIER);
```

.github/CODEOWNERS
ImSapphire, line 1: :cry:

Eirenliel:
LGDR

Erimelowo:
firefox_Tuc91wmafV
holy-

gui/electron/main/index.ts
```diff
@@ -0,0 +1,421 @@
...
+    app.exit(0);
+    return;
+  }
+};
+
+const isServerRunning = async () => !await isPortAvailable(21110)
+
+const spawnServer = async () => {
+  if (options.skipServerIfRunning && await isServerRunning()) {
+    logger.info({ skipServerIfRunning: options.skipServerIfRunning }, 'Server alredy running, skipping');
```
ImSapphire:
```suggestion
    logger.info({ skipServerIfRunning: options.skipServerIfRunning }, 'Server is already running, skipping server start');
```

gui/electron/main/index.ts
```diff
@@ -0,0 +1,421 @@
...
+const checkEnvironmentVariables = () => {
+  const to_check = ['_JAVA_OPTIONS', 'JAVA_TOOL_OPTIONS'];
+
+  const set = to_check.filter((env) => !!process.env[env]);
+  if (set.length > 0) {
+    dialog.showErrorBox(
+      'SlimeVR',
+      `You have environment variables ${set.join(', ')} set, which may cause the SlimeVR Server to fail to launch properly.`
+    );
+    app.exit(0);
```
ImSapphire, line 352: Does this need to be a hard error? It could just try to continue as normal afterward.

gui/electron/main/index.ts
```diff
@@ -0,0 +1,421 @@
...
+    return;
```
ImSapphire: unnecessary return

gui/electron/main/index.ts
```diff
@@ -0,0 +1,421 @@
...
+handleIpc(IPC_CHANNELS.OPEN_URL, (e, url) => {
+  const allowsd_urls = [
+    /steam:\/\/.*/,
+    /ms-settings:network$/,
+    /https:\/\/.*\.slimevr\.dev.*/,
+    /https:\/\/github\.com\/.*/,
+    /https:\/\/discord\.gg\/slimevr$/,
+  ];
+  if (allowsd_urls.find((a) => url.match(a))) open(url);
+  else logger.error({ url }, 'trying to open non allowed url');
```
ImSapphire:
```suggestion
  const allowed_urls = [
    /steam:\/\/.*/,
    /ms-settings:network$/,
    /https:\/\/.*\.slimevr\.dev.*/,
    /https:\/\/github\.com\/.*/,
    /https:\/\/discord\.gg\/slimevr$/,
  ];
  if (allowed_urls.find((a) => url.match(a))) open(url);
  else logger.error({ url }, 'attempted to open non-whitelisted URL');
```

gui/electron/main/index.ts
```diff
@@ -0,0 +1,421 @@
...
+  if (!javaBin) {
+    dialog.showErrorBox(
+      'SlimeVR',
+      `Couldn't find a compatible Java version, please download Java 17 or higher`
+    );
+    app.exit(0);
+    return;
+  }
+
+  logger.info({ serverJar }, 'found server jar');
```
ImSapphire:
```suggestion
  logger.info({ javaBin, serverJar }, 'Found Java and server jar');
```

gui/electron/main/paths.ts
```diff
@@ -0,0 +1,121 @@
...
+const localJavaBin = (sharedDir: string) => {
+  const jre = join(sharedDir, 'jre/bin', javaBin);
+  return jre;
+};
+
+const javaHomeBin = () => {
+  const javaHome = process.env['JAVA_HOME'];
+  if (!javaHome) return null;
+  const javaHomeJre = join(javaHome, 'bin', javaBin);
+  console.log(javaHomeJre);
```
ImSapphire:
```suggestion
  const javaHomeJre = join(javaHome, 'bin', javaBin);
```

.github/workflows/build-gui.yml
ImSapphire, line 1: The workflows just pack the `out/renderer` folder with all the assets, instead of the actual final build

gui/electron/main/index.ts
loucass003: im reproducing previous tauri behaviour

gui/electron/main/index.ts
```diff
@@ -0,0 +1,421 @@
...
+
+  app.on('before-quit', () => {
+    logger.info('App quitting, saving...');
+    server?.close();
+    stores.settings.save();
+    stores.cache.save();
+
+    discordPresence.destroy();
+
+    writeFileSync(getWindowStateFile(), JSON.stringify(windowState));
```
HannahPadd: Returns an error if folder not exists.

gui/electron/resources/scripts/postinstall.sh
```diff
@@ -0,0 +1,26 @@
...
+
+    if [ "$FOUND_DIR" = true ]; then
+        if command -v udevadm >/dev/null 2>&1; then
+            udevadm control --reload-rules
+            udevadm trigger
+        fi
+    else
+        echo "Warning: No udev rules directory found. Hardware may require manual configuration."
+    fi
+fi
```
ImSapphire: This can be cleaned up a bit (FOUND_DIR variable isn't needed, early returns can be added), and perhaps we should account for udev/rules.d not existing yet:

(not too sure about the `exit 1`, that may cause package installation to fail - do we want that behaviour?)
```suggestion
DESTDIRS=("/lib" "/usr/lib")

if [[ ! -f "$SRC" ]; then
    echo "SlimeVR udev rules not found, serial console and dongles may not work" >&2
    exit 1
fi

echo "Configuring SlimeVR udev rules..."

for DIR in "${DESTDIRS[@]}"; do
    if [[ -d "$DIR" && ! -h "$DIR" ]]; then
        echo "Copying rules to $DIR"
        install -Dm644 "$SRC" "$DIR/udev/rules.d/69-slimevr-devices.rules"

        if command -v udevadm >/dev/null 2>&1; then
            udevadm control --reload-rules
            udevadm trigger
        fi
        exit 0
    fi
done

echo "Couldn't copy SlimeVR udev rules, serial console and dongles may not work" >&2
exit 1
```

loucass003: i dont think we want to fail. the server will tell you there is no udev configured. If the script fail it will become the responsability of the user to fix and the gui should tell you how

ImSapphire (approved):

PR #1749, opened 2026-02-15
Smaller release binaries

TheButlah:
Note: fat LTO adds a minute or two to build time, but results in 23% smaller binaries (52MB -> 41MB)

Time for clean build before this PR: 1:24
After this PR: 2:51


loucass003:
Not worh sending + removed 13mb of bundle from the web size because of a wrongly packaged font. 
There is bigger fish to fry. Even if we kept tauri i wouldnt have accepted the extra build time. 
It already talkes to long during dev.

Issue #25, opened 2021-08-17
Natural Locomotion support


Eirenliel:
To support [Natural Locomotion](https://store.steampowered.com/app/798810/Natural_Locomotion/) SlimeVR need to report speed (or acceleration) values in the driver. This requires support from firmware, software, and driver. Creating issue here for this feature, but it needs to be implemented everywhere.

Louka3000:
What's the status on this?
NaLo no longer crashes... I think all we need to do is forward accel data to SteamVR now, right?
Conveniently, Stermere already did most of the job ;)

BlackCube4:
+1

ruti75:
Hello,
Is there an update on this? I've been trying to write my own code (only using the ankle trackers for simplicoty) and I can share it but I'm a total noob amd was hoping someone did a better job than me.

M1DNYT3:
The changes are merged into the main branch. But **they are not yet in the release**, once they're eventually bundled into the release of the app, the issue should be resolved, and to enable Natural Locomotion support, the toggle in the settings has to be turned on (see the screenshot below). The toggle is off by default, and the setting is persistent, so it'll remember its state between server restarts.

Note that the description mentions the fact that enabling Velocity can introduce jitter in FBT, so it's recommended to only enable Velocity when you're not playing anything where you need smooth visuals of your FBT avatar. The jitter impacts titles like VRChat or BeatSaber's third-person camera with custom FBT avatar mods. The jitter will be looked into, and if there's a minimal, plausible fix, then the note about it will be removed. Otherwise, once the release lands and you get access to this feature, please be aware of this problem.

How it'll look:
<img width="1617" height="892" alt="Screenshot 2026-05-04 153350" src="https://github.com/user-attachments/assets/4da32c96-0fb9-42b5-8fc5-9cbc2ddc310d" />

PR #1753, opened 2026-02-17
Add Velocity calculation logic and GUI toggle (end-to-end MVP).

M1DNYT3:
This PR completes velocity transmission end-to-end, adding RPC settings and a GUI toggle on top of the foundation from #1754.

What's added:
- Protobuf velocity transmission via ProtobufBridge, per-tracker
- RPC settings support - toggle persists across restarts
- GUI toggle in General Settings with i18n

Toggle's GUI Path:
Settings > General > Tracker Settings > Toggle is at the very bottom of the section.

Technicalities:                                                                                                                                                                  
Most of the line count (~400) comes from regenerated ProtobufMessages.java. The handwritten change itself is small.

Depends on https://github.com/SlimeVR/SolarXR-Protocol/pull/202, should be merged only after the submodule is updated to the tracked HEAD.

Addresses the issue: https://github.com/SlimeVR/SlimeVR-Server/issues/25


M1DNYT3:
I had some additional feedback from @ButterscotchV in DMs, addressed it all + rebased to the latest main + also did some changes of my own.

Feedback changes:
- drop `applyVelocityPolicyTo()` in `VRServer.kt`; initially made for expandability, but turned out to be redundant because `VRConfig.applyVelocityPolicy()` already covers expandability.
- replace `System.nanoTime()` with Kotlin's TimeSource; self-explaining, did precisely that.

My own improvement:
Noticed that I forgot to wire `hasVelocity` instead of `allowVelocity` in the ProtobufBridge.kt; the former was created for a safeguard to not send anything if we don't have any velocity in the tracker. Technically, both calls would result in the same behavior, but hasVelocity follows architecture better and explicitly states that there has to be something for it to be sent, whereas allowVelocity simply states whether the toggle is on or off.

server/desktop/src/main/java/dev/slimevr/desktop/platform/ProtobufBridge.kt
```diff
@@ -103,23 +104,38 @@ abstract class ProtobufBridge(@JvmField protected val bridgeName: String) : ISte
 
 	@VRServerThread
 	protected fun writeTrackerUpdate(localTracker: Tracker?) {
-		val builder = ProtobufMessages.Position.newBuilder().setTrackerId(
-			localTracker!!.id,
-		)
-		if (localTracker.hasPosition) {
-			val pos = localTracker.position
+		val tracker = localTracker ?: return
```
ButterscotchV: `localTracker` should just be made non-nullable and you do the null check in `dataWrite()` before calling this method

server/core/src/main/java/dev/slimevr/config/VRConfig.kt
```diff
@@ -104,6 +106,11 @@ class VRConfig {
 		return config
 	}
 
+	/** Resolves value from global config and applies to tracker. */
+	fun applyVelocityPolicy(tracker: Tracker) {
+		tracker.allowVelocity = velocityConfig.sendDerivedVelocity
+	}
```
ButterscotchV: idk if we want to put this here, but we also already have `readTrackerConfig()` so it doesn't really matter much to me. tbh this could probably just be a part of `readTrackerConfig()` and that is called instead.

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -345,6 +365,43 @@ class Tracker @JvmOverloads constructor(
...
+
+		val pos = position
+		val state = velocityState ?: VelocityState().also {
+			velocityState = it
+		}
+
+		val prevMark = state.prevMark
+		if (prevMark != null) {
+			val dt = prevMark.elapsedNow().toDouble(DurationUnit.SECONDS)
+			if (dt in 1e-4..0.25) {
```
ButterscotchV: This should be compared directly using durations instead of converting to a double. I'm not entirely sure what's available from Kotlin so it might be a mess to do, if so, then just ignore this comment.

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -345,6 +365,43 @@ class Tracker @JvmOverloads constructor(
...
+				_velocity = Vector3(
+					((pos.x - state.prevPos.x) / dt).toFloat(),
+					((pos.y - state.prevPos.y) / dt).toFloat(),
+					((pos.z - state.prevPos.z) / dt).toFloat(),
+				)
+			} else {
+				_velocity = Vector3.NULL
+			}
+		}
+		state.prevMark = TimeSource.Monotonic.markNow()
```
ButterscotchV: You could call `markNow()` earlier, use that to compute `dt`, then assign that here to avoid having to call it again.

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -345,6 +365,43 @@ class Tracker @JvmOverloads constructor(
...
+		if (!allowVelocity || !hasPosition) {
+			velocityState = null
+			_velocity = Vector3.NULL
+			return
+		}
+
+		val pos = position
+		val state = velocityState ?: VelocityState().also {
+			velocityState = it
+		}
```
ButterscotchV: `velocityState` could just be non-nullable and initialized with a default `VelocityState()` to avoid this weird bit here. In that case, you could set `velocityState.prevMark = null` in the earlier if statement to invalidate any previous state (though should there really be any previous state then? I wouldn't imagine anything would even be set).

ButterscotchV (changes requested):

server/core/src/main/java/dev/slimevr/config/VRConfig.kt
M1DNYT3: Done. applyVelocityPolicy() was removed entirely, the logic inlined into readTrackerConfig(), and the standalone call in trackerUpdated() dropped since updateTrackersVelocityPolicy() already covers the runtime toggle path. I validated that it works correctly after the change in both cold start and runtime + tracker hotplug. Basically, the setting is working as intended in all scenarios.

M1DNYT3:
Review changes were shipped and tested E2E with SteamVR.

One thing to note. Jitter is still noticeable. I may think of a one-euro filter or some velocity smoothing added, but I suggest that to be a separate PR once this one is shipped.

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -111,7 +120,19 @@ class Tracker @JvmOverloads constructor(
...
+
+	/**
+	 * Velocity state server-side differentiation based on sent poses
+	 */
+	private data class VelocityState(
+		var prevMark: TimeSource.Monotonic.ValueTimeMark? = null,
+		var prevPos: Vector3 = Vector3(0f, 0f, 0f),
+	)
+
+	private var velocityState: VelocityState = VelocityState()
```
ButterscotchV: This can be simplified now.
```suggestion
	private val velocityState = VelocityState()
```

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -345,6 +366,43 @@ class Tracker @JvmOverloads constructor(
...
+					((pos.x - state.prevPos.x) / dtSeconds).toFloat(),
+					((pos.y - state.prevPos.y) / dtSeconds).toFloat(),
+					((pos.z - state.prevPos.z) / dtSeconds).toFloat(),
+				)
+			} else {
+				_velocity = Vector3.NULL
+			}
+		}
+		state.prevMark = now
+		state.prevPos = pos
```
ButterscotchV: `velocityState` can just be used directly.
```suggestion
		val now = TimeSource.Monotonic.markNow()

		val prevMark = velocityState.prevMark
		if (prevMark != null) {
			val dt = now - prevMark
			if (dt in 100.microseconds..250.milliseconds) {
				val dtSeconds = dt.toDouble(DurationUnit.SECONDS)
				_velocity = Vector3(
					((pos.x - velocityState.prevPos.x) / dtSeconds).toFloat(),
					((pos.y - velocityState.prevPos.y) / dtSeconds).toFloat(),
					((pos.z - velocityState.prevPos.z) / dtSeconds).toFloat(),
				)
			} else {
				_velocity = Vector3.NULL
			}
		}
		velocityState.prevMark = now
		velocityState.prevPos = pos
```

ButterscotchV (changes requested):

ButterscotchV (approved):
Looks fine, SolarXR needs to be merged/fixed before this PR.

server/core/src/main/java/dev/slimevr/config/VelocityConfig.kt
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Allows to enable/disable sending of optional derived velocity data via Protobuf.
+ * Enables Natural Locomotion Support
+ * May create overprediction in certain titles causing excessive jitter when moving upper body.
+ */
+class VelocityConfig {
+	// Disables derived velocity for all trackers. Driver zeroes out velocity if nothing is returned in protobuf message.
+	var sendDerivedVelocity: Boolean = false
+
+	fun updateTrackersVelocityPolicy() {
```
ImSapphire: "policy" is a wording not used in the rest of the codebase and was confusing for me on the first read. This should probably be named `updateTrackersVelocitySettings`, like in ResetsConfig.

server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt
```diff
@@ -345,6 +366,42 @@ class Tracker @JvmOverloads constructor(
...
 	}
 
+	/**
+	 * Updates the derived velocity of the tracker by differentiating position over time.
+	 *
+	 * This method enforces the [allowVelocity] policy and checks for valid position data before
+	 * proceeding. If conditions are met, it calculates velocity based on the displacement since the
+	 * last update, applying a sanity check on the time delta to filter out noise and ensure data stability.
+	 *
+	 */
```
ImSapphire, line 373: Could be shortened
```suggestion
	/**
	 * If derived velocity is enabled, updates the derived velocity of the tracker, calculating it based on the position displacement
	 * since the last update, applying a sanity check on the time delta to filter out noise and ensure data stability.
	 *
	 */
```

M1DNYT3: I'm not sure why this is not showing up as an outdated comment, and also doesn't show up in the diff for the PR.
If looking directly at my base branch, you can see the update: https://github.com/M1DNYT3/SlimeVR-Server-NaLo-Support/blob/c9c750230ea73da399fbad0276fd919647b8bea8/server/core/src/main/java/dev/slimevr/tracking/trackers/Tracker.kt#L369

M1DNYT3: Hell, even a rebase didn't help. Seems like an issue on GitHub's end.

Anyway, both comments are addressed and the branch is rebased to the latest main, so the merge should be clean now.

ImSapphire (approved):

M1DNYT3:
Hey @ImSapphire, the last CI run was failing, but not because of my code.
https://github.com/SlimeVR/SlimeVR-Server/actions/runs/25044370421/job/73885369136

This step failed on the GCC lib installation.

I also rebased my branch to the current state of main to make it easier to merge.
If the GCC issue was a one-off or already fixed in main, can you re-run the CI?
Much appreciated!

ImSapphire:
yeah, unfortunately that CI step is a bit flaky right now, due to Canonical being the target of a sustained DDoS attack over the past few days. I've let the CI run again

M1DNYT3:
Hi @Erimelowo and @loucass003,

Just wanted to give a quick heads up since you both are showing as the final decision makers in the chain.
CI is all green, solar-xr submodule is merged into main, the branch is rebased to the current main, I did a smoke test after a rebase as well to verify both GUI and Server are running, and nothing is crashing during the operation around the areas that my changes affected.

Anything else you may need from me, or I just need to just wait for the merge to take place?
Thank you in advance!

loucass003:
sorry im just a bit slow to process stuff that i am not 100% in the loop.
Can someone send a screenshot of how the gui part look?

M1DNYT3:
> sorry im just a bit slow to process stuff that i am not 100% in the loop. Can someone send a screenshot of how the gui part look?

Sure thing. 
The green outline highlights the part that this PR adds.
Screenshot 2026-05-04 153350

One toggle, off by default. Nothing else. Either transmits velocity to SteamVR or not. 

Velocity is derived from position delta over time, following the tick rate of the updates that the server sends. Micro-jitter in FBT is hard to deal with, so me and @ButterscotchV agreed that this PR will ship just the core functionality and figure out the jitter later. Potentially can be fixed by a one-euro filter or smoothing to reduce collision with SteamVR's prediction.

loucass003:
I will also note that i would love to have you around for the server rewrite that is going on soon. Since the feature is gonna have to be re-implemented from zero because of the massive architecture changes. #1798

M1DNYT3:
> I will also note that i would love to have you around for the server rewrite that is going on soon. Since the feature is gonna have to be re-implemented from zero because of the massive architecture changes. #1798

Yeah, sure, no problem at all, either ping me on GitHub or on Slime's Discord server (nickname matches the GitHub one), I'm totally up for supporting the code in the near future.

PR #1759, opened 2026-02-26
Make WEBKIT_DISABLE_DMABUF_RENDERER=1 default (Linux)

ImSapphire:
There's really no reason to not have this by default, it fixes many rendering issues and doesn't seem to hurt in any way.


loucass003:
is this still relevant with electron switch comming?
i havent had to set any env on electron yet

PR #1767, opened 2026-02-28
fix: use transaction id in reset response

notpeelz:
Changes the reset response to use the transaction id of the request, for use in [solarxr-cli](https://github.com/notpeelz/solarxr-cli).

The use case is for implementing a blocking mode (`--wait` option) in `solarxr-cli reset <type>`.
This is especially useful for chaining commands, eg:
```
solarxr-cli reset --delay 3 --wait full && notify-send "Full reset done"
```

solarxr-cli branch: https://github.com/notpeelz/solarxr-cli/tree/feat-reset-transaction

### DIFF
--- a/server/core/src/main/java/dev/slimevr/protocol/rpc/RPCHandler.kt
+++ b/server/core/src/main/java/dev/slimevr/protocol/rpc/RPCHandler.kt
@@ -374,15 +374,14 @@ class RPCHandler(private val api: ProtocolAPI) : ProtocolHandler<RpcMessageHeade
 		}
 	}
 
-	@JvmOverloads
-	fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int, respondTo: RpcMessageHeader? = null): Int {
+	fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int, txId: Long?): Int {
 		val data = IntArray(1)
 
 		RpcMessageHeader.startRpcMessageHeader(fbb)
 		RpcMessageHeader.addMessage(fbb, messageOffset)
 		RpcMessageHeader.addMessageType(fbb, messageType)
-		respondTo?.txId()?.let { txId ->
-			RpcMessageHeader.addTxId(fbb, TransactionId.createTransactionId(fbb, txId.id()))
+		txId?.let { txId ->
+			RpcMessageHeader.addTxId(fbb, TransactionId.createTransactionId(fbb, txId))
 		}
 		data[0] = RpcMessageHeader.endRpcMessageHeader(fbb)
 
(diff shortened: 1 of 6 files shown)


server/core/src/main/java/dev/slimevr/protocol/rpc/RPCHandler.kt
```diff
@@ -393,6 +392,10 @@ class RPCHandler(private val api: ProtocolAPI) : ProtocolHandler<RpcMessageHeade
 		return MessageBundle.endMessageBundle(fbb)
 	}
 
+	fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int): Int = createRPCMessage(fbb, messageType, messageOffset, txId = null)
+
+	fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int, respondTo: RpcMessageHeader?): Int = createRPCMessage(fbb, messageType, messageOffset, respondTo?.txId()?.id())
```
ButterscotchV, line 397: I think this can be a little simpler by using `@JvmOverloads` with `txId: Long? = null`, then make `respondTo: RpcMessageHeader` non-nullable. I think this suggestion should work? GitHub UI makes it look weird tho.
```suggestion
	@JvmOverloads
	fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int, txId: Long? = null): Int {
val data = IntArray(1)

RpcMessageHeader.startRpcMessageHeader(fbb)
RpcMessageHeader.addMessage(fbb, messageOffset)
RpcMessageHeader.addMessageType(fbb, messageType)
		txId?.let { txId ->
			RpcMessageHeader.addTxId(fbb, TransactionId.createTransactionId(fbb, txId))
}
data[0] = RpcMessageHeader.endRpcMessageHeader(fbb)

val messages = MessageBundle.createRpcMsgsVector(fbb, data)

MessageBundle.startMessageBundle(fbb)
MessageBundle.addRpcMsgs(fbb, messages)
return MessageBundle.endMessageBundle(fbb)
}

	fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int, respondTo: RpcMessageHeader): Int = createRPCMessage(fbb, messageType, messageOffset, respondTo.txId().id())
```

ButterscotchV (changes requested):
Looks fine besides this overload weirdness, seems to be a continuation of #1527.

server/core/src/main/java/dev/slimevr/protocol/rpc/RPCHandler.kt
notpeelz: That's what I tried at first, but that causes the overload disambiguation to fail:

```
None of the following candidates is applicable:
fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int, txId: Long? = ...): Int
fun createRPCMessage(fbb: FlatBufferBuilder, messageType: Byte, messageOffset: Int, respondTo: RpcMessageHeader): Int
```

ButterscotchV: Hmm yeah I see the problem, this is annoying... The ambiguity when you put defaults really messes it up, so this is fine then.

ButterscotchV (approved):

ImSapphire (approved):

PR #1787, opened 2026-03-14
add tracker timeout setting

### LINKED ISSUE #1505: Freeze rotation when tracker disconnects
Currently if any of my trackers disconnect they just end up going to a default standing pose, which looks very jarring if for example my chest tracker disconnects for a few seconds while laying down (pretty common occurrence for me). I think it would would be way better if the server remembered the last rotation that the tracker was in and used that until it reconnects.

art0007i:
Resolves #1505 

SolarXR-Protocol pr: https://github.com/SlimeVR/SolarXR-Protocol/pull/201


server/core/src/main/java/dev/slimevr/config/TimeoutConfig.kt
```diff
@@ -0,0 +1,22 @@
...
+	var duration = 3.0f
+
+	fun updateTimeoutDuration() {
+		if (this.duration.isNaN() || this.duration.isInfinite()) {
+			this.duration = 3.0f
+		}
+		if (this.duration < 1f) {
+			this.duration = 1f
+		}
+		if (this.duration > 1000000000000000f) {
```
Eirenliel, line 17: why

art0007i: the idea was that it's a floating point setting which is being multiplied by 1000 and turned into a "Long", so I put an upper bound to prevent potential overflow

art0007i:
> This doesn't seem to actually implement the functionality.

It just modifies the default behavior which is 2 seconds of no data = tracker is timed out, 5 seconds of no data = tracker is disconnected.

I made the time when the tracker is timed out but not disconnected adjustable, so I can set it to like a million and my trackers will never show up as disconnected, only timed out, which is what was suggested in the issue.

server/core/src/main/java/dev/slimevr/config/TimeoutConfig.kt
Eirenliel: make it long then maybe...

art0007i: maybe, but if it was a long, it would be in ms, which I think typing values in milliseconds in the ui would be weird. 

I could make it so the server stores the number as a long, and the gui uses floats but performs all of the conversion before sending it, but from what I understand the solarxr stuff allows other apps to also listen to the same messages, in which case receiving a time in seconds as a float is 'nicer' than milliseconds as a long in my opinion

gui/public/i18n/en/translation.ftl
```diff
@@ -582,6 +582,8 @@ settings-general-tracker_mechanics-trackers_over_usb = Trackers over USB
 settings-general-tracker_mechanics-trackers_over_usb-description =
     Enables receiving HID tracker data over USB. Make sure connected trackers have <b>connection over HID</b> enabled!
 settings-general-tracker_mechanics-trackers_over_usb-enabled-label = Allow HID trackers to connect directly over USB
+settings-general-tracker_mechanics-timeout_duration = Timeout duration
+settings-general-tracker_mechanics-timeout_duration-description = How long in seconds should the tracker be considered timed out, instead of disconnected.
```
ImSapphire:
```suggestion
settings-general-tracker_mechanics-timeout_duration-description = How long until a tracker is marked as 'disconnected' when the connection is lost.
```

gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -726,6 +745,31 @@ export function GeneralSettings() {
...
+            <div className="grid gap-3">
+              <Input
+                type="number"
+                control={control}
+                name="timeout.duration"
+                label=""
+                min={1}
+                max={1000000000000000}
+                step="any"
+              />
```
ImSapphire, line 844: why not use a NumberSelector?

art0007i: I want to be able to type into it

art0007i:
what does "SolarXR PR found but no change on the submodule." mean on the failing check?
I just updated my repo so that the submodule matches the current one that's in SlimeVR-Server/main, which includes the merged commit from my SolarXR PR

sctanf:
It got merged so it is probably just confused now, since the merged pr is still linked on this pr

loucass003:
if no done already you need to point solarxr to main

loucass003:
this PR will have to be ported to #1798

art0007i:
> this PR will have to be ported to #1798

alright, does that mean I should close this PR and open a new one for the "server-rewrite" branch?

loucass003:
prob the easiest yeah

PR #1792, opened 2026-03-23
Introduce VID/PID allow lists

9il:
Alternatively, we can adopt another (universal) mechanism if it is defined and adopted for SlimeVR trackers.


sctanf:
the original reason enumeratedevices was so limited is because the enumeration was causing issues on random usb devices
i think only limiting by vendorId is too broad especially since we use the "open" vendorId that could be used by other devices, so this needs a little extra work to enumerate only our productIds when it is enumerating for our vendorId

loucass003 (approved):

PR #1798, opened 2026-03-27
Server rewrite

### LINKED ISSUE #895: Unify driver protobuf protocol to SolarXR
We kinda have 2 different schemas currently being used, this needs an overhaul.

- [ ] Figure out how the whole protocol should look in flatbuffer
- [ ] Use only unix sockets in all OS for the driver IPC (supported since [Insider Build 17063](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/) on Windows)
- [ ] Work out if any extra is needed for Monado to have an easy time using this

Stuff that needs to be thought of:
* Because we are gonna still use sockets/pipes, we should use TCP directly, it will make everything easier by making sure messages are in order (which should still happen in UDP tbh) and take care of the concept of a "connection" which UDP doesn't have. The current implementation in Linux is TCP, so it won't require much change.

### LINKED ISSUE #1543: Better visible telemetry for each tracker
This will increase quality of life of both developers and users:
1. Add "Signal Strength (dB)", "Noise Floor (dB)", "% Packet Loss" fields for each tracker in debug mode in GUI
2. Make debug data gathering time window configurable for GUI. For example, 2 second window will update debug values on screen every two seconds while showing average between these two seconds. Why is it needed? Sometimes we want to test slimes in specific scenarios and it's hard to compare numbers when they jump like crazy and fast. So it's nice to have some form of averaging. Also it limits GUI updates.
3. (optional) Allow to show graphs of such data. Update graphs and current values with the same time specified time window. See also https://github.com/SlimeVR/SlimeVR-Server/issues/1531
4. Make GUI in debug mode also show magnetometer data, if it's possible

loucass003:
# Full server rewite

Rewrite everything with kotlin (multiplatform) and coroutines.
Unit test the crap of the new server
Requires: SlimeVR/SolarXR-Protocol#217

# Server features to implement

### Config
- [x] Basic module to interface with config
- [x] Config profiles (user and settings)
- [x] Profile switching
- [x] Auto save

### Device
- [x] Basic module / Holding the basic state
- [x] Calculate tps based on updates

### Tracker
- [x] Basic module / Holding the basic state
- [x] Need to define how the to categorize trackers
- [x] Position data is set by feeder and driver
	- We have the source, but no stuff like are they trackers computed for steamvr or vmc
- [x] User can:
  - [x] Assign a body part
  - [x] Set the custom name of a tracker
  - [x] Enable mag

### Firmware Tool
- [x] Firmware tool can flash trackers through serial and ota
- [x] Unit testing

### Serial
- [x] Scan for new ports
- [x] open connection for serial
- [x] open connection for flashing
- [x] support multiple connections at the same time
- [x] Platform support:
  - [x] Windows
  - [x] Linux
  - [ ] MacOS
  - [x] Android
- [x] Unit testing

### Height Calibration
- [x] Go through all the steps of the calibration
- [x] Can give the user height
- [x] Can give the user height to scaled proportions
- [x] Unit testing
	- [x] Basic testing without actual data

### HID (usb dongles)
- [x] Can detect dongles
- [x] Create device and trackers
- [ ] Handles all the HID packets
- [x] Calculate packet loss
- [x] Platform support:
  - [x] Windows
  - [x] Linux
  - [ ] MacOS
  - [x] Android

### UDP Trackers
- [x] Can parse all UDP packets
- [x] Create device and trackers
- [x] Handles reconnection / timeouts
- [x] Handle out of order packets
- [x] Calculate ping
- [x] Calculate packet loss

### Wi-Fi Provisioning
- [x] Can connect a tracker to the wifi
- [x] Handle all error cases
- [x] Handle retries
- [x] Handle detection of turned off trackers
- [x] Handle detection of unplugged trackers
- [x] Unit testing

### Skeleton
- [x] Need to figure out a data structure

### Solarxr
- [x] Can handle solarxr messages
- [x] WS support
- [x] File socket / named pipes support
- [x] Platform Support:
  - [x] Windows
  - [x] Linux
  - [ ] MacOS
- [x] Can initialize datafeeds
- [x] Can send RPC packets and listen to them
- [ ] Unit testing

### VRChat Config Warnings
- [x] Send vrc config warning status to solarxr
- [x] Can verify vrc data against recommended values
- [x] Handles reading config from:
  - [x] Windows
  - [x] Linux
  - [ ] MacOS

### Tracking Checklist
- [x] Can mute steps and save to config
- [x] Handle all these steps
	- [x] Network profile detection
	- [x] Steamvr disconnected
	- [x] Tracker error
	- [x] Rest calibration
	- [x] Full Reset
	- [x] Mounting calibration
	- [x] Feet mounting
	- [x] Unassigned HMD
	- [x] Stay Aligned
	- [x] VRC settings

### Scaled Proportions
- [x] Can calculate body proportions based on user height

### OSC Router
- [x] Nuked

### VRC OSC
- [x] Send tracking data to VRC
- [ ] Flightlist to enable "Allow Sending Head and Wrist VR Tracking OSC Data"

### VMC
- [x] Send data to vmc apps
- [x] Handle VRM models
- [x] Anchor hip
- [x] Tracking mirroring

### BVH
- [x] Can record data
- [x] Can save to a file
- [x] Can create a file without prompt if folder is defined

### Server Guards
Handle the following guards:
- [x] Can do mounting
- [x] Can do yaw reset
- [x] Can do height calibration
- [x] Reset can do mounting after 2min since last full reset

### Autobone
- [ ] TBD

### Manual Proportions
- [x] User can set body proportion values
- [x] Skeleton updates accordingly
- [x] Body proportions are saved to user config

### Taps
- [x] Taps are detected from trackers
- [x] Taps trigger the corresponding resets
- [x] Taps amounts / delay / enabled can be configured
- [x] Unit testing

### Stay Aligned
- [x] can record rest poses
- [x] Can correct skeleton based on rest poses
- [x] Adjust trackers and Hide adjustments toggles

### Keybinds
- [x] Can read key inputs:
  - [x] Windows
  - [x] Linux
  - [ ] MacOS
- [x] Trigger the corresponding resets
- [x] Handle GUI assignment

### Driver
- [x] Basic reporting of trackers to the driver
- [x] Platform Support:
  - [x] Windows
  - [x] Linux
  - [ ] MacOS

### Feeder
- [x] Collect tracker data and create devices / trackers to the server
- [x] Platform Support:
  - [ ] Windows
  - [x] Linux
  - [ ] MacOS

### SteamVR
- [x] Enabled trackers

### Leg Tweaks
- [ ] TBD

### Skating correction
- [ ] TBD

### Filtering
- [x] Smoothing
- [x] Prediction

### Skeleton Ratios
- [x] Rework to have less of them

### Cleanup
- [x] Delete Skeleton toggles from GUI
- [x] Delete OSC Router from GUI

## Small touchups
- [x] Make server only assign one body part to one tracker only
- [x] Make sure java version check works on electon. (java 17 -> 25)
- [ ] Bring back option to have a local config folder


server/core/src/main/java/dev/slimevr/tracker/module.kt
```diff
@@ -44,9 +44,9 @@ sealed interface TrackerActions {
 	data class SetMagStatus(val status: MagnetometerStatus) : TrackerActions
 	data class SetStatus(val status: TrackerStatus) : TrackerActions
 	data class SetRotation(val rotation: Quaternion? = null, val acceleration: Vector3? = null) : TrackerActions
-	data class FullReset(val referenceRotation: Quaternion) : TrackerActions
-	data class YawReset(val referenceRotation: Quaternion) : TrackerActions
-	data class MountingReset(val referenceRotation: Quaternion) : TrackerActions
+	data class FullReset(val referenceRotation: Quaternion, val resetHmdPitch: Boolean) : TrackerActions
+	data class YawReset(val referenceRotation: Quaternion, val yawResetSmoothTime: Float) : TrackerActions
```
loucass003: why is it here? is smooth time part of the tracker state?
same question with save mointing reset. i dont think they have any business being there?

Erimelowo: Right, I forgot actions were only for state ;-;
I'll remove them for now

server/core/src/main/java/dev/slimevr/resets/behaviours.kt
```diff
@@ -0,0 +1,45 @@
...
+}
+
+class ResetsMountingTimeoutBehaviour : ResetsBehaviour {
+	@OptIn(ExperimentalCoroutinesApi::class)
+	override fun observe(receiver: ResetsManager) {
+		receiver.context.state
+			.distinctUntilChangedBy { it.lastFullResetTime }
+			.mapLatest {
+				delay(MOUNTING_RESET_TIMEOUT)
+				receiver.context.dispatch(ResetsActions.ClearResets(listOf(ResetType.Mounting)))
```
loucass003: are you sure the delay gets canceled if called again before the timeout?

Erimelowo: Yes - that's why I use mapLatest instead of onEach

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-description =
+    Choose what each output receives.
+    SteamVR/Monado, VRChat OSC and VMC can each get a different set of body parts.
+settings-routing-mode = Routing mode
+settings-routing-automatic-label = Automatic routing
+settings-routing-mode-description =
+    While Automatic is on, SlimeVR picks the best output for every bone so nothing is sent twice: SteamVR/Monado is preferred, VRChat OSC covers the bones it cannot take, and VMC always gets the full skeleton.
+    Turn it off to decide where every bone goes yourself, including sending one bone to several outputs at once.
+settings-routing-bone = Bone
+settings-routing-cell-required = This output needs this bone to work, so it is always sent.
```
ImSapphire: Could probably just say "Mandatory for this output"?

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-cell-unavailable = This output is not available on this platform.
```
ImSapphire: "Output is unavailable on this platform"

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-cell-duplicate = This bone reaches the same app through another output too, so it arrives twice.
+settings-routing-duplicate-warning = { $bones } go to { $outputs } at the same time. Both might end up in the same app or game, which would see two trackers for each.
```
ImSapphire: Looks awkward when only one bone is in the bones list, i.e. "Upper chest go to [outputs] at the same time"
the wording also feels weird, I feel like it could be "[bones] is/are routed to multiple outputs ([outputs]), which may show duplicate trackers in applications"
ideally it should say 'is' if only one bone is in the list, otherwise 'are'

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-outputs = Outputs
+settings-routing-bones = Bones
+settings-routing-bones-description = A tick means the bone is sent to that output. A dash means the output does not support that bone. Turn off Automatic to pick them yourself.
+settings-routing-output-driver = SteamVR / Monado
+settings-routing-output-vrc_osc = VRChat OSC
+settings-routing-output-vmc = VMC
+settings-routing-output-bone-count = { $routed } of { $accepts } bones routed
+settings-routing-output-badge-sending = Sending
+settings-routing-output-sending-description = Connected and receiving the bones below.
```
ImSapphire: having a description feels unnecessary, the 'Sending' badge says the same thing

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-output-badge-stopped = Not running
+settings-routing-output-stopped-description = No SteamVR or Monado driver is connected, so nothing routed to it is sent.
```
ImSapphire: same here, description feels superfluous?

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-output-badge-idle = Idle
+settings-routing-output-idle-description = Turned on, but it has not found anything to send to yet, so those bones are not going out.
```
ImSapphire: "are not going out" sounds weird to me. Maybe "are not being sent"?

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-output-idle-automatic-description = Turned on, but it has not found anything to send to yet, so its bones go to another output until it does.
```
ImSapphire: "so its bones go to another output until it does." -> "so its bones are rerouted to another output in the meantime"

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-output-badge-off = Off
+settings-routing-output-off-description = Turned off in its own settings, so nothing routed to it is sent.
```
ImSapphire: The description feels superfluous, we already have the badge that says it's off

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
+settings-routing-output-badge-empty = Nothing routed
+settings-routing-output-empty-description = No bone is routed to it, so it has nothing to send.
+settings-routing-output-badge-unavailable = Unavailable
+settings-routing-output-unavailable-description = This output does not exist on this platform.
```
ImSapphire: Maybe we should just hide those outputs completely?

gui/app/public/i18n/en/translation.ftl
```diff
@@ -476,36 +478,60 @@ settings-sidebar-firmware-tool = DIY Firmware Tool
...
-    This should only be enabled if you are using SlimeVR for hand tracking.
-
-    Are you sure you want to do this?
-settings-general-steamvr-trackers-hands-warning-cancel = Cancel
-settings-general-steamvr-trackers-hands-warning-done = Yes
+## Bone routing settings
+settings-routing = Bone Routing
+settings-routing-description =
+    Choose what each output receives.
+    SteamVR/Monado, VRChat OSC and VMC can each get a different set of body parts.
```
ImSapphire, line 515: I feel like this should just say "Each output can send a different set of body parts."

server/desktop/src/main/java/dev/slimevr/desktop/ipc/solarxr.kt
```diff
@@ -55,6 +59,18 @@ class OpenKeybindSettingsBehaviour : SolarXRBridgeBehaviour {
...
 }
 
+class OpenBindingsProviderBehaviour(val scope: CoroutineScope) : SolarXRBridgeBehaviour {
+	override fun reduce(
+		state: SolarXRBridgeState,
+		action: SolarXRBridgeActions,
+	): SolarXRBridgeState {
+		if (action is SolarXRBridgeActions.SetDriverInfo && action.name == "SteamVR") {
+			scope.launch { startBindingsProvider() }
+		}
```
loucass003: Nope thats, illegal. dont do that. only pure data in reducers please

server/core/src/main/java/dev/slimevr/hid/packets-legacy.kt
```diff
@@ -0,0 +1,220 @@
...
+	val vz = ((buf shr 21 and 2047u).toFloat() / 2048f) * 2f - 1f
+	val d = vx * vx + vy * vy + vz * vz
+	val invSqrtD = 1f / sqrt(d + 1e-6f)
+	val a = (PI.toFloat() / 2f) * d * invSqrtD
+	val s = sin(a)
+	val k = s * invSqrtD
+	return AXES_OFFSET * Quaternion(cos(a), k * vx, k * vy, k * vz)
+}
+
+private fun decodeBattery(raw: Int): Float? = if (raw == 128) null else (raw and 127).toFloat() / 100f
```
ImSapphire, line 210: Isn't this wrong when no battery level is available? I believe the check should be `if (raw == 0) null`?

PR #1802, opened 2026-04-01
Hannah/keybinds

HannahPadd:
Added page in settings for editing keybinds
Added support in solarXR for sending keybind data to the server and gui.
SolarXR PR: [SlimeVR/SolarXR-Protocol#203](https://github.com/SlimeVR/SolarXR-Protocol/pull/203)


loucass003 (changes requested):
Oof the dbus lib / keybind interfacing need to be moved to the desktop module

PR #1806, opened 2026-04-08
Deactivate feeder bridge when using new driver, add bindings provider

ImSapphire:
The bindings provider purely exists to facilitate binding commonly-used actions (such as yaw/full reset, (feet) mounting calibration, and toggle tracking) to controller buttons, similar to how the current feeder app does it.

The bindings provider uses Flatbuffers to send commands to the server. It will be autostarted by the server when a new enough driver which handles device feeding connects to the server, and it will deactivate the feeder bridge, meaning the existing feeder app, if still being started, should not interfere with anything.

When this is merged, there should probably be a PR prepared to the installer that will remove the feeder app.

Requires: SlimeVR/SlimeVR-OpenVR-Driver#69 (merged)
Requires: SlimeVR/SolarXR-Protocol#209 (merged)

TODO: 
- [x] Actually build the bindings provider in CI and bundle it when packaging.
- [x] Maybe rename the bindings-provider folder? It seems silly to have `svr-` at the beginning.
- [x] Rename the executable file to 'SlimeVR-BindingsProvider[.exe]'
- [ ] Do more testing and sanity checks on the new Windows named pipe implementation (for driver, feeder, and RPC)
- [x] Try to fix the stupid error when loading action manifests (SteamVR bug??)


gui/electron-builder.yml
```diff
@@ -35,6 +35,8 @@ linux:
   extraFiles:
     - from: "../server/desktop/build/libs/slimevr.jar"
       to: "."
+    - from: "../bindings-provider/build/slimevr-bindings-provider"
```
loucass003: i think these are only for linux and mac. the list. 
not 100% sure

ImSapphire: This list here is only for Linux, Mac has a different target below (which does not contain bindings-provider as SteamVR is unsupported for macOS)

PR #1808, opened 2026-04-13
Fix NPE error caused by the second HID endpoint of HID composite devices

KenRouKoro:
This PR fixes a NullPointerException in the hid4java data reader and device enumerator threads that occurs when a single USB device enumerates multiple HID interfaces (with identical VID/PID/serial numbers) due to the physical device having a second HID endpoint, which previously prevented the receiver's tracking data from being recognized.

Additionally, it resolves a secondary issue introduced by the initial fix: if the second HID endpoint is **not** a SlimeVR HID endpoint, it causes a loop of `[TrackerServer] Reopening device ${device.serialNumber} after no data received`. This happens because the system attempts to handshake with the second HID endpoint as if it were a SlimeVR device, despite there being no corresponding return data. To address this, I have introduced a `keepFirstHidPerPhysicalDevice` filter to block extra HID endpoints.

### DIFF
--- a/server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/DesktopHIDManager.kt
+++ b/server/desktop/src/main/java/dev/slimevr/desktop/tracking/trackers/hid/DesktopHIDManager.kt
@@ -71,6 +71,7 @@ class DesktopHIDManager(name: String, private val trackersConsumer: Consumer<Tra
 			// val manufacturer = hidDevice.manufacturer
 			this.devicesBySerial[serial]?.let {
 				this.devicesByHID[hidDevice] = it
+				this.lastDataByHID[hidDevice] = this.lastDataByHID.getOrDefault(hidDevice, 0)
 				synchronized(this.devices) {
 					for (id in it) {
 						val device = this.devices[id]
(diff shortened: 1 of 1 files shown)


loucass003:
@sctanf can you double check that one :pray:

sctanf:
The logic was a little hard to follow, but it seems to be okay.
At line 265: I think it could be changed to just remove devices not in `keptSet`, so making the extra sets `enumeratedSet` and `droppedSet` is redundant and probably will make it clearer to understand. Assuming the rest was done for readability which is okay I think

sctanf:
I am not aware of a use case to have multiple HID interfaces, is this for functionality outside of SlimeVR?

KenRouKoro:
> I am not aware of a use case to have multiple HID interfaces, is this for functionality outside of SlimeVR?

Sorry for the late reply. The extra HID issue I encountered is because I designed a receiver integrated with the Dock, which adds an interface dedicated to desktop software control (mainly used for physical button controls on the dock, as well as invoking commands and obtaining specific debug data, such as retrieving the native sensor data of the tracker. This approach is much more stable than using CDC serial ports, which require filtering too many things).
Line 265 was indeed intended for readability.

9il:
> I am not aware of a use case to have multiple HID interfaces, is this for functionality outside of SlimeVR?

We also plan to use a few HID interfaces: one for Slime VR and other for common peripherals (game controller and etc).

loucass003:
Need conflicts resolved before merge

PR #1816, opened 2026-04-18
v20.0-rc.1 fixes

ImSapphire:
Fixes built packages missing some modules, causing errors on startup and when attempting to print log messages.

### DIFF
--- a/gui/src/components/settings/pages/GeneralSettings.tsx
+++ b/gui/src/components/settings/pages/GeneralSettings.tsx
@@ -1,5 +1,5 @@
 import { Localized, useLocalization } from '@fluent/react';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { DefaultValues, useForm } from 'react-hook-form';
 import {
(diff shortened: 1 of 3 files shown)


gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -179,6 +179,10 @@ export function GeneralSettings() {
   const { l10n } = useLocalization();
   const { config } = useConfig();
   const { currentLocales } = useLocaleConfig();
+  // null = we should show the hands warning on next hand tracker toggle.
+  // false = means the warning should not be shown right now
+  // true = warning is being shown
+  const handsWarning = useRef<boolean | null>(null);
```
loucass003: Sadly it is very bad practice to use a ref in the reactive part of the dom. As it breaks reactivity. 
The only thing that saves you is that the component refresh a lot currently. But this is not a guarantee that changing a ref will trigger a render

loucass003 (changes requested):

loucass003 (approved):

PR #1819, opened 2026-04-20
fix videos

loucass003:


flake.nix
```diff
@@ -39,6 +45,14 @@
...
               export USE_SYSTEM_FPM=true
               export USE_SYSTEM_MKSQUASHFS=true
+
+              # Prevent electron from re-downloading itself
+              export ELECTRON_SKIP_BINARY_DOWNLOAD=1
+
+              # Wayland support for Electron (mirrors nixpkgs slimevr package)
+              if [ -n "$NIXOS_OZONE_WL" ] && [ -n "$WAYLAND_DISPLAY" ]; then
+                export ELECTRON_OZONE_PLATFORM_HINT=auto
+              fi
```
ImSapphire: Does this env var still exist? According to https://github.com/electron/electron/issues/48001 the variable has been removed in newer Electron and the default is now `auto`

loucass003: i updated based on https://github.com/NixOS/nixpkgs/pull/511205/changes
not sure

loucass003: i figured the nix people knew better

ImSapphire: I think it's just a generic copy-paste, but the snippet from that PR also sets some other flags. We can't set that in an env var as far as I'm aware though.

ButterscotchV (approved):

ImSapphire (approved):

PR #1822, opened 2026-04-21
Prevent new windows with Ctrl+Click on links

loucass003:


ImSapphire (approved):
It would probably be nice if we could force Ctrl+Click to just go to the page in the current window, instead of completely inhibiting it

PR #1825, opened 2026-04-29
Add extra SteamVR driver sanity checks + Standable step to flightlist

ImSapphire:
~~Needs to be tested on Windows, I've only tested this on Linux. I think it should work as expected though.~~
Tested on Windows and Linux.


server/core/src/main/java/dev/slimevr/trackingchecklist/TrackingChecklistManager.kt
```diff
@@ -283,15 +332,74 @@ class TrackingChecklistManager(private val vrServer: VRServer) : VRCConfigListen
 				steamvrConnected,
 			) {
 				it.enabled = true
-				if (!steamvrConnected) {
-					it.extraData = TrackingChecklistExtraDataUnion().apply {
+				it.extraData = if (!steamvrConnected) {
```
loucass003, line 325: can you extract this logic somewhere else. i really dont like how big this is

server/core/src/main/java/dev/slimevr/protocol/rpc/RPCHandler.kt
```diff
@@ -610,6 +629,47 @@ class RPCHandler(private val api: ProtocolAPI) : ProtocolHandler<RpcMessageHeade
 		sendSettingsChangedResponse(conn, messageHeader)
 	}
 
+	private fun onEnableSteamVRDriverRequest(conn: GenericConnection, messageHeader: RpcMessageHeader) {
+		val request = messageHeader.message(EnableSteamVRDriverRequest()) as? EnableSteamVRDriverRequest ?: return
+
+		val serverUrl = "http://127.0.0.1:27062"
```
loucass003: i would prefer this to be in a const and reused

loucass003 (changes requested):

gui/src/components/tracking-checklist/TrackingChecklist.tsx
```diff
@@ -157,6 +224,23 @@ function SteamVRHandsEnabled() {
   );
 }
 
+function StandableInstalled() {
+  const { l10n } = useLocalization();
+
+  return (
+    <>
```
loucass003: No need if you only have one child

loucass003 (changes requested):

PR #1826, opened 2026-04-29
Don't send VRChat OSC data when the SteamVR driver is connected

ImSapphire:


Eirenliel:
why though?

ImSapphire:
> why though?

People frequently enable VRCOSC and forget about it, then when they open VRChat they have doubled-up trackers in-game, which can cause weird tracking issues, and makes it harder to tell what is going on during troubleshooting. It's also just not really necessary to be sending both.

loucass003:
isnt that gonna break toe tracking tho?

ImSapphire:
I forgot about toe tracking since it's not in main. I suppose in that case, instead of completely disabling OSC, we could just not send the normal tracker data via OSC while the SteamVR bridge is connected, while still sending toes. Would that be preferred?

loucass003:
my plans for this was te be able to do a priority list for trackers

so you could say for this tracker i whant this order of backends based on what is available

loucass003:
I think its probably better to close this until the end of the rewrite. Stuff are gonna change around that. Not worth it to agrue on what to do yet

Issue #142, opened 2022-03-27
Properly support positional constraints in skeletal model


kitlith:
*(aka: "Let me mix slimevr knees with vive tracker hips/feet damnit")*

@Louka3000 already got elbows working (:heart:) by using the controllers as a new "root", so they're in a working state. Some more enhanced calibration might be nice for it, but as long as you don't need shoulder tracking (looking at VRChat) it seems to work nicely as-is without touching calibration at all (from my limited experience using joycons)

The main thing currently is that vive trackers don't integrate well into the main skeleton. They apparently don't use their position, only maintain their rotation, and somehow affect the rotation of child trackers? (is there some code specifically for vive trackers already that needs changing?)

Short term, if we can fix whatever weird thing is going on with rotation, override the position of the hip/waist using the vive tracker's actual position, then add parameters to offset the tracker into the right place (especially for people using a tracker on their side, instead of in front) then we'll be gold. I don't know much about server internals right now, but this should require minimal touching of the internal skeleton.

An interesting option is to have the option of performing a repeat of elbow tracking -- use foot trackers as a new root, then do the knees based off of them. BVH is a big concern here, though.

I have some thoughts about root bones in general that I've mentioned in discord, but a vauge idea that might require redoing the way the skeleton works is not worth mentioning in this particular issue. tbh, I'm mostly transcribing a discussion from discord into somewhere perhaps slightly more permenant/visible/referencable.

Louka3000:
![](https://github.trello.services/images/mini-trello-icon.png) [HMD pos when using base stations is wrong](https://trello.com/c/bjJNDeS2/29-hmd-pos-when-using-base-stations-is-wrong)

TheButlah:
one thing that would be helpful is a brief explanation of how it works *currently*. @Louka3000 could you elaborate if you have some time?

TheButlah:
As an aside, positional constraints were one of the motivating factors with the implementation of the rust skeletal model. Feel free to reference that preliminary work if its helpful. However, the approach taken there is more advanced than probably what would be used in the java model.

If you *are* curious, I've taken extensive effort to document the algorithm in plain English [here](https://slimevr.github.io/SlimeVR-Rust/skeletal_model/index.html#documentation-on-internals)

Louka3000:
> one thing that would be helpful is a brief explanation of how it works _currently_. @Louka3000 could you elaborate if you have some time?

[...]

PR #1827, opened 2026-04-29
Fix support for external trackers via VMC

Seneral:
This PR intends to fix the existing support for external tracking via VMC.

Known Issues:
- OSCRouter spams "Packet is too large for the buffer in use"
    - Buffer is just randomly small, data is NOT unexpectedly large
    - No apparent reason why buffer is downsized, might be data race
    - WONT FIX: JavaOSC is slated to be replaced with a simpler reimplementation, no point fixing every last issue with it rn
- OSC sometimes fails to listen when restarted due to refreshSettings
    - Causes OSC to stop working randomly, even when not in settings
    - WONT FIX: Hopefully addressed by osc reimplementation
    
Requires https://github.com/SlimeVR/SolarXR-Protocol/pull/213


Seneral:
The first few commits are in PR #1840 
After that, this is ready for review, too

Seneral:
Alright this should be just about everything needed for basic external tracker support via VMC.

There are a few things I'd change about the UI still:
I added a rescale tracking toggle to both OSC Router and VMC settings.
OSC Router really needs it, but it requires a VRM to be loaded in VMC settings. This is not obvious to the user.
The toggle in VMC is perhaps redundant, turning it off matches current behaviour without a VRM supplied at all.
I also don't like that it's not very obvious by just looking at the settings page why and when the VRM is needed.
I'd like to change it or associate it closely with that toggle somehow. Input appreciated.

Since IK Solver is now properly integrated (#142), tracking arms from controllers should now be deprecated IMO.
It's inferior in every way except if you absolutely need the hand position to match up, even if the IK solver determines the arm reach is insufficient.
This would also mean adding a warning to anybody that has "Force Arms From HMD" set to off - it should be on, the skeletal hierarchy should be one tree with head as the root, and IK solver resolving all positional constraints within (including hands).

The IK Solver should be able to handle trackers on any body part, though not optimally yet:
- Two serial IK chains (e.g. head-hip-feet) would not optimally converge, can likely be easily addressed once I have those trackers to test with
- Overconstrained scenarios (e.g. two slime arm trackers + positional hand tracker) may be unstable due to fixed lerp factor

Ideally, there'd be configuration in the GUI whether a given tracker needs positional calibration in IKConstraints.kt or not. Currently, an HMD on the head and hand-trackers are assumed to be aligned properly, all other trackers have a rudimentary offset calibration on reset (untested by me).

Finally, I'd consider not squashing some of these commits as they're quite varied and tracing back the reasoning for each of them would be painful without the context of the messages.
I'll open this for review after I address at least the remaining GUI issues.

gui/src/components/settings/pages/GeneralSettings.tsx
```diff
@@ -924,6 +925,9 @@ export function GeneralSettings() {
                 )}
               />
             </div>
+            {!watch('toggles.forceArmsFromHmd') && (
```
ImSapphire: Add this to the watch on L232 instead

ImSapphire (commented):
Also please run `pnpm run lint:fix` and rebase to main, CI is failing and some indentation is messed up

gui/src/components/settings/pages/GeneralSettings.tsx
Seneral: Done, same in VMCSettings, any watch-ed field is now a const

Seneral:
Alright, all commits are now properly formatted. Sorry I missed most of them because it didn't fix formatting when there were warnings of unused imports & variables.

gui/public/i18n/en/translation.ftl
```diff
@@ -822,6 +823,9 @@ settings-osc-router-network-port_out =
 settings-osc-router-network-address = Network address
 settings-osc-router-network-address-description = Set the address to send out data at.
 settings-osc-router-network-address-placeholder = IPV4 address
+settings-osc-router-rescale_tracking = Rescale Tracking To Avatar Scale
+settings-osc-router-rescale_tracking-description = Rescale external tracking routed through OSC to avatar scale as determined by the VRM loaded in VMC settings.
+settings-osc-router-rescale_tracking-label = Rescale Tracking
```
ImSapphire:
```suggestion
settings-osc-router-external_tracking = External tracking
settings-osc-router-rescale_tracking = Rescale to avatar
settings-osc-router-rescale_tracking-description = Rescale external tracking routed through OSC to avatar scale as determined by the VRM loaded in VMC settings.
```

gui/public/i18n/en/translation.ftl
```diff
@@ -879,12 +883,13 @@ settings-osc-vmc-network-port_out =
 settings-osc-vmc-network-address = Network address
 settings-osc-vmc-network-address-description = Choose which address to send out data at via VMC.
 settings-osc-vmc-network-address-placeholder = IPV4 address
-settings-osc-vmc-vrm = VRM Model
-settings-osc-vmc-vrm-description = Load a VRM model to allow head anchor and enable a higher compatibility with other applications.
+settings-osc-vmc-vrm = VRM Avatar
+settings-osc-vmc-vrm-description = Load a VRM avatar to allow for positioning using that avatars scale. Otherwise, alignment in external programs that use VMC might be wrong.
 settings-osc-vmc-vrm-untitled_model = Untitled model
 settings-osc-vmc-vrm-file_select = Drag & drop a model to use, or <u>browse</u>
+settings-osc-vmc-vrm-required = VRM Avatar is required for proper avatar positioning over VMC!
```
ImSapphire:
```suggestion
settings-osc-vmc-vrm-required = Loading a VRM avatar is required for proper avatar positioning over VMC!
```

gui/src/components/settings/pages/VMCSettings.tsx
```diff
@@ -106,22 +107,32 @@ export function VMCFileUpload() {
...
-        // if modelname is an empty string, it's an untitled model
-        modelName === ''
-          ? l10n.getString('settings-osc-vmc-vrm-untitled_model')
-          : modelName
-      }
-      label="settings-osc-vmc-vrm-file_select"
-      accept="model/gltf-binary, model/gltf+json, model/vrml, .vrm, .glb, .gltf"
-    />
+    <>
+      {modelName === null && required && (
```
ImSapphire: Doesn't seem like `required` is used?

gui/src/components/settings/pages/VMCSettings.tsx
```diff
@@ -106,22 +107,32 @@ export function VMCFileUpload() {
...
+        <WarningBox>
+          {l10n.getString('settings-osc-vmc-vrm-required')}
+        </WarningBox>
```
ImSapphire:
```suggestion
        <Localized id="settings-osc-vmc-vrm-required">
          <WarningBox>Warning</WarningBox>
        </Localized>
```

gui/src/components/settings/pages/VMCSettings.tsx
```diff
@@ -106,22 +107,32 @@ export function VMCFileUpload() {
...
+      )}
+      {modelName === null && !required && suggested && (
+        <TipBox>{l10n.getString('settings-osc-vmc-vrm-required')}</TipBox>
```
ImSapphire:
```suggestion
        <Localized id="settings-osc-vmc-vrm-required">
          <TipBox>Tip</TipBox>
        </Localized>
```

gui/src/components/widgets/SkeletonVisualizerWidget.tsx
```diff
@@ -85,11 +86,10 @@ function initializePreview(
   scene.add(skeleton[0]);
 
   let heightOffset = 0;
-  let skeletonOffset = 0;
 
   const rebuildSkeleton = (
     newSkeleton: (BoneKind | Bone)[],
-    bones: Map<BodyPart, BoneT>
+    _bones: Map<BodyPart, BoneT>
```
ImSapphire: if the parameter is unused, perhaps remove it

gui/src/components/settings/pages/VMCSettings.tsx
Seneral: Intended it as an easy switch to warning if the VRM is ever a hard requirement anywhere. I'll remove it

loucass003:
what is the status of this? with the router possibly gone with the rewrite. Some changes are gonna be needed i think?

Seneral:
> what is the status of this? with the router possibly gone with the rewrite. Some changes are gonna be needed i think?

We had some [talks on the discord](https://discord.com/channels/817184208525983775/948579740225261598/1513898724290465905) with the result to merge a cut-down version of the router (~10 lines) into the normal VMC operation, without the complexity of configuring different ports.
In other words, I can easily add the "Forward all OSC packets" and "Rescale External Tracking" options (router and this PR respectively) back in the new server, with little rework. With the intention not to be a general router like before, but strictly to take tracking input from VMC, and output the skeleton over VMC and if desired, forward the tracking input to the output (with optional rescale).

So yes I'd just need a timeline for the rewrite, from which point it makes sense to redo this PR on that.
A lot of the work done here is integration work, and I'm not sure how much of that is going to apply to the new server, but I don't think it'll be nearly as much effort as the first time.

loucass003:
Rewrite should be in a ok state for it even rn. So if you feel like it we can spend time together figuring this out if you need help with the new codebase.

Seneral:
> Rewrite should be in a ok state for it even rn. So if you feel like it we can spend time together figuring this out if you need help with the new codebase.

Ok, thanks for letting me know!
I'm currently busy with multiple things (FW for new HW, preparing demo & video) so not sure when exactly I'll get to it, but I'll at least look which parts should transfer cleanly and which require larger adaptation.

loucass003:
@Seneral with all the changes to rewrite. i would close this pr. is that ok for you?

Seneral:
> @Seneral with all the changes to rewrite. i would close this pr. is that ok for you?

Yep, looking to find a good time to rebase it. If you could give me a strong signal that the rewrite is in a good, near-final state, that'd help a lot. Doesn't help I'm likely to be even more busy soon, so maybe now is the time?

loucass003:
> > @Seneral with all the changes to rewrite. i would close this pr. is that ok for you?
> 
> Yep, looking to find a good time to rebase it. If you could give me a strong signal that the rewrite is in a good, near-final state, that'd help a lot. Doesn't help I'm likely to be even more busy soon, so maybe now is the time?

Our planning puts end of coding in 2 weeks. IK should be merged during next week. we are very close yes. almost everything is working and is gonna enter testing phase soon

Erimelowo:
You won't be able to rebase it considering that the whole kotlin backend got rewriten 100% from scratch. Please tell us if/when you start rewriting this so we don't step over eachothers toes since I was planning to reimplement receiving VMC trackers and bones properly in rewrite anyways but help is very welcome!

Also please avoid unrelated changes like changing arms from controller. For rewrite, we're probably nuking that option anyways once we can test that it works fine with IK alone.

And OSC Router shouldn't have been touched at all in this PR, but it already got nuked in rewrite so not relevant either.

Seneral:
Sounds good!

> Also please avoid unrelated changes like changing arms from controller. For rewrite, we're probably nuking that option anyways once we can test that it works fine with IK alone.

Not really unrelated, the switch to IK is actually needed for proper external tracker handling and part of the PR (075f965). I just kept the separated controller skeletal tree option as legacy, removing it sounds like a good move. 
Also looking forward to IK solver being validated, was considering doing something like that to ensure the resulting poses are close to the external trackers without much additional jitter. Generally, checking the error of any solution to all sensor inputs (be they positional, rotational or both) sounds like a good idea.

Overall this PR is mostly small bits and pieces of integration work and fixes, some will be easy to move over, some will not apply any longer, and integration testing will likely bring up some new changes to be made.

PR #1829, opened 2026-05-02
Fix settings scroll target positioning

### LINKED ISSUE #1764: Bug: "click here to go to the setting" link in an individual tracker magnetometer settings goes to wrong settings point.
Clicking the the "click here to go to the setting" link for magnetometers in the tracker page sends you to top top of the 'Tracker settings' page rather than the mag setting itself at the bottom of the 'Tracker settings' page.

Image

Image

Could be slightly confusing for a small slice of new users.

aukaukauk:
## Summary
- Fix settings-page scrolling to nested targets.
- Calculate the target position relative to the scroll container instead of relying on `offsetTop`.

## Why
The magnetometer link from a tracker settings page navigates to `/settings/trackers` with `scrollTo: "mechanics-magnetometer"`. That target is nested inside the settings page, so `offsetTop` can be relative to a nearer offset parent and scroll to the wrong place.

## Testing
- Ran `pnpm --filter slimevr --dir gui lint`.
- Started the server and GUI locally with a fake UDP tracker that supports magnetometer.
- Opened the fake tracker's settings page, clicked `click here to go to the setting`, and confirmed it scrolls to the global magnetometer setting.

## AI disclosure
AI-assisted for debugging and PR text, tested locally.

Fixes #1764

### DIFF
--- a/gui/src/components/settings/SettingsPageLayout.tsx
+++ b/gui/src/components/settings/SettingsPageLayout.tsx
@@ -21,24 +21,20 @@ export function SettingsPageLayout({
       `#${typedState.scrollTo}`
     ) as HTMLElement | null;
     if (elem) {
-      // stupid way of doing this, just get the closest overflow-y-auto
-      // usually its just the parentElem
-      const closestScroll = elem.closest(
-        '.overflow-y-auto'
-      ) as HTMLElement | null;
-      if (closestScroll) {
-        elem.scrollIntoView({
-          block: 'start',
-          behavior: 'smooth',
-        });
-      }
+      elem.scrollIntoView({
+        block: 'start',
+        behavior: 'smooth',
+      });
     }
   }, [state]);
 
   return (
     <div
       ref={pageRef}
-      className={classNames('settings-page-layout', className)}
+      className={classNames(
+        '[&_[id]]:scroll-mt-[45px] mobile:[&_[id]]:scroll-mt-[80px]',
+        className
+      )}
       {...props}
     >
       {children}
--- a/gui/src/components/settings/SettingsPageLayout.tsx
+++ b/gui/src/components/settings/SettingsPageLayout.tsx
@@ -1,7 +1,6 @@
 import classNames from 'classnames';
 import { ReactNode, useEffect, useRef } from 'react';
 import { useLocation } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


loucass003:
now that aw are not on tauri / have safari in the way.
we can prob use https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo
which is cleaner imo. Need to confirm with testing

aukaukauk:
Agreed, `scrollTo` is cleaner here. I updated it and re-tested the magnetometer link locally.

loucass003:
woops my bad. i should have linked https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView scrollIntoView

aukaukauk:
That makes sense. scrollIntoView({ block: 'start' }) changes the visual offset from the existing padding. Would you prefer using `scrollIntoView` with CSS `scroll-margin-top` to preserve that spacing? Personally, I think keeping a bit more top spacing helps place the target closer to the visual center.

loucass003:
yeah i think the css property is perfect for that. if i recall it used to be that a while back. but got changed for compatibility issues with safari. not an issue anymore

gui/src/index.scss
```diff
@@ -14,6 +14,14 @@ body {
   scrollbar-width: thin;
 }
 
+.settings-page-layout [id] {
```
loucass003: Use tailwind classes instead. dedicated css classes should only be declared when necessary

aukaukauk:
Done, updated with `scrollIntoView` + `scroll-margin-top`. I kept the existing 45/80px spacing values.

loucass003 (changes requested):

gui/src/components/settings/SettingsPageLayout.tsx
```diff
@@ -21,24 +21,20 @@ export function SettingsPageLayout({
...
+      });
     }
   }, [state]);
 
   return (
     <div
       ref={pageRef}
-      className={classNames('settings-page-layout', className)}
+      className={classNames(
+        '[&_[id]]:scroll-mt-[45px] mobile:[&_[id]]:scroll-mt-[80px]',
```
loucass003: im pretty sure you dont need the id selector.
and i would try to find the closest standard margin instead of hardcoding px values

loucass003 (approved):

Issue #1236, opened 2024-11-11
Sort GUI filtering / manufacturer property issue


Eirenliel:
> Also a PR should clearly say what the change does, why isSlime matters? I have no idea what this change means for HID devices that aren't original slimes, for example

I should have added, isSlime is how the GUI filters when you select the option to only show HMD and Slime trackers

so there is already DIY slimes or any slime made by third-party, is this an original slime? I don't know what type/manufacturer is appropriate, or if this is a wrong implementation for the intended outcome

_Originally posted by @sctanf in https://github.com/SlimeVR/SlimeVR-Server/issues/1235#issuecomment-2469194549_
            

We want to have Manufacturer field to ideally display an actual manufacturer. But right now the UI uses it for filtering and converts into nebulous isSlime check that just checks manufacturer, instead of checking some other field. This needs to be remedied at some point:

- [ ] Find a better property for UI to filter on when you check the show only trackers
- [ ] Add manufacturer field to firmware and protocol, display it properly in UI and in other info, make it definable in build parameters so manufacturers could define it themselves if they want

ishotjr:
just traced my way from `VENDOR_NAME` in the firmware, through `manufacturer` in the protocol, and ended up at https://github.com/SlimeVR/SlimeVR-Server/blob/1799483f265ce961f0b9b04862ec6bc8ed772d4d/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/UDPDevice.kt#L107 where I discovered the value is hard-coded, and then from there discovered this issue; I'd like to help with this, but I'm having a little difficulty understanding how to implement the actual data handoff -- would it be possible to get a little direction on this, and then I'll gladly take on the second task above? 💗

(also, does vendor === manufacturer or should it be an additional field in the fw, as implied by the second checklist item above? (although, it definitely already exists in the protocol, right? https://github.com/SomaticVR/SolarXR-Protocol/blob/main/schema/datatypes/hardware_info.fbs#L106 ))

PR #1835, opened 2026-05-06
Add manufacturer/vendor handling

ishotjr:
addresses part of #1236

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "solarxr-protocol"]
 	path = solarxr-protocol
-	url = https://github.com/SomaticVR/SolarXR-Protocol.git
+	url = https://github.com/SlimeVR/SolarXR-Protocol.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "solarxr-protocol"]
 	path = solarxr-protocol
-	url = https://github.com/SlimeVR/SolarXR-Protocol.git
+	url = https://github.com/SomaticVR/SolarXR-Protocol.git
--- a/gui/public/i18n/cs/translation.ftl
+++ b/gui/public/i18n/cs/translation.ftl
@@ -95,6 +95,8 @@ board_type-ESP32S3_SUPERMINI = ESP32-S3 Supermini
 board_type-GENERIC_NRF = Obecné nRF
 board_type-SLIMEVR_BUTTERFLY_DEV = SlimeVR Dev Butterfly
 board_type-SLIMEVR_BUTTERFLY = SlimeVR Butterfly
+board_type-SOMATICVR_ORION = Somatic Orion
+
 
 ## Proportions
 
--- a/gui/public/i18n/de/translation.ftl
+++ b/gui/public/i18n/de/translation.ftl
@@ -120,6 +120,7 @@ board_type-ESP32S3_SUPERMINI = ESP32-S3 Supermini
 board_type-GENERIC_NRF = Generisches nRF
 board_type-SLIMEVR_BUTTERFLY_DEV = SlimeVR Dev Butterfly
 board_type-SLIMEVR_BUTTERFLY = SlimeVR Butterfly
+board_type-SOMATICVR_ORION = Somatic Orion
 
 ## Proportions
 
--- a/gui/public/i18n/en/translation.ftl
+++ b/gui/public/i18n/en/translation.ftl
@@ -114,6 +114,8 @@ board_type-ESP32S3_SUPERMINI = ESP32-S3 Supermini
 board_type-GENERIC_NRF = Generic nRF
 board_type-SLIMEVR_BUTTERFLY_DEV = SlimeVR Dev Butterfly
 board_type-SLIMEVR_BUTTERFLY = SlimeVR Butterfly
+board_type-SOMATICVR_ORION = Somatic Orion
+board_type-undefined = Undefined
 
 ## Proportions
 skeleton_bone-NONE = None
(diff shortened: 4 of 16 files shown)


Eirenliel:
There is no SOMATICVR_ORION board type in firmware and shouldn't be. The whole idea of sending string as manufacturer was to get rid of the need to hardcode it and argue who should and shouldn't be added to board types.

Also translations should not be committed here, they're done via Pontoon.

loucass003:
yeah only en translations keys should be sent

ishotjr:
Thanks for the insight; I'm still learning! 💗 These changes correspond to the addition of `SOMATICVR_ORION` to `BoardType` in SolarXR:  

https://github.com/SlimeVR/SolarXR-Protocol/pull/211/changes/283641d9f2133f3b085ecd2005712d21b19ec9e6#diff-b05f90ae5bff957780c6d30ffa6cbfe47efb633444ac13b3f32104bb33ff33aeR31  

`[env:BOARD_SOMATICVR_ORION]` exists in this fork:  

https://github.com/SomaticVR/SomaticVR-Tracker-ESP/blob/main/platformio.ini#L254  

and the hope was to be able to have both the manufacturer and main board correctly populated in the server UI. If `BoardType` is set in stone now, should other boards use `CUSTOM` rather than their own types, or ? 💭📛  

Would it be helpful if I stripped the translation and `SOMATICVR_ORION` references and resubmitted the PR as just the manufacturer/vendor functionality? 🥺  

Will investigate Pontoon for translations -- thank you! 💗

Eirenliel:
The SolarXR PR is rejected for the mentioned reason. The text fields in protocol were added specifically for manufacturers to put anything they want in there, please use them. There is Vendor Name, URL and product name. It's not yet implemented in the GUI, but since you started on it would be cool if you could add that too :)

ishotjr:
Understood; I spent quite a lot of time digging around in all of this to try to figure out why [the fields we added](https://github.com/SomaticVR/SomaticVR-Tracker-ESP/blob/main/platformio.ini#L292) didn't seem to be doing anything, trying to understand how it worked or what the intent was -- and reached out as such [here](https://github.com/SlimeVR/SlimeVR-Server/issues/1236#issuecomment-4270527955); is there any documentation around this stuff that I'm missing? 💗

> There is Vendor Name, URL and product name. It's not yet implemented in the GUI, but since you started on it would be cool if you could add that too :)

I would love to! Can you confirm that I got it right with my existing `vendor === manufacturer` implementation? and then I'm guessing product name replaces the current board type/name in the GUI? how should the URL be used? 🤔💭🔗  

I'm guessing `UPDATE_ADDRESS`/`UPDATE_NAME` are best saved for another day, since they'll require extensive changes in order to function? 😅💾

Eirenliel:
It was added for future compatibility, we haven't implemented it ourselves. There is no documentation, but yes you're using it correct. Vendor is manufacturer. You can add URL as a separate line to the GUI that can be clickable by the user to open the manufacturer's site.

ishotjr:
Excellent! Will do!! Thank you so much for your quick and helpful responses!!! 💗✨

ishotjr:
@Eirenliel I'm trying to figure how things line up across the codebases and there seems not be a direct relationship between *SolarXR* and the firmware (unlike the server, which uses it directly), and beyond that, I'm not sure there are any accommodations in *SolarXR* for these new fields? I see [`model`](https://github.com/SlimeVR/SolarXR-Protocol/blob/main/schema/datatypes/hardware_info.fbs#L104) in *SolarXR* -- is that intended to be used for the value in `PRODUCT_NAME` in the firmware? and does `VENDOR_URL` exist at all, or should I add something for it in `HardwareInfo` in `hardware_info.fbs`? thanks for your patience as I try to understand how everything works together or is intended to! 💗

ishotjr:
@Eirenliel just wanted to check and see if you had any insight on this? 💗

loucass003:
our interfacing with solarxr is very rough for all the manufacturer stuff. honestly i wouldnt mind a proposal to clean it up.

nobody use it outside of gui so we can prob deprecate whatever  fields do not make sense and make new ones. ideally we would remove all the hardcoded ids and enums ouside of supported mcus and imus. then boards would become more abstract.
pure strings

ishotjr:
@loucass003 thanks for the quick response! will take a look at this... :bowtie:

unlogisch04:
I want to remind, That we have a problem with the current handshake packet that is still not solved. 
When on the same network are older or not patched ESP32 SlimeVR Compatible Tracker, They might crash, when a other tracker send all Manufacture Data,.. 
This is still not fixed in [SlimeVR-Tracker-ESP/main](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/516)

ishotjr:
@unlogisch04 I've applied @gorbit99's patch from https://github.com/kounocom/SlimeVR-Tracker-ESP/commit/6e21ba0b0d7d719bb924deadef41a9e6d8fccb2d#diff-f9c2c74e084edc8f3bda0c3f9eb75959d82ef5da197cabed835b90c0e92f6524L547 to a fork of the firmware and testing suggests it fixes it; is there anything I can do to help with this?

unlogisch04:
Should nothing holding up it seems like.

loucass003:
Just need to confirm what is the status of this? do we still need fw changes? 
What happened to solarxr changes?

unlogisch04:
> Just need to confirm what is the status of this? 

I believe some of the changes where like denied (adding more const for boards, Instead to use manufacturer. (well that would again break the fw flasher somehow. But would work for just fronted.

> do we still need fw changes? 

No, if you tracker with esp32-c* on a not patched firmware, and have also officials, the esp32-c* tracker will crash.
The needed firmware change, to avoid this would make half of that obsolete, as it would make a new packet and change protocol, So only the server would get big packets.  (well that was at least my idea to keep max compatibility to older trackers firmware)

> What happened to solarxr changes?

loucass003:
yeah new hardware info packet after handshake. 
 and new protocol version

I dont think adding const would break the fw tool. it is supposed to adapt based on the json in the repo. no matter what repo or branch

PR #1844, opened 2026-05-15
Remove some leftover log messages

ImSapphire:
The one in the TapDetectionManager is particularly annoying as it is logged every time a tracker is connected/disconnected, reassigned, or when changing tap detection settings. On startup it gets logged ~10 times.


Aed-1:
oops, left these in by accident while debugging, sorry 😅

loucass003 (approved):

PR #1845, opened 2026-05-16
Cache gcc packages in CI

ImSapphire:


ImSapphire:
okay this action sucks and doesn't seem to work properly on ubuntu-22.04

PR #1854, opened 2026-05-21
Allow creating and manipulating created trackers via SolarXR

ImSapphire:
SlimeVR/SolarXR-Protocol#214


ImSapphire:
Re-implemented in rewrite

PR #1872, opened 2026-06-05
Fix dongle reconnection bug

9il:
Correctly handle the dongle physical reconnection.

The current firmware contains a probabilistic workaround which requires high pull rate and still can fail.


sctanf:
Similar to #1808
TIMED_OUT tracker should not be set to OK, do you have a reason to change this?

9il:
> Similar to #1808 TIMED_OUT tracker should not be set to OK, do you have a reason to change this?

Fixed, thank you for the review!

loucass003 (approved):

sctanf (approved):

PR #1878, opened 2026-06-21
Get rid of grgit

ImSapphire:


loucass003 (approved):

server/android/build.gradle.kts
```diff
@@ -124,8 +123,14 @@ dependencies {
 
 // The android block is where you configure all your Android-specific build options.
 extra.apply {
-	set("gitVersionCode", grgit.tag.list().size)
-	set("gitVersionName", grgit.describe(mapOf("tags" to true, "always" to true)))
+	val tagCount = providers.exec {
+		commandLine("git", "--no-pager", "tag", "-l")
+	}.standardOutput.asText.get().trim().split('\n').size
+	val description = providers.exec {
+		commandLine("git", "describe", "--tags", "--always")
```
notpeelz: The length of the abbreviated commit SHA can change depending on how many objects exist in the git repo, which can lead to non-reproducible builds.

See https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---abbrevn

I recommend adding `--abbrev=7` or `--abbrev=8`.

ImSapphire: I was trying to match the previous behaviour (no explicit `abbrev`), but I guess I might as well fix that while I'm here :P

loucass003:
if we are changing this. we need to not forget the ones in the vite config

loucass003 (approved):

PR #1897, opened 2026-07-29
Toe tracking server rewrite

Sebane1:
Toe tracking, re-written to work with the server rewrite.

Associated solarxr PR
https://github.com/SlimeVR/SolarXR-Protocol/pull/220


server/core/src/main/java/dev/slimevr/resets/bodypart-sets.kt
```diff
@@ -75,3 +75,14 @@ val RIGHT_FINGER_PARTS = setOf(
...
+val LEFT_TOE_PARTS = setOf(
+	BodyPart.LEFT_BIG_TOE,
+	BodyPart.LEFT_INDEX_TOE,
+	BodyPart.LEFT_PINKY_TOE,
+)
+val RIGHT_TOE_PARTS = setOf(
+	BodyPart.RIGHT_BIG_TOE,
+	BodyPart.RIGHT_INDEX_TOE,
+	BodyPart.RIGHT_PINKY_TOE,
+)
```
Erimelowo:
```suggestion
```
This seems unused. These sets are for reset logic and should only be present when needed for that since they're also kinda subjective (for example, are feet part of the legs; are shoulders part of the arm)

server/core/src/main/java/dev/slimevr/skeleton/bone-geometry.kt
```diff
@@ -9,5 +9,13 @@ private val FINGER_OFFSETS = (
 		iterateBodyPartHierarchy(BodyPart.RIGHT_HAND, true)
 	).map { it.second }.associateWith { Vector3(0f, -0.025f, 0f) }
 
+private val TOE_OFFSETS = (
+	iterateBodyPartHierarchy(BodyPart.LEFT_FOOT, true) +
+		iterateBodyPartHierarchy(BodyPart.RIGHT_FOOT, true)
+	).map { it.second }.associateWith { Vector3(0f, 0f, -0.04f) }
```
Erimelowo:
```suggestion
	).map { it.second }.associateWith { Vector3(0f, 0f, -0.025f) }
```
Let's use a more sensible default to not have sausage toes

server/core/src/main/java/dev/slimevr/skeleton/bone-geometry.kt
```diff
@@ -9,5 +9,13 @@ private val FINGER_OFFSETS = (
...
+
 val DEFAULT_BONE_OFFSETS: BodyPartMap<Vector3> = DEFAULT_PROPORTIONS.toBoneOffsets()
-	.mutate { offsets -> for ((bodyPart, offset) in FINGER_OFFSETS) offsets[bodyPart] = offset }
+	.mutate { offsets ->
+		for ((bodyPart, offset) in FINGER_OFFSETS) offsets[bodyPart] = offset
+		for ((bodyPart, offset) in TOE_OFFSETS) offsets[bodyPart] = offset
+	}
```
Erimelowo:
```suggestion
	.mutate { offsets -> for ((bodyPart, offset) in FINGER_OFFSETS + TOE_OFFSETS) offsets[bodyPart] = offset }
```
Kotlin is awesome

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,129 @@
...
+			0 -> processToe(foot, lastAssigned, side, 0, oppositeSide(side), messages)
+
+			1 -> {
+				processToe(foot, lastAssigned, side, 1, oppositeSide(side), messages)
+				processToe(foot, lastAssigned, side, 2, oppositeSide(side), messages)
+			}
+
+			2 -> {
+				processToe(foot, lastAssigned, side, 3, side, messages)
+				processToe(foot, lastAssigned, side, 4, side, messages)
```
Erimelowo: I would have expected the little toe to be by itself. Is this on purpose?

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,129 @@
...
+
+	val euler = quaternionToEulerDegrees(currentRelative)
+
+	val pitch = euler.z
+	val yaw = euler.y
+	val tipToe = pitch < -14f
+	val bending = pitch > 15f && !tipToe
+	val splayed = when (splayDirection) {
+		FootSide.Left -> yaw < -7f
+		FootSide.Right -> yaw > 7f
```
Erimelowo: Please declare constants for these magic numbers
like `private const SPLAY_THRESHOLD_ANGLE = 7f`

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,129 @@
...
+	side: FootSide,
+	toeNumber: Int,
+	splayDirection: FootSide,
+	messages: MutableList<OscContent>,
+) {
+	val footRot = foot.rotation
+	val toeRot = toe.rotation
+	val currentRelative = footRot.inv() * toeRot
+
+	val euler = quaternionToEulerDegrees(currentRelative)
```
Erimelowo: We already have methods for that, please use that instead.
```suggestion
	val euler = currentRelative.toEulerAngles(EulerOrder.YZX)
```
(YZX is the one we use but your method might be XYZ so you can use that if it gives better results)

Erimelowo (changes requested):
Please use LITTLE_TOE instead of PINKY_TOE to follow our finger naming convention.

Also note that I didn't check GUI at this moment.

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
Sebane1: The toe right next to the pinky toe is influenced by the pinky toe more than the other two toes in my anatomical testing.

server/core/src/main/java/dev/slimevr/resets/bodypart-sets.kt
Sebane1: I think I just matched whatever existed for other joints to ensure reference weren't missing.
Toes are intended to have their own reset button (just not re-wired)

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,129 @@
...
+import dev.slimevr.osc.OscArg
+import dev.slimevr.osc.OscContent
+import dev.slimevr.osc.OscMessage
+import dev.slimevr.skeleton.BoneState
+import io.github.axisangles.ktmath.Quaternion
+import io.github.axisangles.ktmath.Vector3
+import solarxr_protocol.datatypes.BodyPart
+import kotlin.math.*
+
+private enum class FootSide { Left, Right }
```
Erimelowo: Please use `Side` from [util/side.kt](https://github.com/SlimeVR/SlimeVR-Server/blob/9b06eec608a7b216404774d632fda10ea329208a/server/core/src/main/java/dev/slimevr/util/side.kt) instead. (I just created it since we'll need such a class in other parts of the codebase like StayAligned.

server/core/src/main/java/dev/slimevr/resets/bodypart-sets.kt
Erimelowo: If you plan to use it eventually sure it's fine but please ass the missing RING and MIDDLE toes.

Erimelowo:
Also you'll have to update your solarxr branch for the PINKY -> LITTLE naming change :3

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
Sebane1: I just realized that neither of the OSC output encoders are housed by classes, and thus cannot do constants. Is this intended?

For now I'll just contain the values in local variables to the method.

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,122 @@
...
+			2 -> {
+				processToe(foot, lastAssigned, side, 2, side.opposite, messages)
+			}
+			3 -> {
+				processToe(foot, lastAssigned, side, 3, side, messages)
+			}
+			4 -> {
+				processToe(foot, lastAssigned, side, 4, side, messages)
+			}
+		}
```
Erimelowo:
```suggestion
		val splayDirection = when (segmentIndex) {
			0, 1, 2 -> side.opposite
			
			3, 4 -> side
		}
		
		processToe(foot, lastAssigned, side, segmentIndex, splayDirection, messages)
```

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,122 @@
...
+import io.github.axisangles.ktmath.Vector3
+import solarxr_protocol.datatypes.BodyPart
+import kotlin.math.*
+import dev.slimevr.util.Side
+import dev.slimevr.util.opposite
+
+private const val Absolute_Splay_Threshold_Angle = 7
+private const val Minimum_Tip_Toe_Pitch = -14
+private const val Maximum_Bending_Pitch = 15
+private const val Maximum_Absolute_Toe_Range = 90
```
Erimelowo: ALL_CAPS please for constants

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,122 @@
...
+
+private fun processToe(
+	foot: BoneState,
+	toe: BoneState,
+	side: Side,
+	toeNumber: Int,
+	splayDirection: Side,
+	messages: MutableList<OscContent>,
+) {
+
```
Erimelowo:
```suggestion
```

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,122 @@
...
+import solarxr_protocol.datatypes.BodyPart
+import kotlin.math.*
+import dev.slimevr.util.Side
+import dev.slimevr.util.opposite
+
+private const val Absolute_Splay_Threshold_Angle = 7
+private const val Minimum_Tip_Toe_Pitch = -14
+private const val Maximum_Bending_Pitch = 15
+private const val Maximum_Absolute_Toe_Range = 90
+internal fun buildToeMessages(bones: Map<BodyPart, BoneState>): List<OscContent> {
```
Erimelowo:
```suggestion
private const val Maximum_Absolute_Toe_Range = 90

internal fun buildToeMessages(bones: Map<BodyPart, BoneState>): List<OscContent> {
```

server/core/src/main/java/dev/slimevr/vmc/mapping.kt
Erimelowo, line 1: Would it not be better to just map the BIG_TOE to [LeftToes](https://docs.unity3d.com/2020.3/Documentation/ScriptReference/HumanBodyBones.LeftToes.html) and [RightToes](https://docs.unity3d.com/2020.3/Documentation/ScriptReference/HumanBodyBones.RightToes.html)? This feels like it won't do anything for receivers.

Erimelowo (changes requested):

server/core/src/main/java/dev/slimevr/vrcosc/toe-plugin-output-encoder.kt
```diff
@@ -0,0 +1,122 @@
...
+		Side.LEFT -> yaw < -Absolute_Splay_Threshold_Angle
+		Side.RIGHT -> yaw > Absolute_Splay_Threshold_Angle
+	}
+
+	messages.add(OscContent.Message(OscMessage("/avatar/parameters/TipToes${side.oscName}", listOf(if (tipToe) OscArg.True else OscArg.False))))
+	messages.add(OscContent.Message(OscMessage("/avatar/parameters/ToeBent${side.oscName}${toeNumber + 1}Bool", listOf(if (bending) OscArg.True else OscArg.False))))
+	messages.add(OscContent.Message(OscMessage("/avatar/parameters/ToeSplay${side.oscName}${toeNumber + 1}", listOf(if (splayed) OscArg.True else OscArg.False))))
+
+	val floatValue = (pitch / Maximum_Absolute_Toe_Range).coerceIn(-1f, 1f)
+	messages.add(OscContent.Message(OscMessage("/avatar/parameters/Toe${side.oscName}${toeNumber + 1}Float", listOf(OscArg.Float(floatValue)))))
```
Erimelowo: Compute stuff then add all at once

```suggestion

	val floatValue = (pitch / Maximum_Absolute_Toe_Range).coerceIn(-1f, 1f)
	
	messages.add(OscContent.Message(OscMessage("/avatar/parameters/TipToes${side.oscName}", listOf(if (tipToe) OscArg.True else OscArg.False))))
	messages.add(OscContent.Message(OscMessage("/avatar/parameters/ToeBent${side.oscName}${toeNumber + 1}Bool", listOf(if (bending) OscArg.True else OscArg.False))))
	messages.add(OscContent.Message(OscMessage("/avatar/parameters/ToeSplay${side.oscName}${toeNumber + 1}", listOf(if (splayed) OscArg.True else OscArg.False))))
	messages.add(OscContent.Message(OscMessage("/avatar/parameters/Toe${side.oscName}${toeNumber + 1}Float", listOf(OscArg.Float(floatValue)))))
```

You could also use [addAll](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/-mutable-list/#-1270570731%2FFunctions%2F532017514) to not repeat .add every line

Erimelowo:
There's also still 'pinky' used in 8 places in  `translation.ftl`

server/core/src/main/java/dev/slimevr/util/Side.kt
```diff
@@ -13,3 +13,8 @@ val Side.opposite: Side
 		Side.LEFT -> Side.RIGHT
 		Side.RIGHT -> Side.LEFT
 	}
+val Side.oscName: String
+	get() = when (this) {
+		Side.LEFT -> "Left"
+		Side.RIGHT -> "Right"
+	}
```
Erimelowo: This was better kept in your toe-plugin file as a private extension method since it's specific to it and not a standard throughout all of the codebase.

Erimelowo (dismissed):
Good apart from GUI but that's gonna be redone anyways

server/desktop/src/main/java/dev/slimevr/desktop/udp/address.kt
```diff
@@ -4,4 +4,7 @@ import io.ktor.network.sockets.InetSocketAddress
 import io.ktor.network.sockets.toJavaAddress
 
 // Prevent DNS lookup when getting the address
-fun resolveDesktopUdpAddress(addr: InetSocketAddress): String = (addr.toJavaAddress() as java.net.InetSocketAddress).hostString
+fun resolveDesktopUdpAddress(addr: InetSocketAddress): String {
+	val socketAddress = addr.toJavaAddress() as java.net.InetSocketAddress
+	return "${socketAddress.hostString}/${socketAddress.port}"
```
Erimelowo, line 9: Why this change?

Sebane1: This was discussed with Futura and made it into this branch by accident.
I was trying to get wifi repeated smols working again, and apparently we were no longer differentiating  connections by port number. I was given this line to replace to restore functionality.

PR #1902, opened 2026-08-06
Support configurable SPI bus pins in Firmware Flasher

Feather-P:
Updates the Firmware Flasher UI to support the configurable SPI bus pins introduced in SlimeVR/SlimeVR-Tracker-ESP#539.

Adds a per-node ComponentLayout that is resolved from the schema path and applied when rendering the board-defaults graph. 
Layout is driven by the schema path rather than hard-coded checks, so the new sections get their layout while the legacy default schema keeps its existing layout and behavior unchanged.

Adds a schema-driven default-value generator that rebuilds a full object from the JSON Schema defaults when a discriminator is switched or a new array item is added.
It also materializes array items from the schema (imu entries) that the old code left empty, which crashed the frontend.

This is the Server counterpart of:

- SlimeVR/SlimeVR-Tracker-ESP#537
- SlimeVR/SlimeVR-Tracker-ESP#539

The firmware-side SPI bus configuration is implemented in
SlimeVR/SlimeVR-Tracker-ESP#539

### DIFF
--- a/gui/app/src/components/commons/Dropdown.tsx
+++ b/gui/app/src/components/commons/Dropdown.tsx
@@ -288,7 +288,7 @@ export function DropdownInside({
     <>
       <div
         className={classNames(
-          'min-h-[42px] text-background-10 text-left dropdown',
+          'min-h-[42px] min-w-0 text-background-10 text-left dropdown',
           displayStyles[display]
         )}
         onClick={() => setIsOpen(!isOpen)}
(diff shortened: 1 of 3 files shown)


ImSapphire (approved):
Changes look sane to me

loucass003 (approved):

ImSapphire:
Thank you!

Issue #1880, opened 2026-06-22
Feature Freeze - Rewrite in Progress


loucass003:
## Feature Freeze - Rewrite in Progress

As many of you know, we've been working on a large-scale rewrite of the SlimeVR codebase. We're happy to share that the rewrite has reached a stage where we're ready to shift our full focus toward it.

To avoid having to continuously backport new features and keep the two codebases in sync until the rewrite is complete, **new feature PRs can no longer target `main`.** Please point them to `server-rewrite` instead.

Note that `server-rewrite` does not yet have full feature parity with `main`, we are actively working towards it. Full feature parity will be reached before any public release of the rewrite.

**What we still accept on `main`:**
- Bug fixes
- Required maintenance and stability improvements

Already opened feature PRs targeting `main` may still be merged if they are reasonably scoped. Overly large ones will be closed and redirected to `server-rewrite`.

We will also be releasing a new version soon with the current contents of main. This will likely be the last feature release on the current codebase, outside of bugfix releases.

We know this may be disappointing if you had a feature PR in mind, and we genuinely appreciate everyone who has contributed. This is a necessary step to get the rewrite across the finish line without falling further behind.

Thank you for your patience and continued support! :heart:

Erimelowo:
[Server rewrite PR](https://github.com/SlimeVR/SlimeVR-Server/pull/1798)

PR #1912, opened 2026-08-21
Add motion detection to perform full & mounting resets only when keeping still

nekomona:
Currently the reset only waits for a fixed timeout before taking the calibration data, which may cause bad data if the user accidentally moves during reset. Also the user would be in a rush to get into the mounting reset position.

This PR adds motion detection to the server and corresponding visual markers to the reset buttons in the GUI. Now during full & mounting resets, if the user is in large motion, the reset button would show a "Keep Still !" info, with the button edge reflecting motion strength identical to the tracker cards. After the motion stops, the countdown would return to normal.

With this PR the reset process should be less tricky and less likely to need for a retry.

### DIFF
--- a/gui/public/i18n/en/translation.ftl
+++ b/gui/public/i18n/en/translation.ftl
@@ -254,6 +254,7 @@ reset-mounting = Mounting Calibration
 reset-mounting-feet = Feet Calibration
 reset-mounting-fingers = Fingers Calibration
 reset-yaw = Yaw Reset
+reset-wait_still = Keep Still !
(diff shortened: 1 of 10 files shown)


loucass003:
please read https://github.com/SlimeVR/SlimeVR-Server/issues/1880.

main is currently frozen. we wont take feature changes on it until rewrite is complete

Erimelowo:
https://github.com/SlimeVR/SlimeVR-Server/issues/1913

nekomona:
oopsie. Will be back for it after the rewrite XP

PR #1914, opened 2026-08-25
Apply signal strength to the tracker it names

9il:
`UDPPacket19SignalStrength` parses a tracker id, but the handler ignored it and wrote the value to every tracker on the connection — so on a device reporting more than one, whichever sent last set the signal strength for all of them. Every neighbouring handler resolves the id; `UDPPacket20Temperature`, right below it, does `getTracker(packet.sensorId)`.

Id 255 still means the device as a whole, which is what the ESP firmware sends here, so existing trackers are unaffected. Any other id now updates only the tracker it names. The 255 convention gets a name, which `setConfigFlag`'s default reuses.

This matters for a proxy device relaying several wireless trackers over one connection: each tracker has its own id there, and today they all display one arbitrary tracker's signal — exactly the value you need per tracker when one is at the edge of range.

`:server:core:compileKotlin` and `:server:spotlessCheck` pass (JDK 21). No test: there's no harness for the UDP path, and this handler needs a live `UDPDevice` plus the `VRServer` singleton — happy to add one if you want it.

Per CONTRIBUTING: I used Claude Code to draft this patch (the commit carries a `Co-Authored-By` trailer). I found the behaviour on my own device and reviewed the change.

### DIFF
--- a/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
+++ b/server/core/src/main/java/dev/slimevr/tracking/trackers/udp/TrackersUDPServer.kt
@@ -271,7 +271,7 @@ class TrackersUDPServer(private val port: Int, name: String, private val tracker
 	)
 
 	private val queues: MutableMap<Triple<SocketAddress, ConfigTypeId, Int>, Deque<ConfigStateWaiter>> = ConcurrentHashMap()
-	suspend fun setConfigFlag(device: UDPDevice, configTypeId: ConfigTypeId, state: Boolean, sensorId: Int = 255) {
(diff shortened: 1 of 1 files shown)


loucass003:
First of all we do not accept AI generated PRs.
Second of all, main server is under code freeze until the rewrite is complete #1880 

And most of all, why would a tracker have separate rssi from its device.
as per slime terminology device = mcu. tracker = imu.

9il:
1. You told you use AI yourself for SlimeVR a lot. What is wrong with the a proper PR?
2. OK, so you just closes PRs you don't like?
3. Ah, so you didn't even read it. OK, I can repeat: "This matters for a proxy device relaying several wireless trackers over one connection"

9il:
nevermind. good luck

PR #1926, opened 2026-09-12
Fix highlight intensity based off accumulated motion samples only advancing when rotation dependency changes

hiccup444:
Addresses issue #1917, updating motion samples to expire after 500 ms even when updates stop. Acceleration-only changes update the highlight, and timers clean up on unmount.


Erimelowo:
Hi, thanks for you PR! Can you please target https://github.com/SlimeVR/SlimeVR-Server/tree/server-rewrite

hiccup444:
My former patch uses an unused file in the server rewrite, I will just make a new PR.

PR #1933, opened 2026-09-20
fix: use base64 serial command for Wi-Fi provisioning

qus20000:
## Summary

This changes serial Wi-Fi provisioning to send credentials with `SET BWIFI` instead of quoted `SET WIFI`.

The SSID and password entered in the UI are still handled as normal text in the app, but are encoded as UTF-8 base64 immediately before being written to the tracker serial port.

## Why

The tracker firmware already supports `SET BWIFI` for base64-encoded Wi-Fi credentials. This avoids command parser ambiguity when SSIDs or passwords contain spaces, quotes, non-ASCII characters, or punctuation.

This is a real setup issue for users in Korea: ISP-provided routers from the three major carriers commonly ship with SSIDs and/or default passwords containing special characters, for example `+` or `@`. Sending those values through the quoted `SET WIFI "<ssid>" "<password>"` command can fail during provisioning even though the same credentials are valid.

Using `SET BWIFI` makes the serial payload command-safe while keeping the UI behavior unchanged.

## Compatibility

`SET BWIFI` is supported by the stock SlimeVR Tracker ESP firmware and was added as part of the firmware's base64 Wi-Fi credential command support.

## Testing

- Ran `git diff --check`
- Ran `.\gradlew.bat :server:desktop:compileKotlin --no-daemon`
- Android compile was not run locally because Android SDK is not configured in this environment


loucass003:
Please implement it to server rewrite. see #1880

qus20000:
> Please implement it to server rewrite. see #1880

Hello, Thanks for the clarification. I wasn’t aware that the server rewrite was already in progress, and I’m sorry I missed that.

I’ll review the server-rewrite structure, rework the changes if needed, and then submit a PR against the rewrite branch.

PR #1934, opened 2026-09-21
Fix Android inset bug

abb128:
before
image

after
image

This is needed because of https://developer.android.com/about/versions/15/behavior-changes-15#window-insets


gui/app/src/AppLayout.tsx
```diff
@@ -26,6 +26,13 @@ export function AppLayout() {
         `${config.textSize}rem`
       );
     }
+
+    if (window.__ANDROID__?.isThere()) {
+      const color = window
```
loucass003: I would prefer if you could. inject a root tag for android. then do the color override from css.
That way other css features could be overriden from here in the future 🙏

loucass003 (approved):

